PR #4, opened 2021-07-09
Fix wireguard command

jason-jaia:


tsaubergine:
Thanks for catching these. I also fixed the typo with the wrong subnet (192.168 should be 172.20): https://github.com/jaiarobotics/jaiabot/pull/4/files#diff-c0a253e086a317b724262826c0bd1fad733ce1bfa318c590adfb044d446e72e8L118

tsaubergine (approved):

PR #9, opened 2021-07-15
Add stats to jaiabot_lora_test

tsaubergine:
Add packet success and range to "stats" debug group and also add to LoRaReport:

image


jason-jaia (approved):
Looks good - thanks! I was hoping to catch an error, but no.

PR #23, opened 2022-03-16
Stub application to send command from hub to bot

tsaubergine:
No need to merge if you don't need it, just putting here for your reference.

```
cd config/launch/standard
./all.launch
```
```
jaiabot_hub_command_publisher --interprocess 'platform: "hub"' -vvv --app 'simulation { time { use_sim_time: true  warp_factor: 10 }}}'
```


jason-jaia:
Learned what we needed - no need to merge and build it all the time.

PR #24, opened 2022-03-28
Add DivePacket

tsaubergine:
See https://trello.com/c/Muyo49H5/22-dive-packets

This DivePacket gets published after each dive over intervehicle, and is currently subscribed for by `jaiabot_hub_manager` which just republishes it on interprocess.

image

Still need to figure out "Deceleration indication". 

Also added support to the jaiabot_simulator for temperature and salinity, read off a basic profile in the config file with linear interpolation centering a normal distribution with configurable standard deviations to provide some randomness to the values. These feed into the applicable Goby applications using the string based comma delimited syntax that the python part of the driver produces.


jason-jaia:
Thanks for adding the simulated salinity and temperature! That was something we kept thinking about, but never had the time to investigate and implement. As far as the deceleration, I'll break that off into a new card since it needs a bit of work and can be added to your dive packet fairly easily now.

PR #25, opened 2022-04-03
Handle DesiredSetpoint messages in bot_pid_control app

edsanville:
The bot_pid_control app should now subscribe to, and handle, the DesiredSetpoint messages.  I'm testing it using the config/launch/high_low_test launch files.  These are basically the jaia simulator, with the bot_pid_control app added to catch the setpoint messages published by the mission_manager app.  However, while testing, I didn't actually see any messages come through.  Perhaps Toby can figure out why, and tweak it so the bot_pid_control app gets these messages?


tsaubergine:
> The bot_pid_control app should now subscribe to, and handle, the DesiredSetpoint messages. I'm testing it using the config/launch/high_low_test launch files. These are basically the jaia simulator, with the bot_pid_control app added to catch the setpoint messages published by the mission_manager app. However, while testing, I didn't actually see any messages come through. Perhaps Toby can figure out why, and tweak it so the bot_pid_control app gets these messages?

The receiving of DesiredSetpoint looks fine to me, perhaps the confusion is that no jDesiredSetpoint message is sent except when the mission manager is in the InMission state (https://docs.jaia.tech/md_page60_mission_management.html#autotoc_md66)? 

If it's helpful I can update the mission manager to send SETPOINT_STOP in all the other states as well.

image

src/bin/bot_pid_control/app.cpp
```diff
@@ -413,3 +419,81 @@ void jaiabot::apps::BotPidControl::handle_command(const jaiabot::protobuf::PIDCo
...
+
+    if (desired_course.has_heading()) {
+        rudder_is_using_pid = true;
+        target_heading = desired_course.heading();
+    }
+    if (desired_course.has_speed()) {
+        throttleMode = PID_SPEED;
+        target_speed = desired_course.speed();
+    }
+    // TO DO:  PID for the depth that uses elevators while moving forward
```
tsaubergine, line 514: We can ignore pitch, roll, z_rate, and altitude options for now. The IvP behaviors I'm using only use the heading, speed, and depth control domains.

tsaubergine:
So that we have the same code format, could you please ensure you've run  `./git-pre-commit-format install` from within scripts/clang-format-hooks once after cloning this project to enforce clang-format on each Git commit?

https://docs.jaia.tech/md_page19_repository.html

config/launch/standard/bot.launch
```diff
@@ -1,7 +1,5 @@
 #!/usr/bin/env -S goby_launch -L -P -d 1000
 
-[kill=SIGTERM] gpsd $(../../gen/bot.py gpsd)
```
tsaubergine: Why is this removed? It's required for the lat/lon input from the simulator to goby_gps

PR #26, opened 2022-04-05
Update simulator to use PID values from bot_pid_control

tsaubergine:
- No need for new launch set (high_low_test), better to update existing "standard" launch set
- Add bot_pid_control in place of pMarinePID in "standard" launch set: implements "Simulator Pass 2 system" at https://docs.jaia.tech/md_page11_simulator.html
- Ran clang-format on edited files, please use the git hook so we don't get out of sync on this formatting in the future: https://github.com/jaiarobotics/jaiabot/pull/25#issuecomment-1088358034
- Changed timeout to goby::time::MicroTime and set configurable default (of 5 seconds) so that if we don't receive a command in bot_pid_control we timeout after that.

**Doesn't work yet, but that's due to problems in bot_pid_control I think**


edsanville:
Merging this in

PR #27, opened 2022-04-18
Dive to seafloor

tsaubergine:
- Adds constant seafloor depth to simulator (set to 50m currently)
- Add seafloor detection to Dive state using `bottoming_timeout` parameter: the number of seconds where the vehicle doesn't increase its depth by more than dive_depth_eps before assuming the bottom has been reached.
- After the seafloor is detected, this depth is held and the rest of the dive depth targets are removed.

To command a dive to the seafloor, omit the dive_depth parameter or set it to a value greater than the seafloor depth.


edsanville (commented):
Merged

PR #29, opened 2022-04-22
Systemd

tsaubergine:
See also https://github.com/jaiarobotics/jaiabot-debian/pull/1 in coordination with this PR (and https://github.com/jaiarobotics/jaiabot-python-venv).

## Systemd
This pull request implements code to use `systemd` (the standard `init` on most modern Linux systems) instead of `goby_launch` for JaiaBot/JaiaHub process lifecycle management on the bots/hubs (we'll still use `goby_launch` for simulation and the like):
- systemd "service" files are generated much like the jaiabot application configuration using a Python Template generation script (config/gen/systemd.py) and a set of templates (config/templates/systemd).
- Environmental variables used by systemd are still "seeded" by `preseed.goby` at the time of generation by `systemd.py`. This may eventually change, but for now I want to maintain parity between `goby_launch` and `systemd` as much as possible.

## Python / Config packaging
Additionally, it updates the Debian packages built during the CI/CD pipeline to include all the code in `jaiabot` (specifically the Python code and the configuration generation code was not previously packaged):
- Added new [jaiabot-python-venv](https://github.com/jaiarobotics/jaiabot-python-venv) Git repository which contains a virtual environment (venv) of all the `pip` installed Python dependencies. Rationale: We control the dependency versions and don't rely on a specific package on PyPi being available or upgraded. This can then be included in the `jaiabot-python` Debian package. Use of this venv is optional (you can still do a `pip install` at the system level if desired). The venv is included as a Git submodule at jaiabot/src/python
- Renamed Python scripts to be prefixed with `jaiabot_`.

## Embedded setup

I also added a `jaiabot_embedded` Debian "metapackage" that installs all the jaiabot run dependencies (`jaiabot-apps`, `jaiabot-python`, `goby-apps`, `moos-ivp-apps`, etc.). Within this metapackage I've added Debian lifecycle scripts (postinst, prerm, etc.) to run `system.py`, enable and start the systemd processes (and remove them upon package removal). As well, I've added a `debconf` configuration to grab the type (bot or hub) bot or hub id, and fleet id and use these to set the parameters for `systemd.py`.

This does not remove all the functionality of scripts/setup_embedded/setup_embedded.sh (some of it is low level setup that probably belongs in cloud-init or similar). So for, now, at the expense of creating more complication, I moved scripts/setup_embedded/setup_embedded.sh to `setup_embedded_dev.sh` for use with the existing setup (that is, compiling jaiabot from source or using Docker), and then added `setup_embedded_release.sh` which does the remaining setup steps required for setup when using the new `jaiabot_embedded` Debian package.

My goal is to get away entirely from these scripts over time (at least the "_release" one), probably via cloud-init or live-build generation of the Rasp Pi filesystem. One option would be to fork this project (which I mostly wrote, for the RP CM3): https://bitbucket.org/ooicgsn/cgsn-mooring_rootfs-build

## Minor
- Added Ubuntu "message of the day" (MOTD) that gives systemctl information about all the jaiabot services upon login.
- Moved `/etc/jaiabot/dev` symlinks to `/dev/jaiabot` to be more standard (doesn't change functionality, but we should try to be predictable and standardized wherever possible so our customer developers have the best impression and user experience possible).
- `libxbee.so` -> `libjaiabot_xbee.so`
- Runtime logs are in /var/log/jaiabot now (not ~/jaia-logs)


edsanville:
Merged

PR #39, opened 2022-05-24
Only set datum using new mission in initial PreDeploy state, not during Replan

tsaubergine:
Hopefully this will solve the pesty race condition causing pHelmIvP to declare the mission point reached on single-point back-to-back missions.


tsaubergine:
[Update to use 1 datum per launch](https://trello.com/c/2P2JYSLy/197-update-to-use-1-datum-per-launch)

PR #40, opened 2022-05-24
Set active goal in bot status message

tsaubergine:
Added "active_goal" index to the BotStatus message which maps onto the order in which the Goals are defined in the MissionPlan (starting with 0). When in MOVEMENT__TRANSIT, this value indicates the next waypoint. When in TASK__*, this indicates the Goal for which the current task resides.

Also tidied up use of `boost::units::degree::plane_angle()` to `boost::units::degree::degrees`.

### DIFF
--- a/src/bin/fusion/fusion.cpp
+++ b/src/bin/fusion/fusion.cpp
@@ -32,12 +32,12 @@
 
 #include "config.pb.h"
 #include "jaiabot/groups.h"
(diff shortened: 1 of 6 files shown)


tsaubergine:
[Add next waypoint to status or separate message.](https://trello.com/c/bgEKsWPL/203-add-next-waypoint-to-status-or-separate-message)

PR #57, opened 2022-06-29
Create log directory with jaiabot so it's readable/writable with normal user

tsaubergine:
Otherwise root service such as jaiabot_health can make the log subdirectories with user root.


ferrom-jaia (approved):

jason-jaia (approved):
This looks good!

PR #66, opened 2022-07-18
Implement some PostDeployment states to allow shutdown, redeploy

tsaubergine:
- Re-added jaiabot_health to bot and hub simulation launch files (not sure how these got removed).
- Finished some more the jaiabot_mission_manager state machine (set SelfTest to require a HEALTH_OK goby_coroner health report), and added in some of the PostDeployment transitions.
- Add shutdown to jaiabot_health (as this is run as root and thus can execute these commands) in two ways:
   - SHUTDOWN is the "normal" route through jaiabot_mission_manager and can only be used when EvShutdown is handled (see src/doc/figures/mission-states.png), namely in PreDeployment and PostDeployment::Idle.
   - SHUTDOWN_COMPUTER is the debug path that acts like RESTART_COMPUTER and RESTART_ALL_SERVICES and is handled immediately regardless of the jaiabot_mission_manager state.


tsaubergine:
[Shutdown button](https://trello.com/c/eq4QXcZH/221-shutdown-button)

tsaubergine:
This still needs a "RECOVERED" button or similar to kick off the log copying and put the vehicle in PostDeployment::Idle from where SHUTDOWN will be accepted, see: https://docs.jaia.tech/md_page60_mission_management.html

edsanville:
Ok, will add.

tsaubergine:
Merging so I can get `packages` merged and stop trying to upload packages to the old repo. We can address the UI issues in another PR. https://trello.com/c/eq4QXcZH/221-shutdown-button

PR #68, opened 2022-07-20
Health

tsaubergine:
*base branch for this PR is "shutdown" as that branch fixes running jaiabot_health in simulation*

Initial framework for health reporting (lots of implementation to do still; could use a hand).

- Defines enumerations for Jaiabot specific Errors (corresponding to HEALTH_FAILED) and Warnings (-> HEALTH_DEGRADED) that are now sent via BotStatus (up to 5).
  - These need to be displayed on Central Command.
- Implements conversion of goby_coroner error (goby::middleware::protobuf::ERROR__PROCESS_DIED) when a process doesn't respond to its health request into the appropriate `ERROR__NOT_RESPONDING__XXX` enum value. This mapping (string app name that goby_coroner uses to Error enum) is automatically included at compile time by CMake (JAIABOT_HEALTH_PROCESS_MAP_ENTRIES cpp define) so when we add a new app (`jaiabot_yyy`), compilation will fail if we forget to add a corresponding `ERROR__NOT_RESPONDING_JAIABOT_YYY` value.
- Placeholders (in health.proto) for all the other errors and warnings we need to implement (at least those listed so far on Trello: https://trello.com/c/ImFwmxJ3/16-11-error-health-handling) and a quick note where I think they should be implemented.
- Doc page describing health reporting infrastructure. 
- Implemented NTP, RAM, CPU load, and disk usage checks using forked GPL code that I wrote for WHOI

E.g. BotStatus if a couple apps die:

image


ferrom-jaia (commented):
Looks Great!

PR #71, opened 2022-07-21
packages.jaia.tech

tsaubergine:
- Set up new packages.jaia.tech VM with 3 repositories:
    - continuous/1.y: all commits to 1.y
    - beta/1.y: all tags that contain  a _ (e.g. 1.0.0_beta1) which is changed to ~ (e.g. 1.0.0~beta1). Git does not allow ~ in tag names, hence the substitution.
    - release/1.y: all tags that do not contain a _ (e.g. 1.0.0).
- Implements CircleCI building packages to the three new Ubuntu repositories on packages.jaia.tech  
    - Tested by creating somewhat meaningless 0.1.0 and 0.1.0_beta1 releases that sort before 1.0.0 real releases. *Deleted the old 1.0.0_beta1* testing tags.
- Removed all jaiabot packages from packages.gobysoft.org
- Updated to "Large" CircleCI instance when building .deb packages and "Small" when doing low CPU usage tasks. Shortens the build time.


tsaubergine:
[1. Create packages.jaia.tech](https://trello.com/c/XGTYNZQ5/324-1-create-packagesjaiatech)

tsaubergine:
Merging so we don't keep creating downstream issues when merging other PRs (and trying to upload to old package server / use old GobySoft GPG signing key). Please feel free to review still and mark any concerns as new Trello cards / issues.

PR #74, opened 2022-07-22
added ntp for running sim locally

jason-jaia:
I added ntp to the runtime tools which was causing jaiabot_health to crash in the simulator.


ferrom-jaia (approved):

oceanzus:
I was wondering why my local sim was bombing out. Nice catch.

PR #93, opened 2022-07-28
Fix ERROR__FAILED__PYTHON_JAIABOT_WEB_APP var names

edsanville:


tsaubergine (approved):
Thanks for fixing my sloppiness.

PR #102, opened 2022-08-01
Implement data offload

tsaubergine:
This finalizes the Mission Manager state machine by implementing the DataOffload state.

Video walking through usage (too large for Github):

https://youtu.be/4kVOqJNHx0Q

Video showing Offload failure and retry:

https://user-images.githubusercontent.com/732276/183573946-1247c8cd-9d13-444e-8a79-a87709811151.mp4

The DataOffload state runs a shell script in its own thread, which is set to the new script `scripts/jaiabot-dataoffload.sh`
- scripts/jaiabot-dataoffload.sh: 
   1. rsync non-.txt data to `/var/log/jaiabot/bot_offload` so all the .goby files are in the same directory for the Analyze tool
   2. Compress any .txt (glog) files using xz to reduce space.
   3. Remove any .txt.xz (compressed glog) files that are older than 7 days
   4. Remove any other log files (e.g., .goby) files that are older than 14 days

Assumptions:
- jaiabot-dataoffload.sh assumes IP address for 'hub0' is defined in /etc/hosts as that is where the data are copied.
- User of hub0 is `jaia`
- SSH keys are already shared (public key of bot is on hub's /home/jaia/authorized_keys). This is card https://trello.com/c/ndNNODmA/347-create-pathway-for-generating-and-sharing-ssh-keys-on-setup-1
- rsync and xz are available on the hub/bot (rsync was already installed, xz added at https://github.com/jaiarobotics/jaiabot-rootfs-gen/pull/10)

Other changes:
- Added configuration option for mission manager (stop_logging_while_idle) to disable logging when in Pre/PostDeployment::Idle (defaults to true, so we must explicitly set it to false if we want to log everything including during Idle).
- Added StartingUp state before Idle which waits until EvStarted (triggered off a non-HEALTH__FAILED health report). This ensures that every app has a chance to start up and publish their configuration prior to entering PreDeployment::Idle (and thus disabling logging). If the system does not get an EvStarted within a configurable `startup_timeout` seconds (default 120s), it goes into PreDeployment::Failed.
- Added EvRetryDataOffload and corresponding command type RETRY_DATA_OFFLOAD to retry the data offload. Since the jaiabot-dataoffload.sh uses rsync and is reentrant, running this command after a successful offload will just simply succeed again within a few seconds.
- Exclude the goby internal groups from the .goby file.
- Add warnings for infeasible missions


tsaubergine:
[S2- Add a Get Logs feature to the Hub [0.5 for the autonomy part]](https://trello.com/c/DJVoJrJr/318-s2-add-a-get-logs-feature-to-the-hub-05-for-the-autonomy-part)

jason-jaia:
[S3 - Data offload using rsync from the bot to the hub [0.5]](https://trello.com/c/2KaUxuXY/364-s3-data-offload-using-rsync-from-the-bot-to-the-hub-05)

ferrom-jaia (approved):
Awesome Job!

PR #106, opened 2022-08-01
Add optional JAIA_BUILD_NPROC to control the number of processes that the build scripts use

edsanville:
Motivation:  I have a lot of cores on a couple of my machines, and building can take all the RAM and crash the OS.


tsaubergine:
> Motivation: I have a lot of cores on a couple of my machines, and building can take all the RAM and crash the OS.

So clearly you need more RAM :)

tsaubergine (approved):

PR #123, opened 2022-08-09
removed Restart=never

jason-jaia:


ferrom-jaia (approved):
Restart will use the default value "no" by removing this line

PR #127, opened 2022-08-11
Created a doc page with a timing table (incomplete).

edsanville:
The major unknown here is the radio latency.  How long does it take the typical Command or EngineeringCommand to go from hub to bot over the radio?  We may need to write some precise tests to measure this.


jason-jaia (approved):

ferrom-jaia:
@jason-jaia @edsanville 
I think it would be useful to also point to where these times are located so they can be easily changed

PR #130, opened 2022-08-12
Hardcode motor_max_step to be five times faster when slowing the moto…

edsanville:
…r down


jason-jaia (approved):

jason-jaia:
[S4 Feature/Asymetric ramping of motor [0.5]](https://trello.com/c/JnCqCRvK/460-s4-feature-asymetric-ramping-of-motor-05)

PR #132, opened 2022-08-15
bug/Add-Try-Catch-Arduino-Parse

ferrom-jaia:
Added try catch for parsing arduino response, made the arduino proto fields required, and updated number of tries for report failure app to get a acknowledgement


jason-jaia:
We need to always send the LED state. I think ON when sending commands to the motor and actuators and OFF when in all-stop.

ferrom-jaia:
> We need to always send the LED state. I think ON when sending commands to the motor and actuators and OFF when in all-stop.

Updated!

michaelrockJAIA (approved):

PR #145, opened 2022-08-19
task/Rename-plot-and-converter

ferrom-jaia:


edsanville (commented):
In the README,.md, we need to update that ~/jaiabotplot-logs path to the current /var/log/jaiabot/bot_offload path that is being used.

jason-jaia (approved):

PR #150, opened 2022-08-20
feature/separate_arduino

jason-jaia:
I split the arduino into two groups so we could more easily follow in liaison


jason-jaia:
I meant to say too that I didn't test it, but someone with your eagle coding eyes could see that it was either obviously good or obviously trash.

michaelrockJAIA (approved):

PR #166, opened 2022-08-23
Jaiaplot GUI tweaks to make our lives easier

edsanville:
* Map path is only the part corresponding to the plot x-axis
* Plots and map keep the time domain when adding a new series
* Map allows zooming to level z=20, for those small missions


ferrom-jaia (approved):
Tested in sim. I did notice that maxzoom causes the background map to disappear.

PR #168, opened 2022-08-25
Fix error when pidControl is not present in engineering object

edsanville:
Fix warnings about missing keys in the error/warning message divs in BotDetails


ferrom-jaia (approved):
Reviewed in sim

PR #170, opened 2022-08-25
Task/change logging levels poll rates

ferrom-jaia:


jason-jaia (approved):
Code review looks good. Approving to get the updated logging levels, but we should do a quick scrub of all the warnings in src/bin to see if any others should really be debugs.

PR #178, opened 2022-08-29
Allow dive without GPS by directly subscribing to pressure in mission_manager

tsaubergine:
- Use pressure directly in mission_manager to compute depth, defaulting to latitude of 45 degrees unless a valid latitude is received.
- Allow disabling GPS in simulator.

Stops at ReacquireGPS, but that is fixed by Indoor Mode in https://github.com/jaiarobotics/jaiabot/pull/187


tsaubergine:
[S6 - feature/RC-Dive-No-GPS [1]](https://trello.com/c/eManl1P5/507-s6-feature-rc-dive-no-gps-1)

ferrom-jaia (approved):
Tested in simulation. Looks good!

PR #180, opened 2022-08-29
CC issues RCMode commands using bot's current location

edsanville:
If there is no GPS location, CC will warn the user and confirm that they still want to use (0, 0) as the datum.


ferrom-jaia (approved):
Tested in sim!

PR #183, opened 2022-08-29
CC: Bot list on left side, no bins, Jaia logo, Engineering Wrench right side

edsanville:
Some other tweaks, too


ferrom-jaia (approved):
Tested in Sim.

PR #190, opened 2022-08-30
Fixed issues with MOOS download

edsanville:
Disable MOOS download button, unless user has selected log(s)


ferrom-jaia (approved):
Code looks good. Tested with sim data.

PR #192, opened 2022-09-01
moved heading over 360 check to after the addition of the magnetic de…

jason-jaia:
…clination

Tested on a bot - saw the error and deployed and don't see the error. I realize now that if the magnetic declination can be negative, then we need to have a check for degrees under 0. This is in a code block for attitude from tpv data - maybe we should just have the same logic for both code blocks.


ferrom-jaia:
@jason-jaia Updated with function to correct heading

jason-jaia (commented):
Code review looks good.

michaelrockJAIA (approved):
review in rush

PR #193, opened 2022-09-02
Make Dead Man's Switch into a clickable/touchable div

edsanville:
For tablets, etc.


ferrom-jaia (approved):
Approving based on code and comments from ed. Creating bug for mouse click

PR #195, opened 2022-09-08
Make xbee driver configurable, with network ID and a list of peer map…

edsanville:
…pings from node_id to serial_number


tsaubergine:
The generation of the xbee config should be added to fleet_config.sh: https://github.com/jaiarobotics/private_jaia/pull/3

This could be a new Trello card instead.

Otherwise this looks ok to me.

config/gen/bot.py
```diff
@@ -31,7 +31,7 @@
 node_id=common.bot.bot_index_to_node_id(bot_index)
 
 verbosities = \
-{ 'gobyd':                                        { 'runtime': { 'tty': 'WARN', 'log': 'WARN'  }, 'simulation': { 'tty': 'WARN', 'log': 'QUIET' }},
+{ 'gobyd':                                        { 'runtime': { 'tty': 'VERBOSE', 'log': 'WARN'  }, 'simulation': { 'tty': 'WARN', 'log': 'QUIET' }},
```
ferrom-jaia: Do we want this to be set to VERBOSE?

jason-jaia (approved):
Tested on bots in the lab - works as intended!

PR #196, opened 2022-09-09
Hub Status details in Central Command

edsanville:


ferrom-jaia (approved):
We tested in sim and on the hub!

PR #197, opened 2022-09-09
Bug/disappearing heading

ferrom-jaia:


jason-jaia (approved):
Tested on bot - works great.

PR #200, opened 2022-09-12
Xbee driver config

edsanville:
- Convert properly between big and native endian when communicating with the xbee
- Optional flag to enable node discovery (false by default)


src/bin/drivers/xbee/xbee.cpp
```diff
@@ -345,14 +342,14 @@ void XBeeDevice::process_frame_at_command_response(const string& response_string
...
 
     if (at_command == "SL") {
         assert(response->command_status == 0);
-        uint32_t lower_serial_number = *((uint32_t *)&response->command_data_start);
-        my_serial_number |= ((SerialNumber) lower_serial_number << 32);
+        uint32_t lower_serial_number = big_to_native(*((uint32_t*)&response->command_data_start));
+        my_serial_number |= ((SerialNumber)lower_serial_number);
         glog.is_verbose() && glog << "serial_number= " << std::hex << my_serial_number << std::dec << " node_id= " << my_node_id  << " (this device)" << endl;
 
         // Broadcast our node_id, and request everyone else's node_id
```
ferrom-jaia, line 355: I know that we ignore this broadcast, but is there a need to do it in the first place?

ferrom-jaia (approved):

ferrom-jaia (approved):
Code looks good, but we are not able to get it working.

src/bin/drivers/xbee/xbee_driver.cpp
```diff
@@ -87,18 +87,28 @@ int32_t decode_modem_id(const string& modem_id) {
...
 
+    glog.is_verbose() && glog << "Before device_.startup " << endl;
     device_.startup(driver_cfg_.serial_port(), driver_cfg_.serial_baud(),
-                    encode_modem_id(driver_cfg_.modem_id()), network_id);
+                    encode_modem_id(driver_cfg_.modem_id()), network_id, discover_peers);
+    glog.is_verbose() && glog << "After device_.startup " << endl;
 
-    for (auto peer : driver_cfg_.GetExtension(xbee::protobuf::config).peers())
+    for (auto peer : config_extension.peers())
     { device_.add_peer(peer.node_id(), peer.serial_number()); }
```
ferrom-jaia, line 110: Maybe we should convert big endian to little endian here

PR #213, opened 2022-09-22
bug/Reacquire-GPS

ferrom-jaia:
Added check incrementor for gps mode, Added discard location check when in a particular mode


jason-jaia (approved):
Matt explained to me and described how he tested in the simulator. Code has no glaring issues to me.

PR #218, opened 2022-09-27
bug/jaiabot-plot-depends

ferrom-jaia:
Fixing deps issues with jaiabot-plot and making all python apps use our virtual env


jason-jaia (approved):
Code review looks good.

PR #220, opened 2022-09-28
feature/Add-Drift-Duration-To-Dive-Parameters

ferrom-jaia:
Adding drift time to dive state parameters and fixed measure tool

Tested in sim


jason-jaia (approved):
Tested in sim

PR #221, opened 2022-09-29
changed default values of simulation

jason-jaia:
5 bots, no bot0, warp 5


ferrom-jaia (approved):
Nice! This is for Ian

PR #224, opened 2022-09-30
Logs show task packets

edsanville:
First run at showing those task packets.
Hover above the dive and drift markers to see the details as a tooltip.  Click the markers to bring up a popup.
Still using Leaflet at the moment, but Log Analyzer is going to need full OpenLayers functionality soon.
This version of the log analyzer is live on mercury, so that may be the easiest way to take a look at it.


src/web/server/jaia.py
```diff
@@ -11,6 +11,7 @@
 from time import sleep
 import datetime
 from math import *
+import task_mapper
```
ferrom-jaia: I am getting an error when I try to run this. 

`Traceback (most recent call last):
  File "./app.py", line 9, in <module>
    import jaia
  File "/home/ferrom/repos/jaiabot/src/web/server/jaia.py", line 14, in <module>
    import task_mapper
ModuleNotFoundError: No module named 'task_mapper'`

ferrom-jaia (approved):
Made a couple of changes to get the web side running, but the plot tool shows the task_packets when testing with sim

PR #225, opened 2022-10-01
feature/Reacquire-GPS-In-Transit

ferrom-jaia:
When bot is in transit state, we check gps hdop to check if it is below 1, if not we go into reacquire gps state.
Once we have reacquired gps we then go back into transit state.

The reason to reacquire during transit is sometimes the bot gets degraded gps, which causes the bot to drive in unexpected directions.

[Video](https://jaia-innovation.atlassian.net/wiki/spaces/PD/pages/917176321/ReacquireGPS+state+transition+while+in+Transit)


jason-jaia (approved):

jason-jaia:
Walked through the code with Matt and know he ran through simulator extensively.

PR #226, opened 2022-10-03
Mission planning large update.

oceanzus:


ferrom-jaia (approved):
Tested in sim!

A couple of bugs that are edge cases that we can talk about

PR #229, opened 2022-10-04
Bathymetric depth layer

edsanville:
CC now loads depth contours as GeoJSON, and styles them with a simple green -> red gradient.


edsanville:
Prints warning when all mesh points are colinear, and returns an empty GeoJSON for the contours.

edsanville:
Returns empty contour list, if all dives reached the same depth.

ferrom-jaia (approved):
Looks good! Tested in sim!

PR #231, opened 2022-10-04
shrunk buttons for tablet

jason-jaia:
Shrunk all buttons.
Changed inactive color to white (from green).
Added flag icon.
If someone can add the Material Design icon for Undo to the Undo button, that would be sweet.


ferrom-jaia (approved):
Looks good in sim!

PR #232, opened 2022-10-05
Mission Planning Survey Stats Update

oceanzus:
Adds survey stats to bottom of MP control panel.
Fixes blinking goals as max range.


ferrom-jaia (approved):
Looks good in sim!

PR #237, opened 2022-10-06
Mission Planning Bugs

oceanzus:
Fixes NONE mission goals.
Fixes rally point logic for less than 3 bots (line). 
Keeps area and perimeter at 0 for less than 3 mission points (line).


ferrom-jaia (approved):
Looks great! Tested in sim!

PR #238, opened 2022-10-06
feature/hub-icon-on-map

ferrom-jaia:
Added: hdop to botstatus, distance to hub status, simulated hub location, taskpacket bug, voltage coloring


jason-jaia (approved):
most of the code looks good - the rest I'm crossing my fingers on

PR #239, opened 2022-10-07
Mission Planning Update

oceanzus:
Adds first pass at line goal exclusion areas.
Cleans up MP interface.
Adds Rally Point Spacing input.


ferrom-jaia (approved):
Looks good! Tested in sim!

PR #242, opened 2022-10-07
Log Analyzer:  new log selection dialog box

edsanville:
Allowing filtering of logs by fleet and/or bot, and selection of multiple logs
Log list is sorted by time, so selecting logs for all bots for a mission should be simple


ferrom-jaia (approved):
Code looks good!

PR #247, opened 2022-10-07
Log date range - minor bug fix

edsanville:


jason-jaia:
No changes after merging with 1.y.

PR #249, opened 2022-10-09
task/refine-gps-and-imu-detections

ferrom-jaia:
Refined gps fix detection and imu issue detection


jason-jaia (approved):
Code review looks good.

PR #252, opened 2022-10-10
Removing vestiges of the Central Command name

edsanville:


ferrom-jaia (approved):
Tested in sim, searched for central command name, looked at changes. Looks good to me!

PR #253, opened 2022-10-11
feature/handle-bot-is-more-than-1km-from-dist

ferrom-jaia:
Bot dist to goal now indicates > 1000, dist to recovery now works, rc mode is fixed, fixed not displaying current information if the drift time is <= 0


michaelrockJAIA (approved):
basic review

PR #254, opened 2022-10-11
feature/recover-all-button

ferrom-jaia:
Added Recover all button, also changed undo button and added icon


michaelrockJAIA (approved):
basic review looks good

PR #261, opened 2022-10-26
Feature/health voltage

ferrom-jaia:
Added health warning and errors for low battery, very low battery, and critically low batter, they are configurable in health configuration


jason-jaia (approved):
Code review looks great

PR #263, opened 2022-10-26
Log openlayers

edsanville:
Converted to OpenLayers
Symbols changed to a new (cleaner?) set of SVG files


ferrom-jaia (approved):
Looks good! Tested with sim.

PR #264, opened 2022-10-27
task/Document-Health-App-Updates

ferrom-jaia:
Added document to show the timing for health warnings and errors


jason-jaia (approved):
lovely

PR #265, opened 2022-10-27
bug/Hold-Of-Zero-Task-Packet

ferrom-jaia:
Fixed bug where the depth achieved would be posted as zero on occasion, this bug was seen when setting hold = zero


jason-jaia (approved):
Code review looks good.

PR #266, opened 2022-10-31
battery voltage to % map in fusion

michaelrockJAIA:


ferrom-jaia (approved):
Code looks good. Saw a sim with code running.

PR #267, opened 2022-10-31
feature/Add-Radio-Config-To-Meta

ferrom-jaia:
Added ability to post serial number and node id of the radio in meta data


jason-jaia (approved):
Saw it working with mine own eyes

PR #268, opened 2022-11-03
JCC shared GUI

edsanville:
JCC now uses the shared OpenLayers GUI components from the Log Analysis tool.


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #270, opened 2022-11-08
bug/speed-protection

ferrom-jaia:
Using a slider to adjust speeds for mission, added comment for increasing speed, and a confirm popup if we do increase speed past max

[Demo](https://jaia-innovation.atlassian.net/wiki/spaces/PD/pages/949649421/Bot+Mission+Speed+Protection)


jason-jaia:
Looking good. Let's set Safety Limit at 3 m/s.
Add warning comment in jaiabot_pid_control.pb.cfg.in
Change rpm to throttle

PR #271, opened 2022-11-08
updated latent plots to data_vision

jason-jaia:
Not sure if I caught everything.


ferrom-jaia (approved):
Looks good, added health update

PR #272, opened 2022-11-09
Bug/speed protection

ferrom-jaia:


jason-jaia (commented):
I think the warnings should say 'may result' instead of 'will result'.  But otherwise I like it.

jason-jaia (approved):
Saw a nearly finished demo and it was pretty sweet.

PR #274, opened 2022-11-09
bug/camel-case-voltage

ferrom-jaia:
Fixing camel case to underscore for voltage vars


jason-jaia (changes requested):
JED is experiencing this issue as well. Good fix though!

jason-jaia (approved):
didn't run it myself, but judging by this outstanding work, it would hum

PR #277, opened 2022-11-09
bug/pitch-is-reporting-incorrectly

ferrom-jaia:
Flipping sign of pitch based on IMU orientation


jason-jaia (approved):
Code review results in satisfactory results.

PR #282, opened 2022-11-10
bug/forgot-to-add-change-log-1-y

ferrom-jaia:
The release has this already, merging into continuous1.y branch


jason-jaia (approved):
peachy keen, jelly bean

PR #286, opened 2022-11-10
feature/set-battery-levels-and-doc

ferrom-jaia:
Added new very low battery level and added documentation


jason-jaia (approved):
lookin' good in the neighborhood

PR #291, opened 2022-11-15
bug/wrong-health-error-plot

ferrom-jaia:
I added in the wrong health enum by mistake...

patch fix 1.1.1 release-1.


michaelrockJAIA (approved):
1 line bug fix

PR #292, opened 2022-11-15
bug/wrong-health-error-plot

ferrom-jaia:
I added in the wrong health enum by mistake to release...

Need to update change log

patch fix 1.1.1_beta0 1.y.


jason-jaia (approved):
makes sense

PR #293, opened 2022-11-16
Feature/jdv time slider

edsanville:
Time slider for JDV, with live data table under plots


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #294, opened 2022-11-16
feature/version-control-arduino

ferrom-jaia:
Added version control to the arduino code.

The arduino driver will not be able to send commands if the jaiabot-embedded version is not compatible with the arduino code.

A health error will be thrown if the versions are not compatible.


ferrom-jaia:
Tested on fleet2

PR #295, opened 2022-11-17
bug/jed-speed-warn-and-imu-port

ferrom-jaia:
* Fixed IMU port switching
* Added warning in JED for throttle control


jason-jaia (approved):
code review looks good

PR #297, opened 2022-11-17
Feature/jdv csv download

edsanville:
Added CSV download button, re-designed the plot pane's interface, added ability to delete charts


ferrom-jaia (approved):
Tested in Sim! I like it!

PR #300, opened 2022-11-28
feature/gps-req-transit-aft-dive

ferrom-jaia:
Added different requirements for gps during transit and after a dive


jason-jaia (changes requested):
Please change increment counter to start at 0.

jason-jaia (approved):

PR #301, opened 2022-11-29
feature/watch-bot-batt-percentage

ferrom-jaia:
We can detect low voltage and low battery or just one of them


jason-jaia (approved):
Code review looks good.

PR #302, opened 2022-11-30
Fix/goal edit box

edsanville:
Don't display goal editing box after a mission deletion, or an undo operation, because the goal may have been deleted.


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #303, opened 2022-11-30
Add tooltips to all buttons

edsanville:


ferrom-jaia (approved):
Looks good! Tested in sim!

PR #304, opened 2022-12-01
Fixed camel-case issues in JED

edsanville:
bot_id wasn't snake-case, so was coming through as undefined.  Meanwhile, status age variables were snake-case, but should still be camel-case.


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #305, opened 2022-12-02
Remember dive/drift parameters after editing goals

edsanville:


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #306, opened 2022-12-02
Feature/jcc botdetails number hotkeys + SHIFT-# to stop a bot

edsanville:


ferrom-jaia (approved):
This works for bots 0 - 9. We will have to see how to get bots greater than 9 to stop using a shortcut as well. Tested in sim.

PR #309, opened 2022-12-05
Feature/jcc active mission layer

edsanville:
JCC shows a layer with the active missions at 50% opacity


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #310, opened 2022-12-06
added i2c table and led tables and pictures

jason-jaia:


src/doc/figures/power-board-leds.png:Zone.Identifier
```diff
@@ -0,0 +1,3 @@
+[ZoneTransfer]
```
ferrom-jaia: Do we need this file?

src/doc/figures/logic-board-leds.png:Zone.Identifier
```diff
@@ -0,0 +1,3 @@
+[ZoneTransfer]
```
ferrom-jaia: Do we need this file?

jason-jaia: No - I have no idea what it is.

ferrom-jaia (approved):
Sweet! Reviewed.

PR #311, opened 2022-12-07
Feature/jcc incontrol

edsanville:
Red border around screen in JCC and JED if another client has control.  JCC confirms taking control with a dialog, and JED has a "Take Control" button.


ferrom-jaia (dismissed):
Looks good! Tested in sim.

ferrom-jaia (approved):
Looks good! Tested in sim.

PR #312, opened 2022-12-09
JED migrate from cookies to localStorage

edsanville:


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #316, opened 2022-12-12
Task/jdv typescript

edsanville:
Converted the shared map symbol library and JDV to TypeScript for maintainability and scalability.
Updated the node and npm versions used for JDV and JCC
Use nvm to precisely manage the node version during build


ferrom-jaia (approved):
Tested in Sim. Fixes current bug that we are seeing with jdv (entries errors).

PR #317, opened 2022-12-13
Added some missing requirements to JDV and the main requirements.txt

edsanville:


src/python/requirements.txt
```diff
@@ -11,3 +11,8 @@ rpi-ws281x==4.3.4
 RPi.GPIO==0.7.1
 smbus==1.1.post2
 sysv-ipc==1.1.0
+flask==2.2.2
```
ferrom-jaia: Do we need this requirement here? We have it in the web/requirements.txt

ferrom-jaia (approved):
Updated to include matplotlib

PR #318, opened 2022-12-15
bug/JAIAB-189-fix-bugs-found-during-1-3-0-beta-testing

ferrom-jaia:
Bugs:
Hub details undefined
Bot detail buttons do not work on predeployment failed
Shortcuts should not work while we type saving mission
Warning for max points
The shortcut for bot details does not save the accordian settings (open/close)
Multihold dive
PID depth


jason-jaia (changes requested):
PID gains should always be positive. Dive code needs to be reviewed so this is true rather than changing the PIDControl message.

jason-jaia (dismissed):
Maybe it would be a little clearer to say something like: "Adding this goal exceeds the limit of MAX_GOALS"

src/web/command_control/client/style/CommandControl.less
```diff
@@ -1174,16 +1174,12 @@ div.closeButton {
   padding: 8pt;
   border-radius: 4pt;
 
-  button {
-    color: white;
-    background-color: #06c006;
+  #takeControlButton {
+    color: black;
```
jason-jaia: I wonder if this color should be a @button-text or something similar. Not sold on it, but want to make sure it's considered.

michaelrockJAIA (approved):
matt tested and approving to test on the hardware

PR #319, opened 2022-12-15
JAIAB-190-task-rename-plot-directory-to-jdv

ferrom-jaia:


jason-jaia (approved):
Code review looks good although it might not have been super thorough.

PR #328, opened 2022-12-28
Task/jcc typescript

edsanville:
JCC is fully TypeScript, although some parts of Jim's code have the "any" type, because I couldn't figure out the exact type names.  I leave it to him to correct these annotations at his convenience.


ferrom-jaia:
Thanks for fixing the conflicts Ed!

I noticed that the bot list color is orange. Not sure if this was intended or not. 
I also noticed that we can no longer turn on/off NOAA ENC charts or GEBCO Bathymetry

edsanville:
Ok, conflicts should be resolved, and the two issues are fixed.  I merged in the latest 1.y branch.

> Thanks for fixing the conflicts Ed!
> 
> I noticed that the bot list color is orange. Not sure if this was intended or not. I also noticed that we can no longer turn on/off NOAA ENC charts or GEBCO Bathymetry

ferrom-jaia (approved):
Looks good! Tested in sim.

PR #332, opened 2023-01-10
feature/query-engineering-status

ferrom-jaia:
Instead of bot sending engineering status every 5 seconds, the bot should only send the status when queried


michaelrockJAIA (approved):
approving to test on the bench

PR #333, opened 2023-01-11
feature/adjusted-pressure-message

ferrom-jaia:
We now use an pressure adjustment when we begin a dive


michaelrockJAIA (approved):
Explained by Matt. Decent code review - logic seems sound. Going to test on a bot

PR #336, opened 2023-01-16
bug/potential-race-condition-adjusted-pressure

ferrom-jaia:
Fixed potential race condition with setting pressure before a dive to adjust

bug-fix-pressure


jason-jaia (approved):
Code review looks good.

PR #337, opened 2023-01-16
changed to loop

jason-jaia:
changed to loop to add output on the successes (as well as the failures we were only seeing)


ferrom-jaia (approved):
Looks good!

PR #339, opened 2023-01-17
feature/increase-manual-control

ferrom-jaia:
increasing pid control app and aruino app loop speeds to 10 Hz, this will increase manual control performance


michaelrockJAIA (approved):
easiest review. tested in duck

PR #340, opened 2023-01-19
Feature/speed window

edsanville:
Added a new config parameter for the jaiabot_pid_control app as follows:
```    // The threshold heading_error below which the PID algorithm will target full desired speed
    optional double full_speed_window = 33
        [default = 30, (dccl.field).units = {
                derived_dimensions: "plane_angle"
                system: "angle::degree"
                }];
```


ferrom-jaia (approved):
Looks good!

PR #343, opened 2023-01-22
bug/fix-heading-flip-when-bot-rolls-over

ferrom-jaia:
Fixed bot heading flip when bot rolls over


michaelrockJAIA (approved):
I wrote and tested it with matt

PR #345, opened 2023-01-25
Added a button to download map features as KML file

edsanville:


michaelrockJAIA (commented):
zooms correctly when 1 file is exported but zooms to south America when two are selected

ferrom-jaia (approved):
Does what we want!

PR #346, opened 2023-01-25
feature/hub-on-map-and-command-for-hub

ferrom-jaia:
Can restart, restart services, and shutdown hub. We also see the hub on the map now.


michaelrockJAIA (approved):
tested otw

PR #347, opened 2023-01-25
bug/build-failed-during-nvm-install

ferrom-jaia:
Fix for nvm race condition


michaelrockJAIA (approved):
seems as though this just moved a block of code

PR #348, opened 2023-02-02
bug/build-fails-fiona-version

ferrom-jaia:
Updated requirements txt in src/web so that fiona is kept at version1.8.21 for now


michaelrockJAIA (approved):
1 liner

PR #351, opened 2023-02-03
Fallback to USER environmental variable when no login

tsaubergine:
Needed for preseed headless install.


ferrom-jaia (approved):
Looks good!

PR #352, opened 2023-02-03
IMU Quaternions

edsanville:
Use quaternions from the IMU to calculate Euler angles, and report those directly from the python driver


src/python/adafruit_BNO055/jaiabot_imu.py
```diff
@@ -85,71 +130,68 @@ def getData(self):
...
+    while True:
+
+        data, addr = sock.recvfrom(1024) # buffer size is 1024 bytes
+
+        current_time = time.time()
+
+        if (previous_time + imu_timeout) <= current_time:
+            previous_time = current_time
+            print("reset imu")
+            imu.reset()
```
ferrom-jaia: Do we still want to reset?
The reason I ask is if think quarernions fix our imu issues it would be good to comment this out for now to test.

src/python/adafruit_BNO055/jaiabot_imu.py
```diff
@@ -70,11 +111,15 @@ def setup(self):
         pass
 
     def getData(self):
+        quaternion = (1, 0, 0, 0)
```
ferrom-jaia, line 115: Can we add a comment to describe what these inputs are?

src/python/adafruit_BNO055/jaiabot_imu.py
```diff
@@ -43,16 +77,23 @@ def setup(self):
             self.sensor.mode = adafruit_bno055.NDOF_MODE
             self.is_setup = True
 
+            self.sensor.axis_remap = (0, 1, 2, 1, 1, 0)
```
ferrom-jaia, line 81: Can we add a comment to describe what these inputs are?

src/python/adafruit_BNO055/jaiabot_imu.py
jason-jaia: Good point - I commented it out.

src/python/adafruit_BNO055/jaiabot_imu.py
jason-jaia: Done

jason-jaia (approved):
Tested on flatbot through to JCC and JED.

PR #354, opened 2023-02-13
task/jdv-use-node-status-instead-of-botstatus

ferrom-jaia:
Changed from using bot status to node status


jason-jaia (approved):
Code review looks good.

PR #355, opened 2023-02-16
Fix goal options component to NOT share parameters between goals (dee…

edsanville:
…pcopy the goal parameters)


ferrom-jaia (approved):
Looks great! Tested in sim and it worked!

PR #356, opened 2023-02-16
feature/program-led-hub-button

ferrom-jaia:
Added python code to handle the hubs led button lights and button presses

LED solid red while pi is off and initial booting of pi
LED is blinking green while services are starting up
LED is blinking red while the pi is shutting down (2 button clicks)
LED blinks red and green when stopping all bots (0.5 button hold)
LED is solid green when all of the services are running.

Also requires:
https://github.com/jaiarobotics/jaiabot-debian/pull/23

Used by:
https://github.com/jaiarobotics/jaiabot-rootfs-gen/pull/41


src/python/requirements.txt
```diff
@@ -11,3 +11,4 @@ rpi-ws281x==4.3.4
 RPi.GPIO==0.7.1
```
jason-jaia: I think we want to remove this callout as we discussed, since gpiozero has this as a dependency already.

jason-jaia:
The activity diagram needs to be updated to make a little more sense (not really using decisions).

jason-jaia:
Can you add that when the light isn't on, all the electronics in the hub are off?

ferrom-jaia:
Tested with new hub.

PR #357, opened 2023-02-17
bug/toast-need-to-be-moved

ferrom-jaia:
Cleaned up ui tablet flipping button, moved location of the toasts to be at the bottom, cleaned up configuration files, goal timeout feature is off by default since it is still in beta form.


jason-jaia (approved):
Code review looks good.

PR #360, opened 2023-02-21
feature/goal-no-forward-progress-second-pass

ferrom-jaia:
We now have the ability to use linear regression line slope for goal timeout.

Default is history size 150 (30 seconds best case), eps is 0.1

x = time, y = dist

When history gets to 150 then we drop oldest and add newest to back.

We have the ability to use both the linear regression line slope and the timer to get a goal timeout if we want to.


ferrom-jaia:
TODO: Do not calculate linear regression at the 5hz speed that we get gps at. We should check at 1 hz.

PR #363, opened 2023-02-24
feature/update-no-heading-response

ferrom-jaia:
Detect no heading data quicker and restart imu


jason-jaia (approved):
Code review looks good - I also know Matt tested in sim.

PR #364, opened 2023-02-27
task/imu-restart-variable-updates

ferrom-jaia:
Changed corrected heading to corrected orientation in imu python driver
Added comments and changed config names for imu detection


src/python/adafruit_BNO055/jaiabot_imu.py
```diff
@@ -168,12 +168,12 @@ def do_port_loop():
         gravity = data['gravity']
         calibration_status = data['calibration_status'] # 1 is calibrated, 0 is not
         bot_rolled = int(abs(euler.roll) > 90) # Did we roll over?
-        corrected_heading = euler.heading + 90
+        corrected_orientation = euler.heading + 90 # correct orientation since we have a 90 degree offset
```
jason-jaia: I'd really like us to do the math here to correct for headings over 360.  Also, PEP guidelines advise against inline comments.

jason-jaia (commented):
looks good aside from comments made

src/python/adafruit_BNO055/jaiabot_imu.py
ferrom-jaia: This is going to be resolved in the wave height detection code

PR #365, opened 2023-02-28
bug/bot-no-gps-crashes-jcc

ferrom-jaia:
If bots do not have a gps fix it would crash jcc, Added to check to see if the geom is finite.


jason-jaia (approved):
Code looks good - Matt explained it to me.

PR #366, opened 2023-03-01
JAIAB-21 Systemd sim

tsaubergine:
See https://jaia-innovation.atlassian.net/jira/software/projects/JAIAB/boards/2/backlog?assignee=557058%3A3760d625-55bf-42ee-8cf7-1fa98e11d936&selectedIssue=JAIAB-21

Update systemd.py to include `--simulation` flag which, if set, will generate service files and other configuration to run the bots and hubs in simulation. This allows the use of the Virtualbox fleet (added in https://github.com/jaiarobotics/jaiabot-rootfs-gen/pull/29) in simulation mode for more accurate development scenarios where the standard "all on one computer" simulation *.launch files don't suffice, such as:
- Customer demos
- Customer training
- Development that includes requires realistic networking (data offload, etc.)
- Development for testing systemd services

Other uses for this could be load testing on the real bots without hardware connected, in-lab tests, etc.

Also requires:
- https://github.com/jaiarobotics/jaiabot-debian/pull/20

Used by:
- https://github.com/jaiarobotics/jaiabot-rootfs-gen/pull/35


ferrom-jaia (approved):
Looks good! Saw this working in Toby's video review.

PR #371, opened 2023-03-17
feature/add-mean-temp-and-salinity-to-kmz

ferrom-jaia:
Added mean depth, temp, and salinity to kmz


jason-jaia (approved):
Seems ok to me w/o testing it.

PR #372, opened 2023-03-21
Fixed bug where bot status panel close button was in wrong location

michael-jaia:
Overrode "margin: auto;" set on all divs by adding "margin: 0;" to the more specific selector: div.closeButton


jason-jaia (approved):
code review looks good

PR #374, opened 2023-03-22
feature/mission-creation-panel

ferrom-jaia:
Added a mission creation panel. It is at the same baseline as current tool, but with the added benefit of assigning/unassigning runs. We can also duplicate missions as well.


ferrom-jaia (commented):
Tested in the field. This is a baseline. Tickets were added to follow up on any features that are needed.

PR #377, opened 2023-03-27
Fix hub selection highlight bug in BotListPanel

michael-jaia:
Added selectedHub to state becasue the active hub info needed to be separated from detailsBoxItem which holds the info for the most recent click in the BotListPanel. Otherwise, if the user clicked (in the panel) the hub, then a bot, and then the hub again, React would register it as the hub being clicked on, not off.


michael-jaia:
Restarted on branch: bug/hub-detail-button-does-not-highlight-1

PR #378, opened 2023-03-28
Update build_messages.sh to not hardcode Goby/DCCL paths

tsaubergine:
Required when building with -DGOBY_DIR=/some/path and/or -DDCCL_DIR=/some/path that is not the default system install (/usr/include).


ferrom-jaia (approved):
Looks good!

PR #379, opened 2023-03-29
bug/CMakelist-fails-with-x-y-version-change

ferrom-jaia:
Fixed issue with build failing locally with new change to use circleci to automatically correct version of project


jason-jaia (approved):
Code review seems good.

PR #380, opened 2023-03-29
feature/send-task-packets-to-customer-port

ferrom-jaia:
Added another udp config in web portal to send specific data to a customer


michaelrockJAIA (approved):
tested in Sim

PR #381, opened 2023-03-29
Bug/hub-detail-button-does-not-highlight-1

michael-jaia:


ferrom-jaia (commented):
Can we deselect the bots mission when we select the hub? I can see this being confusing to the user.

ferrom-jaia (changes requested):
Can we deselect the bots mission when we select the hub? I can see this being confusing to the user.

michael-jaia:
Moved to branch bug/hub-detail-button-does-not-highlight-2

PR #382, opened 2023-03-30
Feature/add-delete-button-in-bot-detail-commands

michael-jaia:


ferrom-jaia (changes requested):
Hey Michael, this no longer works when I merged 1.y into it. I think the version you were using was before the new mission creation panel got merged in. This will probably need another implementation.

PR #384, opened 2023-04-02
Tweaking the constant heading task panel

edsanville:


ferrom-jaia (changes requested):
Looks good Ed just a couple of things. 

1. The speed should be limited to 3 m/s.
2. I got this error when I did a clean build:

I ran the run.sh script in src/web
```
ERROR in ../../jdv/client/src/gui/MissionFeatures.ts 33:15-36
Module not found: Error: Can't resolve '@turf/turf' in '/home/ferrom/repos/jaiabot/src/jdv/client/src/gui'
Did you miss the leading dot in 'resolve.extensions'? Did you mean '[".*",".js",".jsx",".ts",".tsx"]' instead of '["*",".js",".jsx",".ts",".tsx"]'?
 @ ./client/components/CommandControl.tsx 73:37-69
 @ ./client/App.js 3:0-57 16:90-104
 @ ./client/index.js 3:0-24 6:46-49

ERROR in /home/ferrom/repos/jaiabot/src/jdv/client/src/gui/MissionFeatures.ts
../../jdv/client/src/gui/MissionFeatures.ts 9:42-54
[tsl] ERROR in /home/ferrom/repos/jaiabot/src/jdv/client/src/gui/MissionFeatures.ts(9,43)
      TS2307: Cannot find module '@turf/turf' or its corresponding type declarations.
 @ ./client/components/CommandControl.tsx 73:37-69
 @ ./client/App.js 3:0-57 16:90-104
 @ ./client/index.js 3:0-24 6:46-49

webpack compiled with 2 errors
```
I had to run make in the jdv directory then run the run.sh script in src/web to get things running 

4. When we run the run script I like the output. I no longer saw the pinging server message. Can we add this back to the output?
5. We found an edge case bug doing testing
Steps to reproduce: Click on goal -> select constant heading on map -> do not close the goal settings -> click on another goal -> then click cancel.

This will jump the second goal you clicked on to the lat lon of the first point.

6. Can we make the dashes a little longer than they currently are?

Thanks Ed!

edsanville:
I believe I've addressed all the issues.  Merging was tough, due to some changes I made to the way JDV and JCC build.
Notes:
*  I moved JDV to src/web/jdv
*  I'm installing a lot of shared JDV/JCC dependencies in the src/web directory (which of course go into src/web/node_modules)
*  The "gui" library is now moved to src/web/shared, (it has JAIA protobuf message interfaces for TypeScript AND the styles and features for OpenLayers, both required by JDV and JCC)
* JDV and JCC now depend on the "npm_dependencies" target in CMake, to make sure their shared dependencies get installed in src/web/node_modules before either app gets built

edsanville:
I think, long-term,  we might want:

* JDV and JCC to be two apps built by the same package.json.
* Build our web apps WITHIN the build directory, so all those node_modules get installed in the build directory.  This will allow us to easily do a clean build by `rm -rf build/*`

ferrom-jaia:
I think we need another pull request in jaiabot-debian that takes into consideration jdv is not longer in the same location. I think we also need to update the CMakeList.txt to put jdv in the build directory.

ferrom-jaia (dismissed):
Looks good. Tested in Sim.

ferrom-jaia (approved):
Looks great! Tested in sim.

PR #387, opened 2023-04-04
Feature/bottom type characterize

edsanville:
First pass: enclose the max_acceleration in the DivePacket


edsanville:
Changed amplitude to height in the Component class

PR #388, opened 2023-04-04
task/add-wave-and-bottom-type-fields-into-task-packet

ferrom-jaia:
Added significant wave height, wave height, wave period, and bottom type to task packet dccl message


michaelrockJAIA (approved):
matt tested

PR #389, opened 2023-04-05
task/set-major-minor-patch-local-build

ferrom-jaia:
Setting major, minor, patch version when building project locally


jason-jaia (approved):
built locally with no issues.

PR #391, opened 2023-04-05
task/change-index-of-goal-to-start-with-1

ferrom-jaia:
Changed the run wpt to start at 1


jason-jaia (approved):
Tested locally and works nicely.

PR #393, opened 2023-04-05
Highlight hub when selected

michael-jaia:


ferrom-jaia (approved):
Looks great! Thanks man!

PR #394, opened 2023-04-05
feature/add-ability-to-use-controller-for-remote-control

ferrom-jaia:
Tested with an xbox controller


michaelrockJAIA (approved):
tested in sim and on bench

PR #395, opened 2023-04-05
Task/add-button-states-for-open-close

michael-jaia:
Close other view control panels when a new view control is selected


ferrom-jaia (approved):
Looks great!

PR #397, opened 2023-04-06
task/hide-power-buttons-in-a-accordion

michael-jaia:
Place bot power buttons in "Advanced Commands" nested accordion


ferrom-jaia (approved):
Looks great! I added new data accordion that holds gps, imu, sensor, and power accordions. I also updated the data retry button so it uses the same icon as the recover command.

PR #398, opened 2023-04-06
Feature/alert-operator-when-runs-are-deleted

michael-jaia:


ferrom-jaia (approved):
Looks good! Added confirmations around the delete buttons in the mission panel.

PR #399, opened 2023-04-10
bug/active-mission-and-new-mission-show-duplicate-active-goal

michael-jaia:


ferrom-jaia (approved):
Looks good! Need to look into resetting active goal index quicker/ disabling it when a new mission is sent. This covers most of the confusion though.

PR #400, opened 2023-04-11
Fix deb versioning circleci

tsaubergine:
The old script that determined which Debian APT repo to push packages to had a bug that didn't set the correct repo version when building for tagged releases. This bug was exposed when I added the new test/X.y repo for non-1.y test builds.

I moved the logic into a standalone script (`.circleci/deb_repo.sh`) and created a unit test for it (`test_deb_repo.sh`). I fixed the script logic so it gets the version from:

- For tagged releases, the tag major and ".y" -> 1.4.0 -> "1.y". For tags that have an underscore (e.g. 1.4.0_beta2) use the "beta" repo, otherwise use the "release" one.
- For branches that match [0-9]+\.y (e.g. 1.y, 2.y, etc.) use that branch name. Otherwise use test/X.y

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,25 +52,8 @@ job-template-amd64: &job-template-amd64
     - run: &run-set-deb-rep-and-version
         name: Set environmental variables for which packages.jaia.tech repo to use
         command: |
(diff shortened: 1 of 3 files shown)


ferrom-jaia (approved):
Looks great! Thanks Toby!

PR #405, opened 2023-04-12
feature/more-alerts-and-confirmations

michael-jaia:
* Alert operator what bot they are powering off, restarting, or restarting services for

* Confirm when operator hits all stop, all next task, and all undo btns


src/web/command_control/client/components/CommandControl.tsx
```diff
@@ -2247,6 +2247,8 @@ export default class CommandControl extends React.Component {
 	}
 
 	restoreUndo() {
+		if (!confirm('Click the OK button to go back to the previous task for all mission-active bots:')) return
```
ferrom-jaia: Instead of:
Click the OK button to go back to the previous task for all mission-active bots:
Use:
Click the OK button to undo the previous run edit that was made:

ferrom-jaia (approved):
Looks good! Tested in sim!

PR #406, opened 2023-04-12
bug/cmake-fails-with-strip

ferrom-jaia:
fixing cmake error with stripping a variable that doesnt exist


jason-jaia (approved):
code review looks good

PR #412, opened 2023-04-13
feature/data-offload-progress

ferrom-jaia:
This adds a percentage to the data offload state so we can watch the progress. Needs to be tested on bots and hubs.


michaelrockJAIA (approved):
worked on bench

PR #414, opened 2023-04-14
bug/change-gps-checks-to-reflect-new-output-speed

ferrom-jaia:
Added ability to update gps checks based off electronics stack

This is important because on electronics stack 1 and 2 we use the gps to output at 5hz and on electronics stack 0 we output at 1hz.

This would mean the logic would not be suitable for electronics stack 1 and 2. We do not want to make a change that breaks stack 0 though.

We know what electronics stack each bot is using so by utilizing this information we set the gps check values.


michael-jaia (approved):
Looked over code and Matt tested

PR #417, opened 2023-04-20
Fix regressions caused by use hub instead of hub0 as data offload target

tsaubergine:
Fixes regression caused by https://github.com/jaiarobotics/jaiabot/commit/9d06c6ab39b6a8fbf2c5f56441cb3055e61f0923

See https://github.com/jaiarobotics/private_jaia/pull/17#issuecomment-1517024789


jason-jaia:
code review looks good

jason-jaia (approved):

PR #421, opened 2023-04-25
task/change-min-depth-safety-config

ferrom-jaia:
Changing the min safety config for bottom depth to 0 by default (Turning Off)


jason-jaia (approved):
Code review makes sense

PR #422, opened 2023-04-25
bug/stop-detecting-imu-when-motor-is-off

ferrom-jaia:
removing false detections for imu and course over ground warnings


jason-jaia (approved):
Code review makes sense

PR #423, opened 2023-04-25
task/slow-down-botstatus-poll

ferrom-jaia:
Slowed polling interval back to 500 ms


jason-jaia (approved):
Code review makes sense

PR #424, opened 2023-04-27
task/update-comms-settings

ferrom-jaia:
Changing comms configuration and updating detecting imu issues


jason-jaia (approved):
Tested in the field and worked.

PR #427, opened 2023-04-28
Allow user to setup an End Task while configuring a Survey Mission

edsanville:


ferrom-jaia (approved):
Looks good! Made changes to the defaults (station keep to NONE)

PR #429, opened 2023-05-01
Task/task-packets-should-be-defaulted-to-off-in-jcc

michael-jaia:
* Fixed bug/cannot-create-two-tasks-of-same-type-if-close-btn-not clicked
* Fixed bug/no-confirmation-on-download-all
* Completed task/make-width-of-engineering-panel-btns-uniform
* Completed task/measurements-should-be-defaulted-to-off-in-jcc
* Restyled map layers panel to match other jcc panels


ferrom-jaia (dismissed):
Looks great!

ferrom-jaia (approved):
Looks great!

PR #185, opened 2022-08-29
CC Fixes mission botId list to use actual botId from podStatus.bots

oceanzus:
image

PR #433, opened 2023-05-02
bug/goal-settings-and-optimize-mission-panel

michael-jaia:
React Error #185:
Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.

GoalSettings component had an onChange event handler called in componentDidUpdate which called updateMissionLayer. When the line tool created a mission, the number of state calls made from creating a mission with the line tool and from the GoalSettings component exceeded the React max depth.


ferrom-jaia (approved):
Nice fix and explanation!

PR #435, opened 2023-05-02
feature/add-bottom-depth-safety-task

ferrom-jaia:
Added ability to dynamically set minimum bottom depth safety in engineering panel


michaelrockJAIA (approved):
need to test on the water

PR #437, opened 2023-05-04
task/update-data-offload-script-to-move-files-to-archive

ferrom-jaia:
Updated the offload script so we mv the logs after each offload to an archive directory so we get an accurate rync percentage


jason-jaia:
Doesn't handle symlinks for latest files correctly. I think it probably is best to just delete them as the goby log tool remakes them, but that should be verified.

ferrom-jaia (commented):
Jason Tested, but cannot approve because he is currently driving to NH

PR #438, opened 2023-05-04
Feature/automate kmz generation for task packets

edsanville:
Fixed two crashing bugs, and create venv for jdv


ferrom-jaia (approved):
Looks great! Thanks for updating the deploy script as well!

PR #440, opened 2023-05-05
task/change-font-style-of-map-layers-panel-to-match-JCC

michael-jaia:


jason-jaia (approved):
Checked locally - seems good.

PR #441, opened 2023-05-08
task/increase-comms-slot-time

ferrom-jaia:
The increased speed messed up the subscriptions between the hub and the bots


jason-jaia (approved):
Code review ok

PR #443, opened 2023-05-11
feature/add-xbee-encryption

ferrom-jaia:
Added ability to turn on encryption using xbee

I put a template xbee_encryption.pb.cfg in jaiabot/config/templates

Move this file to /etc/jaiabot to configure encryption

### DIFF
--- a/config/gen/bot.py
+++ b/config/gen/bot.py
@@ -92,11 +92,6 @@ try:
 except FileNotFoundError:
     xbee_info = 'xbee {}'
 
-try:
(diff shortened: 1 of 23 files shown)


jason-jaia (changes requested):
Please add a comment for the password length.  Also please use the common.py where appropriate.  Also please test with the older radio as well.

ferrom-jaia (commented):
Looks great!

jason-jaia (approved):
Changes are excellent

PR #444, opened 2023-05-15
task/make-global-font-variable

michael-jaia:


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #445, opened 2023-05-15
task/add-jdv-link-to-jcc

michael-jaia:


ferrom-jaia (changes requested):
After thinking about this, it will not work. We need to point to the hub ip address.

10.23.X.X

We have the hub id so we can figure out the last part 

10 + hub id

10.23.X.10 + hubid

We just need the fleet number.

ferrom-jaia (approved):
Looks great! Tested in sim and on a hub.

PR #446, opened 2023-05-17
jaiabot_log_converter processes .taskpacket files into .kmz files

edsanville:


ferrom-jaia (approved):
Looks good! Tested on my laptop.

PR #447, opened 2023-05-17
feature/create-task-packet-file-on-bot

ferrom-jaia:
This will allow laconic to get the data quicker because they only care about the task packet information.

This other pull request is needed to convert the taskpacket file to a kmz.

https://github.com/jaiarobotics/jaiabot/pull/446


jason-jaia (approved):
Code review / walkthrough seems legit.

PR #448, opened 2023-05-17
Jaiab 389/fix multihub restart

tsaubergine:
See https://jaia-innovation.atlassian.net/browse/JAIAB-389

1. Removes persistent subscriptions on the bot.
2. Removes unsubscribes from hub to the bot (as these are no longer necessary as the bot will clear its subscriptions on restart: see 1.)
3. jaiabot_hub_manager now always responds with subscribe from the hub to the bot when we receive a subscribe from the bot to the hub (not just when we get a "new" bot. as this could be a restarted bot...)
4. xbee driver persists the last hub on the bot and use this to subscribe on restart


ferrom-jaia (approved):
Looks good! Tested on flat fleet. Looked over code. Diagram super helpful! 

Merging into 1.y. Need to do field testing with the jaia fleet.

PR #449, opened 2023-05-17
bug/fix-data-offload-script-with-task-packet-arg

ferrom-jaia:
Fixed bug with not excluding goby files when in task packet file only mode


ferrom-jaia:
Tested with hub

PR #452, opened 2023-05-22
task/re-style-rc-panel-making-dual-control-default

michael-jaia:


ferrom-jaia (approved):
Looks great! Added directional output for xbox controller use case

PR #453, opened 2023-05-22
task/move-rc-button-to-top-toolbar-in-bot-details

michael-jaia:


ferrom-jaia (commented):
We do not want to do this anymore :(

Can you move constant heading change to the rc pull request?

PR #455, opened 2023-05-24
bug/fix-endpoint-typo

ferrom-jaia:
fixing typo with from tansit -> transit


michael-jaia (commented):
I think there may be one more instance of "tansit" in the return statement on line 237

ferrom-jaia:
> I think there may be one more instance of "tansit" in the return statement on line 237

Nice catch!! Thanks man!

michael-jaia (approved):

PR #456, opened 2023-05-24
Task/refractor command control 1

edsanville:
Refactored map, layer, and interaction creation code out of CommandControl.tsx


michael-jaia (commented):
The abstractions are great! I was wondering if we should change Layers.ts to a file of getters for each map layer instead of accessing them directly through the class. Or maybe we drop the constructor to prevent multiple ways of accessing the layers. With the current implementation, we don't need the "all" property in the constructor, so dropping it would prevent someone from changing a layer by writing "layers.all[index]..." I think this would help with future debugging and controlling the way in which layers are modified. Getters are probably the "most secure", and if there was a situation where we wanted to update all layers, we could create a setter that does so. These are just a couple of ideas; let me know what you think.

edsanville:
Good points.  I can revisit this and use getters that get a private singleton instead.

If we ever need multiple Maps, (perhaps in some dialog box or something), then we can do it a different way.

ferrom-jaia:
Closing pull request because task refractor command control 2 was merged in which included the changes in this pull request.

PR #458, opened 2023-05-26
Check the mtime of the .taskpacket versus .kmz file before converting

edsanville:
It was re-converting the taskpacket files every 30 seconds, even when unchanged


ferrom-jaia (approved):
Good catch!

PR #460, opened 2023-05-26
Feature/throttle goby log file

edsanville:
Added a new jaia_goby_log_level environment variable, which defaults to RELEASE.
RELEASE mode will filter out MOOS messages from the .goby files, saving large amounts of disk space (some 90%)
DEBUG mode will not filter out those messages, and should only be used when debugging the MOOS-Goby interface


ferrom-jaia (approved):
Looks great! We may want to add an option in jaiabot-debian for the user to choose. Tested in sim.

PR #464, opened 2023-06-01
task/limit-number-of-waypoints

ferrom-jaia:
Limit the number of wpts to 15 because of a known bug


jason-jaia (approved):
Code review with Matt - looks good.

PR #465, opened 2023-06-01
bug/fix-arduino-service-description-for-gpio-pin

ferrom-jaia:


ferrom-jaia:
Merging in without approval because it is just a description change

PR #466, opened 2023-06-04
Use existing entries in runtime.env, so we don't clobber things like fleet and bot ID when we re-deploy

edsanville:
… don't overwrite on re-deploy


edsanville:
We could do this with a .env file like a "local.env" file.  

Disadvantage:
*  A json file can include much more complicated and nested information, if we want to expand this configuration file to other info

We could just avoid overwriting the runtime.env file.

Disadvantage:
*  If we want to change some of the other stuff, like jaia_log_dir or something, it wouldn't get updated if there's an existing "/etc/jaiabot/runtime.env"

jason-jaia:
I've seen this type of thing get out of hand in the past.

Perhaps a nice way to handle it would be to have an update script that only adds information, it doesn't rewrite any values (although I suppose it could if you agreed).

We could eventually make it smart and add little #new comments to new lines the first time they're added and maybe erase them later.

Another idea I've had in the past is keeping a version number in the config file so you know you have the latest information specified.

Up for any and all ideas.

edsanville:
You're right, and I was semi-cringing when I went with the idea TBH.  I think only adding and not updating will work for our purposes.  I will update this PR to go with that strategy.

edsanville:
Done!

ferrom-jaia (approved):
Looks great! Tested on hub 1 fleet 0

PR #472, opened 2023-06-09
feature/allow-download-data-at-startup

ferrom-jaia:
Changed mission state behavior so we can recover bot in pre deployment states


michael-jaia (approved):
Tested the data download button in multiple states and received the correct console.log statements. Example below:
(https://github.com/jaiarobotics/jaiabot/assets/109538793/33765798-fb95-4e4d-bdcd-750d6e22347a)

PR #475, opened 2023-06-10
Bug/weird blue circle when adding wpt then goes away

edsanville:
This was caused by the selectInteraction, which doesn't appear to do anything other than reset the formatting on out hub, bot, and waypoint features.  So, I removed it and so far, so good.


ferrom-jaia (approved):
Looks great! Tested in sim!

PR #476, opened 2023-06-10
Converted all camelCase endpoints to kebab-case for JCC, JED, and JDV

edsanville:


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #477, opened 2023-06-10
Task/eslint

michael-jaia:
*Integrate Prettier and typescript-eslint into web dir


michael-jaia:
This is probably something you, Ed, and I would want to discuss more closely because of the wholesale changes involved with the initial integration.

PR #478, opened 2023-06-12
task/cleanup-open-layers-map-buttons

michael-jaia:


ferrom-jaia (changes requested):
We discussed the issue with the bottom right size indicator.

ferrom-jaia (approved):
Looks good! Tested in sim!

PR #482, opened 2023-06-16
Task/refactor command control 2

edsanville:
Many changes, but here are the most important:

1) Now using componentDidUpdate, and checking for state changes to update the missionLayer and missionPlanningLayer (now separate)

2) Not re-creating sources every time, just clearing and updating features

3) Not updating the missionLayer or missionPlanningLayer when they have not changed.  In other words, the updating functions are now separate and only get called when the relevant state has changed

4) Moved survey mission feature and command generation into SurveyMission.ts

5) Combined a bunch of methods related to getting and setting the active hub/bot

6) Properly trigger a map update immediately upon placement of start/stop rally points

7) Changed names of start and stop rally points from red/green


ferrom-jaia (approved):
Looks good! Tested in sim!

PR #484, opened 2023-06-21
Catch and log serialization/parsing exceptions in xbee driver.

edsanville:
The protobuf ParseFromString C++ function actually returns false if parsing fails, which is already being dealt with and logged.  

The SerializeToString C++ function returns a bool, but there's zero documentation on what it means, so it's being ignored at the moment.  

It's possible that it's a success boolean, but it's not specified in the documentation so I don't know if it's wise to treat it as such.

The Serialization/Parsing helper functions in the xbee driver are now using try/catch blocks, and errors are glogged out (is_warn style).


ferrom-jaia (approved):
Looks great!

PR #485, opened 2023-06-21
Feature/send proto message to imu python driver

edsanville:
IMUCommand protobuf message implemented and used.  Currently, it only has one command type:  TAKE_READING


ferrom-jaia (dismissed):
Looks great! Tested in sim.

ferrom-jaia (approved):

PR #486, opened 2023-06-21
bug/moos-log-overload

ferrom-jaia:
Blacklist moos logs to reduce goby size

In release we set this manually. In 1.y we have the ability to switch between goby log output.


michaelrockJAIA (approved):
needed to ship cyber (fleet 16)

PR #488, opened 2023-06-22
Task/reorder functions in jcc

edsanville:
setMissionState and getMissionState appear to never be used, so they were removed.

zoomToAll() and trackingTarget=='all' are not used and do not function, so they were removed.

Also, use `self` again for the local close panel functions, because `this` doesn't inherit from the surrounding scope in JS.


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #489, opened 2023-06-22
bug/task-packet-error-jcc

ferrom-jaia:
Added protection when accessing taskpacket data to ensure one bad taskpacket does not mean we cannot view the rest. 

Michael saw this happen during the Marine Exercise Training Day. We do not have a way to reproduce unfortunately.


michael-jaia (commented):
Should we have a standard for checking if a property exists on an object such as "use: if (object.hasPropertyType(propertyName) {}" instead of using if (object.propertyName !== undefined) {}. I just bring this up because we could still encounter bugs if the data comes in as null.

ferrom-jaia:
Merging this one in after talking with Twomey. We have a follow on ticket to add a standardized way to check if a property is undefined or null.

PR #492, opened 2023-06-23
Change label for dives in KMZ file to an orange color (for maximum co…

edsanville:
…ntrast on dark blue seas)


edsanville:
Please let me know how this looks... the problem with KMZ files is that the appearance is highly dependent on the exact client application used to render it.  They look totally different in Google Maps versus Google Earth... and those are two apps by the same company.  And it's the company that controls the KML file format, no less!

PR #496, opened 2023-06-26
feature/add-config-for-xbee-data-rate

ferrom-jaia:
Added ability to configure xbee data rate.

The lower the data rate the more sensitive the xbee radio will be able to pick up comms.

This means we could see greater ranges.

By default the xbee radio sets the data rate high.

This pull request needs to be bench/field tested.


jason-jaia:
@ferrom-jaia Do both our radios support this? And if not - what happens if you try to set it for one that doesn't support it?

PR #498, opened 2023-06-27
Change depth labels to orange, to contrast well with dark and blue ba…

edsanville:
…ckgrounds


ferrom-jaia (approved):
looks great! Tested in sim.

PR #504, opened 2023-06-29
Bug/should-be-able-to-delete-unassigned-runs-in-jcc

michael-jaia:
Also includes: bug/should-not-be-able-to-unassign-bot-while-in-mission


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #510, opened 2023-07-05
bug/imu-warning-sensor-not-setup

ferrom-jaia:
IMU was not getting setup at start so if we request data we should make sure the setup is complete


ferrom-jaia:
Tested this on bot5 fleet 1

PR #511, opened 2023-07-05
feature/click-on-flag-for-delete-run-option

michael-jaia:


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #513, opened 2023-07-07
bug/jcc-crash-bot-no-gps

michael-jaia:
Also removed the filter method since it did not include a function that altered the array produced by the map method


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #514, opened 2023-07-07
Feature/bin-speed-and-rudder-for-rc

michael-jaia:


ferrom-jaia (approved):
Tested in sim and on a bot

PR #515, opened 2023-07-08
Allow user to import the exported KMZ files

edsanville:


ferrom-jaia (approved):
Looks good. Tested in with kmz generated during marine test.

PR #519, opened 2023-07-11
bug/bot-getting-stuck-while-doing-bottom-dive

ferrom-jaia:
**Dive Prep State**

- 10 second timeout
- Move to vertical orientation (70 degrees)
- Grab Gps for Dive when we exit state
- Zero pressure sensor 

**Powered Descent**

- Delay detecting bottom for 15 seconds or detect we are diving if we move 0.3 m down
- Keep checking if depth is changing
- If it is not for three seconds then we assume we hit bottom
- Safety timer of 60 seconds to exit powered descent

**Unpowered Ascent**

- Delay detecting rising of bot for 5 seconds
- Keep checking if depth is changing
- If it is not for three seconds then assume we are stuck

**Powered Ascent**

- Keep checking if depth is changing
- If it is not for then increase both throttle and motor time on to a max of 10 seconds on and 60% throttle
- If we are rising then switch back into unpowered ascent
- Safety: check if the pitch is less then 30 degrees  (at least 3 pitch values and for at least 1 second) 
- If it is less then assume we are at the surface and switch back into unpowered ascent


michaelrockJAIA (approved):
need to test otw

PR #522, opened 2023-07-13
Feature/click-on-map-to-move-wpt

michael-jaia:


ferrom-jaia (approved):
Looks good, Tested in sim

PR #524, opened 2023-07-14
bug/play-button-in-details-should-be-disabled-if-no-run

michael-jaia:


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #527, opened 2023-07-14
task/mission-panel-toolbar-for-static-buttons

michael-jaia:


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #528, opened 2023-07-14
task/side-panels-should-grow-as-needed-jcc

michael-jaia:


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #530, opened 2023-07-19
Feature/during drift start calc for wave height

edsanville:
State machine commands IMU driver to start/stop sampling for wave height at beginning/end of surface drift
State machine commands IMU driver to start/stop sampling for bottom type at beginning/end of dive


ferrom-jaia (approved):
Approving first pass. This will allow us to incrementally make improvements to the calculations for wave height and bottom type. Tested in sim and on the flat bots.

PR #531, opened 2023-07-19
Fit open log dialog to a screen height of 601 points (web tells me th…

edsanville:
…is is a good target for tablets)


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #532, opened 2023-07-19
Show filenames of all selected logs

edsanville:


ferrom-jaia (approved):
Awesome! Looks good. Tested in sim.

PR #533, opened 2023-07-19
feature/wpt-details-lat-lon-input

michael-jaia:


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #536, opened 2023-07-19
feature/reject-missions-if-there-are-health-errors

michael-jaia:


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #537, opened 2023-07-19
Fix/archive create dir if not exist, fix script to work with python3.8 (broken zipfile module)

edsanville:
This wasn't a card, but I had to fix the script to clean up archives on mercury


ferrom-jaia (approved):
Looks good!

PR #541, opened 2023-07-21
bug/no-satellite-icon-on-station-keep-reacquire-gps

michael-jaia:


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #547, opened 2023-07-24
feature/add-bot-pitch-to-imu-detection

ferrom-jaia:
Cleanup imu detect 
added logic for checking the desired vs current heading diff to detect turning 
added logic for checking if the bot is horizontal


michaelrockJAIA (approved):
test otw

PR #551, opened 2023-07-28
task/lock-screen-in-landscape-orientation

michael-jaia:
I think it's just a quick one liner, but it needs to be tested on the tablet


ferrom-jaia:
Tested on tablet but did not work.

PR #552, opened 2023-07-31
bug/update-simulator-with-wifi-and-pressure-changes

ferrom-jaia:
Added wifi quality and fixed pressure sim naming for simulation


ferrom-jaia:
Merging after testing in sim.

PR #554, opened 2023-08-03
bug/duplicate-missions

ferrom-jaia:
Checking the time of the command and task packet to make sure they are unique

We witness a bug when the bot receives a command and sends an ack to the hub but the hub does not get the ack.
The hub will send the same mission again.

The issue is the same with the task packets sent to the hub.


jason-jaia (commented):
I think we should check that the date is increasing so a lost packet (either intentionally or not) doesn't create spurious effects.

jason-jaia (commented):
I think we can eschew the array of times and not check for uniqueness directly. Knowing that a command is newer than the last guarantees that it is unique.

jason-jaia (approved):
Approving based on code review - I haven't tested myself.

PR #559, opened 2023-08-09
JAIAB-519 bug/Add new chunk and compression command line parameters for goby_log_tool in JDV

tsaubergine:
Requires Goby 3.0.17 for the new chunking/compression settings in `goby_log_tool` (currently in staging for `test` repo only).

Suggested order of operations to merge this PR:

- Update packages.jaia.tech to use the GobySoft staging mirror for "continuous" builds. See https://github.com/jaiarobotics/jaiabot/blob/feature/split-gobysoft-repos/src/doc/markdown/page20_build.md#createupdate-the-mirror; In short: 
```
ssh ubuntu@packages.jaia.tech
sudo /opt/jaia_packages/update_gobysoft_mirror.sh
# Choose "1.y"
# Choose "Choose Staging and Release Repositories": select test + continuous.
```
- Merge https://github.com/jaiarobotics/jaiabot/pull/557 into 1.y
- Merge this PR into 1.y

PR #560, opened 2023-08-09
Split gobyd in gobyd and goby_intervehicle_portal

tsaubergine:
By using the separate gobyd and goby_intervehicle apps we will make the interprocess comms continue to work even if the intervehicle comms fail (e.g., XBee driver). No functional change otherwise.

Requires Goby 3.0.17 before merging (Same merge process as in https://github.com/jaiarobotics/jaiabot/pull/559)


ferrom-jaia (approved):
Looks great!

PR #561, opened 2023-08-09
task/add-course-over-ground-and-heading-indicators-to-jcc

michael-jaia:


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #563, opened 2023-08-09
task/add-scale-into-jdv-map

michael-jaia:


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #564, opened 2023-08-11
task/add-measurement-tool-into-jdv

michael-jaia:


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #566, opened 2023-08-17
bug/rework-deploy-script

ferrom-jaia:
Revert back to overwriting runtime script, in the deploy script read bot index, hub index, fleet index, and electronics stack before overwriting so we can update with correct values


tsaubergine:
This is great, thank you.

PR #567, opened 2023-08-17
task/increase-number-of-wpts-allowed

ferrom-jaia:
Allow 30 points again since we fixed bug that crashed gobyd


michaelrockJAIA (approved):
fully vetted. ready to test on the bench

PR #573, opened 2023-08-23
task/rework-rc-bins

ferrom-jaia:
Changing bins to map rudder control better


ferrom-jaia:
Tested in the water and bench

PR #574, opened 2023-08-24
task/improve-function-flow-of-command-control

michael-jaia:


michael-jaia:
Command Control performs many different tasks, and some of the code for those tasks started blending together. I went through the file and separated the various tasks into "sections" using beginning and ending comments. It's difficult to see the difference on GitHub, but I believe the improvements will be noticed when the file is opened in a code editor.

ferrom-jaia (approved):
Looks great! Tested in sim.

PR #576, opened 2023-09-05
bug/not-all-health-errors-warnings-appear

ferrom-jaia:
Jaiabot fusion is randomly not responding. This should give the proper error message.

This error is seen on fleet 2 (with the usbs that do not function as well)


ferrom-jaia:
Tested in sim.

PR #577, opened 2023-09-05
feature/add-safety-to-prevent-diving-lower-than-60m

ferrom-jaia:
Added safety protection for exceeding max depth


ferrom-jaia:
Tested in sim.

PR #578, opened 2023-09-05
feature/panel-for-task-packet-visualizations

michael-jaia:


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #579, opened 2023-09-06
bug/status-age-greater-than-30

ferrom-jaia:
Added guards around send all buttons and output to let the user know  what is happening. The main reason for this is to not send to bots that have high status age (over 30) because this causes comms issues waiting for replies from bots that are no longer available

It also protects against sending missions to bots that are on but not activated. This caused frustration in the field because the operator was forced to activate all bots. This meant using the rally button would be chaotic because it would send the mission to the bots that were not being actively used.


ferrom-jaia:
Tested in sim. Will need field testing.

PR #580, opened 2023-09-06
bug/distance-to-hub-not-working

ferrom-jaia:
Grabbing first hub to get the distance. Currently we should only see one hub on jcc because they do not talk to each other.


michaelrockJAIA (approved):
looks good. overlook when switching from single to multihub

PR #581, opened 2023-09-06
Feature/imu significant wave height pass 2

edsanville:
Updated IMU glitch filtering
Expanded bandpass filter for significant wave height calculations
Some other refactoring


ferrom-jaia (approved):
Looks good. Will merge in knowing field testing is needed.

PR #582, opened 2023-09-07
feature/delete-a-run-while-in-mission

michael-jaia:


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #583, opened 2023-09-07
task/update-imu-detection-reaction

ferrom-jaia:
Updating imu detection case statement.

This includes being able to report and still go into imu restart state without restarting the imu service.


ferrom-jaia:
Merging, Needs field testing.

PR #584, opened 2023-09-07
feature/disabled-button-output

ferrom-jaia:
Added ability to click disabled buttons so the user can get an explanation why it is disabled.


ferrom-jaia:
Tested in Sim. Still need to do a field test.

PR #585, opened 2023-09-08
feature/allow-only-one-run-to-be-editable-at-a-time

michael-jaia:
Replaced old edit mode toggle code with a more concise version that checks to see which run is in edit mode.

For unassigned runs, I used the reverse numbering convention. Negative 'bot ids' represent unassigned runs in this version. They are the multiplicative inverse of the run number.


ferrom-jaia (approved):
Tested in sim. Looks good!

PR #587, opened 2023-09-08
feature/bot-details-and-right-panels-should-use-entire-height

michael-jaia:


ferrom-jaia (approved):
Looks good, Saw it in sim

PR #588, opened 2023-09-08
bug/optimize-mission-panel-apply

michael-jaia:


ferrom-jaia (approved):
Looks good, saw it working in sim

PR #589, opened 2023-09-10
Bug/jdv does not have layer for drift information

edsanville:


ferrom-jaia (dismissed):
Looks good. Tested in sim.

PR #591, opened 2023-09-11
bug/saving-mission-in-browser-does-not-work

ferrom-jaia:
Missions did not save correctly. They were not saving in local storage because of how the update method was handling them. I created a separate save and load function for missions because I believe we handle them differently.


ferrom-jaia (commented):
Tested in sim. Going to field test both the saved missions and the update made to imu detection code.

michael-jaia (approved):

PR #592, opened 2023-09-11
bug/tap-to-move-not-working

michael-jaia:


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #593, opened 2023-09-11
bug/allow-tap-to-move-for-unassigned-runs

michael-jaia:


ferrom-jaia (approved):
Looks good! Test in sim.

PR #594, opened 2023-09-12
bug/pressure-correction-start-state

ferrom-jaia:
Need to set the correction for adjusted pressure at the start of dive prep to accurately get depth


ferrom-jaia:
Tested in sim. Going to test in field.

PR #596, opened 2023-09-12
Key for heading and course over ground, when bot is clicked on map

edsanville:


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #597, opened 2023-09-12
Time slider is a child of the main vertical flexbox, so it doesn't co…

edsanville:
…ver the map at all


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #602, opened 2023-09-13
task/remove-rccontroller-panel-prop-clearInterval

ferrom-jaia:
removing clearInterval property in the RCControllerPanel as we no longer use it


ferrom-jaia:
@ealvar3z 

Removed clearInterval property as we are no longer using it. It was not the best name for the property either.

Thank you for your help!

PR #603, opened 2023-09-14
task/split-task-packet-layer-into-drift-and-dive

michael-jaia:


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #604, opened 2023-09-15
bug/saving-mission-without-cached-ones

ferrom-jaia:
Fixed saving missions when we do not start with a local storage one. Also, fixed loading in saved mission


ferrom-jaia:
Merging in, tested in sim, did peer coding with Twomey.

PR #605, opened 2023-09-18
feature/current-icons-grow-base-on-speed

michael-jaia:
- Let me know what you think of the flashing and the size of the arrow


ferrom-jaia (approved):
Looks good! Tested in sim. Icon may need a small change in the future.

PR #607, opened 2023-09-19
Constant heading arrow is gold, if canEdit

edsanville:


ferrom-jaia (dismissed):
Looks great! Tested in sim.

PR #613, opened 2023-09-25
bug/fix-flag-indexing-on-map

michael-jaia:
Now an operator can change the rally points in the middle of creating a line tool mission


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #619, opened 2023-09-28
bug/rally-confirmation

ferrom-jaia:
Added a return bool to deleteAllRuns function to determine if we should continue to execute mission


michael-jaia (approved):
Tested with two bots in sim performing simultaneous runs, then tested loading a mission and running that mission. The bots and UI performed as I expected.

PR #621, opened 2023-10-03
Added new h5error.py script, which scans a JaiaH5 file for any error …

edsanville:
…and warning statuses, and prints the times and processes on which they occurred


ferrom-jaia (approved):
Looks great! Tested with h5 file and worked as expected.

PR #622, opened 2023-10-03
task/a-run-should-not-be-in-edit-mode-after-using-line-tool

michael-jaia:


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #623, opened 2023-10-03
bug/web-portal-bug

ferrom-jaia:
Fixing bug introduced when we added limitations to the server sending data to web portal


ferrom-jaia:
Reverting these changes because we do not need to treat these as radio messages. Bug was caused when we tried a multimessage command and it was not handled correctly by these changes. Tested on Bot 4 and Hub 1 fleet 2.

PR #624, opened 2023-10-04
bug/ansible-mount-not-working

ferrom-jaia:
Changed from using the built in mount that ansible provides to a shell command to remount /boot/firmware.

The built in mount command was not remounting the /boot/firmware partition so it was failing to upgrade when new firmware was available.


ferrom-jaia:
Tested on fleet 0

PR #626, opened 2023-10-04
bug/optimize-mission-disable-line-creation-without-start-and-end

michael-jaia:


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #627, opened 2023-10-05
Feature/create a layer for a uniform grid of drift directions

edsanville:
The layer currently uses the same drift icon style as the drift packet layer.
I also consolidated some common python code into a `pyjaia` module.  This way, we can use all of our Goby messages just by importing pyjaia, and also have access to the interpolation and whatever other Jaia algorithms we create.


ferrom-jaia (approved):
Looks great! Saw a Image of it working from Ed.

PR #629, opened 2023-10-06
bug/task-packet-to-kmz

ferrom-jaia:
Added a try catch just in case for no task packets and made measurements optional for dive packets in cases where the vehicle does not perform a hold


michael-jaia (approved):
Used dataset that caused the error: pre-changes the log converter threw an error, post-changes that error was resolved

PR #630, opened 2023-10-07
Feature/jdv not needing to refresh to see new logs

edsanville:
The Log Selector refreshes the list every 2 seconds now


michael-jaia (approved):
Tested adding + removing log files without refresh

PR #631, opened 2023-10-08
Feature/jdv loading screens

edsanville:
Loading indicator when user loads a log and when user adds a new plot to the charts


edsanville:
It's a bit ugly, but functional.  We can easily change it to be more attractive.

michael-jaia (dismissed):
Tested loading log files + adding plots

PR #632, opened 2023-10-09
Intervehicle API version

tsaubergine:
https://jaia-innovation.atlassian.net/browse/JAIAB-634?atlOrigin=eyJpIjoiYzQ5YWY5MTIxYTZiNDI2N2JhZDlmMDNlOWMzODBhNzIiLCJwIjoiaiJ9

**Requires** DCCL 4.2.0 (hash computation).

**Requires** Updates to Goby3 (Release 3.1.0) to fix group values > 255 in SubscriptionReport (will keep Draft PR until this is done). https://github.com/GobySoft/goby3/pull/307

Use existing instructions to pull in new required DCCL and Goby releases: https://docs.jaia.tech/md_page20_build.html#autotoc_md61 (staging mirror has already been updated and set to `test` repo, but not yet `continuous` - this should be done just before merging this PR)

**Requires** https://github.com/jaiarobotics/jaiabot-debian/pull/43

Add integer intervehicle API version to detect mismatch between hubs/bots using incompatible versions of DCCL messages.

- This version (set in jaiabot/CMakeLists.txt: `set(PROJECT_INTERVEHICLE_API_VERSION 1)`) starting at 1 will be used as the group for all intervehicle pub/sub. 
- This version must be incremented when any of the DCCL message contents change.
- CMake will compare the hash of the output `dccl --analyze --hash_only` for each of the four DCCL messages in use and give an error if this changes (to attempt to catch cases where a developer forgets to update the PROJECT_INTERVEHICLE_API_VERSION but made changes to the DCCL messages).
- hub_manager will detect a mismatch using the SubscriptionReport and publish a BotStatus message with a new error:

If hub API version > bot
```
1 | jaiabot::bot_status;1 | jaiabot.protobuf.BotStatus | 2238-Nov-08 10:54:40.945545 | bot_id: 0 time: 8484173680832225 error: ERROR__VERSION__MISMATCH_INTERVEHICLE__UPGRADE_BOT
```
or  the inverse
```
1 | jaiabot::bot_status;1 | jaiabot.protobuf.BotStatus | 2238-Nov-08 10:54:40.945545 | bot_id: 0 time: 8484173680832225 error: ERROR__VERSION__MISMATCH_INTERVEHICLE__UPGRADE_HUB
```

**JCC appears to stop working when this message is sent until a restart is sent to Apache2 - this will need to be fixed.**


michael-jaia:
Hey Toby, I started reviewing this pull request, and I was wondering if we should just check the major, minor, and patch versions of the code running on the bots and hubs. I guess it comes down to: is backwards compatibility something worth supporting since we currently tell customers different versions will not work together. It may also keep issue tracking simpler if we know a customer system is all on the same version and reduce the amount of code needed to identify version mismatches.

tsaubergine:
> Hey Toby, I started reviewing this pull request, and I was wondering if we should just check the major, minor, and patch versions of the code running on the bots and hubs. I guess it comes down to: is backwards compatibility something worth supporting since we currently tell customers different versions will not work together. It may also keep issue tracking simpler if we know a customer system is all on the same version and reduce the amount of code needed to identify version mismatches.

Right. It's not clear to me which is better. As you said if we check software version and enforce that to be compatible across a fleet, it's easier to ensure functionality and reliability.

On the other hand it seems problematic in some scenarios to enforce that all bots be on the same version, especially when there might just be minor differences (1.9.1 versus 1.9.2, for example) totally unrelated to comms.

jason-jaia:
This is a good discussion. Personally I think the message API has a few benefits:

  1.
Mixed fleets of HYDROs and PAMs might run different software in the future, but share a message architecture.
  2.
On a selfish, but practical side, if we get a vehicle in for repair, we potentially don't have to upgrade/downgrade the vehicle or our internal Hub to match the exact version. Huge time savings if we only need to know if the pressure sensor is working for instance.
  3.
A customer (or us) might have forked the repo or made a new branch and is doing a lot of cool work - they might only be concerned that a particular asset stays compatible on the message layer so they don't have to tweak the rest of their fleet. They know they can upgrade the rest of the fleet without serious repercussions.

As the system gets more stable (every day!) the changes to the messages should decrease and the interoperability above could be a nice timesaver for us and others. On the Fleet Upgrade page, it should still highlight that some vehicles are running different, but compatible versions.

  -jason

michael-jaia:
Sounds good. One other question I have is about setting and grabbing the api version through the hub command group. I see how the api version is used in creating the hub command group numeric and then how the `api_version_from_hub_command` undoes that formula to extract the bot api version. I'm wondering why we need to use the bot_id and bot_id_total rather than just using the `jaiabot::INTERVEHICLE_API_VERSION`. Thanks - Michael

tsaubergine:
> Sounds good. One other question I have is about setting and grabbing the api version through the hub command group. I see how the api version is used in creating the hub command group numeric and then how the `api_version_from_hub_command` undoes that formula to extract the bot api version. I'm wondering why we need to use the bot_id and bot_id_total rather than just using the `jaiabot::INTERVEHICLE_API_VERSION`. Thanks - Michael

Each bot uses a different group for its subscription to HubCommand so that we don't waste radio bandwidth sending all the commands to all the bots (for example - no need for bot 3 to have commands to bot 1). This PR extends that so each group is unique for a given bot id AND api version.

I hope I understood the question right, if not please clarify.

michael-jaia:
Ok so just to clarify: previously hub commands were broadcasted (based on groups.h), but now they will be sent specifically to the  intended bot? Creating the uniqueness makes sense to me, I'm just wondering if the reason you multiply the api version by the bot_id_total is to prevent conflicts with the groups that use the api version as their group number?

tsaubergine:
> Ok so just to clarify: previously hub commands were broadcasted (based on groups.h), but now they will be sent specifically to the intended bot? Creating the uniqueness makes sense to me, I'm just wondering if the reason you multiply the api version by the bot_id_total is to prevent conflicts with the groups that use the api version as their group number?

Previously hub commands used a numeric group that was equal to the bot id:
https://github.com/jaiarobotics/jaiabot/pull/632/files#diff-ecfc0b7dd4ca5701a84324e26d9fb6a2278d458981d37c150d617193dcf9693eL521-L524

Now they use a numeric group equal to the bot id + bot id max * api version, so each api version has a unique set of groups, making it impossible to receive a DCCL message that would be decoded incorrectly due to the API mismatch:
https://github.com/jaiarobotics/jaiabot/pull/632/files#diff-8c1f59e7a9f2e75a2ac5e200336d5e80e3660b5fc016b3882a6cb2ee8a540c30R39-R44

michael-jaia:
Thank you, it's a little thing but I'm wondering if we need to multiply by the bot_id_total? Would it work if we just did bot_id + api_version or bot_id * api_version? It's nit-picky but just want to make sure I have it understood fully.

tsaubergine:
Because neither are unique.

For example
Bot id + api version: bot 2, version 1 is the same as bot 1 version 2

Bot id * api version: bot 2, version 2 is the same as bot 4 version 1

tsaubergine:
@michael-jaia, @ferrom-jaia : Updated to 1.y and resolved merge conflicts. Ready for review.

michael-jaia:
Hi Toby -- sorry for the lag in the review -- I got around to testing on the flat fleet today. One little thing I noticed is when I deploy code to the hub to set its version number greater than that of the bot, I do not see the mismatch error until I restart the bot's services. I didn't think much of it until I deployed code to the bot, setting its version number greater than that of the hub, and no restart of the hub's services was required to catch the mismatch error. Do you know if there is a reason the bot needs a restart to detect the version mismatch, but the hub doesn't? Thanks

tsaubergine:
> Hi Toby -- sorry for the lag in the review -- I got around to testing on the flat fleet today. One little thing I noticed is when I deploy code to the hub to set its version number greater than that of the bot, I do not see the mismatch error until I restart the bot's services. I didn't think much of it until I deployed code to the bot, setting its version number greater than that of the hub, and no restart of the hub's services was required to catch the mismatch error. Do you know if there is a reason the bot needs a restart to detect the version mismatch, but the hub doesn't? Thanks

This may be due to the way the subscriptions are handled to support multiple hubs (see https://docs.jaia.tech/md_page50_communications.html#autotoc_md189). I'd have to think hard about it, but I'd rather not unless this behavior is problematic.

ferrom-jaia:
@tsaubergine Do you want to look into why this is failing the jammy build?

tsaubergine:
> @tsaubergine Do you want to look into why this is failing the jammy build?

Should be all set now.

ferrom-jaia (approved):
Looks great! This was tested by Twomey in multiple different ways.

PR #635, opened 2023-10-16
GPS icon rotates with bot

edsanville:


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #636, opened 2023-10-16
Fixed "null" profiles saved when user clicks Cancel on the save profi…

edsanville:
…le prompt


ferrom-jaia (approved):

PR #637, opened 2023-10-17
Feature/add indicator for files being converted in jdv

edsanville:
Logs only convert when opened, with a busy indicator while the backend does the conversion.


ferrom-jaia (approved):

PR #638, opened 2023-10-17
task/configure-jaiabot-data-vision-package

ferrom-jaia:
Updated templates to use jaiabot/web/venv and utilizing just one requirements.txt

Requires: https://github.com/jaiarobotics/jaiabot-debian/pull/45


edsanville:
It looks good to me, except I added a line to lock the version of Werkzeug in the requirements.txt file.  The 3.0.0 version deprecated some stuff and broke Flask, which had a ">=" dependency.  Everything else looks good!

PR #639, opened 2023-10-18
Bug/hitting cancel in load profiles gives you a forever loading screen

edsanville:
Also fixed issue with rendering the bot markers, where the heading and other status information was not being set properly


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #640, opened 2023-10-18
Bug/google maps needs contribution on screen

edsanville:
Due to the Google Maps Terms of Service, I switched to using the ESRI/ArcGIS satellite imagery viewable here:
https://earthexplorer.usgs.gov/


ferrom-jaia (dismissed):
This is a great start. Tested in sim and it works. Not as useful as google maps because of zoom restrictions. We needed to remove google maps.

ferrom-jaia (approved):

PR #642, opened 2023-10-20
Fix CircleCI debian/rootfs build

edsanville:
Consolidated python virtual environments into one (src/python/requirements.txt)
Must be tested in conjunction with the corresponding changes in the jaiabot-debian and jaiabot-rootfs-gen repositories.


ferrom-jaia (approved):
Looks great!

PR #644, opened 2023-10-23
feature/remove-alerts-and-create-custom-popup

edsanville:
Replaced all alert() and confirm() calls with CustomAlert.alert() and CustomAlert.confirm() / CustomAlert.confirmAsync()

We can customize the look of the confirm / alert dialog boxes using the `CustomAlert.css` file.

We're gonna need to test all the parts of the code that used "confirm()" to confirm an action.


michael-jaia (approved):
Tested in sim

PR #651, opened 2023-10-24
Add two comms radii around the hub

edsanville:


ferrom-jaia (approved):
Looks great! Tested in sim!

PR #655, opened 2023-10-30
Allow deletion of goals in unassigned runs

edsanville:


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #657, opened 2023-10-30
bug/copying-a-run-should-copy-repeats

edsanville:
Suggestion:
Refactor MissionInterface into a Mission class, and move some of those static functions into that class as class methods?
Potentially also move RunInterface into a Run class.

Both classes could live in their own source files.


ferrom-jaia (dismissed):
Looks great! Tested in sim.

michael-jaia (approved):
Tested in sim -- fixed delete run bug that arose from resolving merge conflicts

PR #659, opened 2023-10-31
feature/indicator-for-run-repeat-it-is-on

edsanville:


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #662, opened 2023-11-02
task/remove-mui-animations

michael-jaia:
Also includes drift icon svg patch


ferrom-jaia (dismissed):
Looks great! Tested in sim.

ferrom-jaia (approved):
Looks great! Tested in sim.

PR #663, opened 2023-11-03
bug/openlayers-style-uses-different-colors-than-css

edsanville:
Map styles are now in shared/CSSStyles.less, with TypeScript code in shared/CSSStyles.ts for easy auto-completion when the styles need to be used from TS code.


ferrom-jaia:
Closing for now. We need to discuss if this is needed.

PR #665, opened 2023-11-03
bug/cannot-select-hub-icon-to-open-details

michael-jaia:


ferrom-jaia (approved):
Good work! Tested in sim.

PR #666, opened 2023-11-05
task/jdv-file-size-bytes-to-megabytes

edsanville:


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #667, opened 2023-11-09
bug/omp-clicking-on-rally-point-closes-panel-but-tool-still-active

michael-jaia:
# bug/omp-clicking-on-rally-point-closes-panel-but-tool-still-active

### Issue:
* When `this.state.mode` is set to `Mode.NEW_RALLY_POINT` and the operator goes to open Mission Planning Panel, the panel does not open
* Note the original bug in the ticket name no longer existed when I started...I found the bug stated in the previous bullet while trying to reproduce the bug in the title name

### Solution:
* We needed an event.stopPropagation call to prevent the event on the Mission Planning button from propogating to `handleJccContainer()` click where `this.state.mode` gets set to `Mode.NONE`

### Tests
omp = 'optimized mission planning tool
* Add 2 rallys, click to add a third rally but hit the omp before adding it
    * Does the omp work as expected?
        * PASS (for tasks: none, dive, drift & with end task)

* Add 1 rally, click to add a second rally, hit the omp
    * OMP should not open
        * PASS

* Add 3 rallys, click to add a fourth, hit the omp before adding it
    * Does the omp work as expected?
        * PASS

* Open the omp, plan a mission, click the add rally button
    * Does it work as expected?
        * PASS

* Open the omp, half plan a mission, click the add rally button
    * Does it work as expected?
        * PASS

* Open the omp, don't plan a mission, click the add rally
    * Does it work as expected?
        * PASS

* Retry all of these tests with the details panel open
    * [PASS, PASS, PASS, PASS, PASS, PASS]


src/web/command_control/client/components/CommandControl.tsx
```diff
@@ -3171,11 +3170,12 @@ export default class CommandControl extends React.Component {
 		) : (
 			<Button
 				className="button-jcc"
-				onClick={() => {
+				onClick={(evt) => {
 					if (!this.canUseSurveyTool()) {
 						return
 					}
 
+					evt.stopPropagation() // Allows seemless switch between Modes (without it, handleJccContainerClick fires and causes bugs)
```
ferrom-jaia: I do not think we should use inline comments. Can we place this above?

ferrom-jaia (changes requested):
Found one small issue

ferrom-jaia:
Tested in sim. Looks/works great!

ferrom-jaia (approved):
Looks great!  Tested in sim.

PR #669, opened 2023-11-09
bug/task-packet-date-selection-does-not-reset-when-toggled

michael-jaia:
# bug/task-packet-date-selection-does-not-reset-when-toggled

### Description:
When you toggle off Edit Dates, the "Keep end date current" check box is not reset to the default of being turned on which prevents you from getting new task packets

### Approach:
* "Keep end date current" should be in state and it should be changed if the user toggles the Edit Dates view off

#### Important Variables & Functions:
`props.taskPacketsTimeline.keepEndDateCurrent`
`props.handleTaskPacketEditDatesToggle()`

### Tests
ED = Edit Dates
accord = accordion
* Open ED accord, turn off checkbox, close ED accord, reopen it...checkbox should be on
    * PASS
* Open ED accord, turn off checkbox, change end date, close ED accord, open it...date should reset to now and box should be checked
    * PASS
* Open ED accord, turn off checkbox, change start date, close ED accord, open ED accord...should keep the start date the user set
    * PASS
* Open ED accord, turn off checkbox, change start and end date, close ED accord, open ED accord...should keep the start date the user set and end date current
    * PASS

### DIFF
--- a/src/web/command_control/client/components/CommandControl.tsx
+++ b/src/web/command_control/client/components/CommandControl.tsx
@@ -2129,6 +2129,7 @@ export default class CommandControl extends React.Component {
 			}).catch((err) => {
 				console.error('Task Packets Retrieval Error:', err)
 			})
+			// Set "Keep End Date Current" checkbox to checked
 			taskPacketsTimeline.keepEndDateCurrent = true
 		}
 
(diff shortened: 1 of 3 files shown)


ferrom-jaia (approved):
Code review looks great! Saw it working in sim.

PR #670, opened 2023-11-11
 bug/jdv-filters-not-resetting-correctly-after-reload

edsanville:
Fix bug where GUI wouldn't properly show the log filters being applied, especially on reload


ferrom-jaia (approved):
Looks great! Tested on my computer.

PR #673, opened 2023-11-12
Use fmod to enforce constant heading to the [0, 360] range

edsanville:


edsanville:
Right now, these input fields are React "controlled" elements, which means that their values follow that of the props that are given to the component.  

If we want to make the field "wait" for the user to press enter or for them to lose focus before they validate and submit the changes to the value, we would need to make the heading field an uncontrolled component.

PR #675, opened 2023-11-12
Bug/task constant heading input

edsanville:
FYI:  Because the heading field is a React "controlled" field (whose value always comes from the props), and we validate the value in the onChange handler, there isn't a possibility of going outside the [0,359] range and getting the read outline for the field.


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #680, opened 2023-11-13
feature/when-saving-mission-it-should-keep-name

edsanville:
Name is set when user saves a mission using the Save Mission dialog


ferrom-jaia (changes requested):
The mission name does not set until you have saved it in local storage and select the mission name to save to disk.

ferrom-jaia (changes requested):
Tested in sim and looks great! 

Can we add function comments to the ones that were modified: https://github.com/jaiarobotics/jaiabot/blob/1.y/src/doc/markdown/page110_coding_standards.md

src/web/command_control/client/components/SaveMissionPanel.tsx
```diff
@@ -99,6 +99,13 @@ export class SaveMissionPanel extends React.Component {
         this.setState({selectedMissionName: name})
     }
 
+    
+    /**
+     * Called when user clicks the Save button (to save a mission to localStorage in their browser)
+     *
+     * @async
+     * @returns {*}
```
ferrom-jaia: Can we use {void} instead of {*}

src/web/command_control/client/components/SaveMissionPanel.tsx
```diff
@@ -141,8 +149,12 @@ export class SaveMissionPanel extends React.Component {
         this.props.onDone?.()
     }
 
+    
+    /**
+     * Called when user clicks the Download button (to download a json mission)
```
ferrom-jaia, line 153: can we add @returns {void}

src/web/command_control/client/components/TaskSettingsPanel.tsx
```diff
@@ -12,6 +12,20 @@ import { Point } from 'ol/geom';
...
 // For keeping heading angles in the [0, 360] range
+
+/**
+ * Returns the modulus between two real numbers, without returning any negative numbers.
+ * This is useful, for example, for keeping angles between 0 and 360, even after adding and subtracting 
+ * angle values from one another.
+ *
+ * @param {number} a The dividend of the modulus operation
+ * @param {number} b The divisor of the modulus operation
+ * @returns {*} `a` FMOD `b`
```
ferrom-jaia: This would return a number right? {number}

ferrom-jaia (approved):
Looks great! Tested in sim.

PR #681, opened 2023-11-13
task/jdv-message-for-downloading-csv

edsanville:
Added message telling user to select some series to plot or export to CSV

image


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #682, opened 2023-11-13
feature/add-upgrade-link-in-hub-links

edsanville:
image


ferrom-jaia (approved):
Looks great! Tested in sim and on the flat hub.

PR #683, opened 2023-11-13
bug/deleting-run-has-lag

michael-jaia:
# bug/deleting-run-has-lag

## Problem
* When the operator deletes a run from the Details Panel or using the delete all runs button in the Mission Panel there can be lag in the run(s) disappearing from the map
#### Cause of Problem
* The run(s) is deleted from the runList, but setState is not called after

## Solution
* Call `setRunList` passing the updated runList as the argument. This function calls setState and also updates the runList version which is needed for the undo button

## Tests
3x = repeat 3 times
* Delete a run with no throttle from the Details Panel...does it happen instantly? 3x
    * PASS
* Delete a run with no throttle from the Mission Panel (unassigned)...does it happen instantly? 3x
    * PASS
* Delete a run with no throttle from the Mission Panel (assigned)...does it happen instantly? 3x
    * PASS
* Delete 3 runs with no throttle from the Mission Panel using the delete all button...does it happen instantly?
    * FAIL
    * The delete all runs button was also not updating state immediately after. I added the code to make it do so and it passed.
    * PASS
* Delete a run from the Details Panel with 4x throttle...does it happen instantly? 3x
    * PASS
* Delete 3 runs individually from the Mission Panel with 4x throttle...does it happen instantly? 3x
    * PASS
* Delete 3 runs from the Mission Panel with 4x throttle using the delete all button...does it happen instantly? 3x
    * PASS
* Repeat the tests 1 and 3 on a different bot
    * PASS
* Undo a deletion from Details Panel, Mission Panel (individual run deletion and all runs deleletion)
    * PASS

#### Errors Encountered
* `Uncaught TypeError: Cannot read properties of null (reading 'getBoundingClientRect')`
    * Resolved with:
        ```
        if (!parentContainer || !dropdownContainer) {
            return
        }
        ```
* `Uncaught TypeError: Cannot read properties of undefined (reading 'id')
    at CommandControl.addWaypointAt (CommandControl.tsx:1669:11)`
    * Resolved with:
        * `if (!run) { return }`

### DIFF
--- a/src/web/command_control/client/components/CommandControl.tsx
+++ b/src/web/command_control/client/components/CommandControl.tsx
@@ -1244,16 +1244,16 @@ export default class CommandControl extends React.Component {
 	// 
 	async deleteAllRunsInMission(mission: MissionInterface, needConfirmation: boolean, rallyPointRun?: boolean) {
 		return new Promise((resolve, reject) => {
+			let updatedMission = {...mission}
 			const doDelete = () => {
-				const runs = mission.runs
-				for (const run of Object.values(runs)) {
-					const runNumber = Number(run.id.substring(4)) // run.id => run-x
-						delete mission.runs[run.id]
-						delete mission.botsAssignedToRuns[run.assigned]
+				for (const run of Object.values(mission.runs)) {
+					delete updatedMission.runs[run.id]
+					delete updatedMission.botsAssignedToRuns[run.assigned]
 				}
-				mission.runIdIncrement = 1
-				mission.runIdInEditMode = ''
-	
+				updatedMission.runIdIncrement = 1
+				updatedMission.runIdInEditMode = ''
+				this.setRunList(updatedMission)
+
 				resolve(true)			
 			}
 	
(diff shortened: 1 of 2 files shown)


src/web/command_control/client/components/CommandControl.tsx
```diff
@@ -1231,16 +1231,16 @@ export default class CommandControl extends React.Component {
 	// 
 	async deleteAllRunsInMission(mission: MissionInterface, needConfirmation: boolean, rallyPointRun?: boolean) {
```
ferrom-jaia: Since we are updating this function can we update the comment to use the correct /**/

src/web/command_control/client/components/CommandControl.tsx
```diff
@@ -1651,6 +1652,9 @@ export default class CommandControl extends React.Component {
 		} else {
 			run = runs[botsAssignedToRuns[botId]]
 		}
+
```
ferrom-jaia, line 1671: Since we are updating this function can we update the comment to use the correct /**/

src/web/shared/Utilities.tsx
```diff
@@ -150,6 +150,11 @@ function handleAccordionDropdownClick(event: Event, targetClassName: string, par
 
 export function adjustAccordionScrollPosition(parentContainerId: string, dropdownContainer: HTMLElement) {
     const parentContainer = document.getElementById(parentContainerId)
+
```
ferrom-jaia, line 164: Since we are updating this function can we update the comment to use the correct /**/

ferrom-jaia (changes requested):

ferrom-jaia (approved):
Looks great! Thanks for the documentation in the function headers! Tested in sim.

PR #686, opened 2023-11-14
bug/hub-not-always-showing-on-map

ferrom-jaia:
Mimicked how we update bot icons to update hub icons

https://github.com/jaiarobotics/jaiabot/assets/109534520/4eb4a5ba-0180-481f-bf37-d50913933770

### DIFF
--- a/src/web/command_control/client/components/HubLayers.ts
+++ b/src/web/command_control/client/components/HubLayers.ts
@@ -96,7 +96,7 @@ export class HubLayers {
         } // end foreach hub
 
         // Remove hub layers for hub_ids that have disappeared
-        const defunctHubIds = Object.keys(this.layers).filter((hub_id) =>  !(String(hub_id) in hubs))
+        const defunctHubIds = Object.keys(this.layers).filter((hub_id) => !(String(hub_id) in hubs))
 
         defunctHubIds.forEach((hubIdString) => {
             this.deleteHubLayer(Number(hubIdString))
--- a/src/web/command_control/client/components/HubLayers.ts
+++ b/src/web/command_control/client/components/HubLayers.ts
@@ -53,7 +53,7 @@ export class HubLayers {
 	}
 
     /**
-     * Used to update the hub icons on the openlayers map
+     * Used to update the hub icons on the OpenLayers map
      * 
      * @param {{[key: string]: PortalHubStatus}} hubs - map of hubs to loop through and check for updates
      * @param {HubOrBot} selectedHubOrBot - an object to determine if a hub id is selected so we can update state accordingly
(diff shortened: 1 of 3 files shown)


src/web/command_control/client/components/HubLayers.ts
```diff
@@ -52,34 +52,14 @@ export class HubLayers {
...
-        })
-
-        source.addFeature(newFeature)
-    
-        return newFeature
-    }
-
-	update(hubs: {[key: string]: PortalHubStatus}, selectedHubOrBot: HubOrBot) {
+    /**
+     * Used to update the hub icons on the openlayers map
```
michael-jaia: Let's make this OpenLayers to match the spelling on their website

src/web/command_control/client/components/HubLayers.ts
```diff
@@ -88,11 +68,38 @@ export class HubLayers {
 
             const hubLayer = this.getHubLayer(hub_id)
             const hubSource = hubLayer.getSource()
-            const hubFeature = this.getHubFeature(hub, this.map, hubSource)
-    
+            let hubFeature: Feature<Point> = hubSource.getFeatureById(hub_id) as Feature<Point>
```
michael-jaia: I think we can remove one of the Feature<Point> phrases

src/web/command_control/client/components/HubLayers.ts
```diff
@@ -88,11 +68,38 @@ export class HubLayers {
...
+            if (hubFeature == null) {
```
michael-jaia: Let's make this === since we know the getFeatureById returns null if no feature is found

src/web/command_control/client/components/HubLayers.ts
```diff
@@ -88,11 +68,38 @@ export class HubLayers {
...
+                hubFeature = new Feature<Point>({
+                    name: String(hub_id),
+                })
+                hubFeature.setId(hub_id)
+                hubSource.addFeature(hubFeature)
+            }
+
+            if (hub?.location !== undefined) {
```
michael-jaia, line 80: Just want to make sure that if a value does not come through the protobuf message it comes across as undefined (since we have the strict conditional check here)

src/web/command_control/client/components/HubLayers.ts
```diff
@@ -88,11 +68,38 @@ export class HubLayers {
...
             hubFeature.set('selected', selected);
+
+            hubLayer.changed();
+
         } // end foreach hub
-    }
-}
 
+        // Remove hub layers for hub_ids that have disappeared
+        const defunctHubIds = Object.keys(this.layers).filter((hub_id) => {return !(String(hub_id) in hubs)})
```
michael-jaia: I think we can make this cleaner with by removing the brackets and the return keyword
`const defunctHubIds = Object.keys(this.layers).filter((hub_id) =>  !(String(hub_id) in hubs))`

ferrom-jaia:
Twomey Approved.

PR #690, opened 2023-11-15
feature/update-slip-radius-for-wpt-with-no-task

ferrom-jaia:
Added slip radius updates based on if a waypoint has a task or not


michael-jaia (approved):
Statically reviewed code with Matt + Matt saw slip radius updating properly in log file

PR #696, opened 2023-11-16
task/remove-log-converter

ferrom-jaia:
Removing log converter systemd setup


michael-jaia (approved):
Deployed to fleet 2...approving pull request

PR #697, opened 2023-11-17
task/remove-undo-feature-from-help-window

ferrom-jaia:
Removed undo feature and added fix for node module build failures


src/web/install_dependencies.sh
```diff
@@ -9,5 +9,29 @@ fi
 
 echo 🟢 Installing npm dependencies to: $NODE_MODULES_DIR
 
+if [ -d "${NODE_MODULES_DIR}node_modules" ]; then
```
michael-jaia, line 15: Not sure if comments are a thing in scripting...if they are, I think it could be helpful to say why we are removing the node modules or specifically the ol-layerswitcher module in particular

src/web/jdv/client/build.sh
```diff
@@ -8,10 +8,6 @@ pushd "$DIR"
 output_path='dist'
 [[ ! -z "$1" ]] && output_path="$1"
 
-pushd ../../
```
michael-jaia, line 11: Could this cause an error for someone only running/launching the JDV?

michael-jaia (changes requested):

src/web/jdv/client/build.sh
ferrom-jaia: No, this was a duplicate installation of the dependencies.

The top level run.sh in the web/jdv folder calls the same install_dependencies.sh that web/runs.sh calls.

The build doesn't need to install the messages twice because the python deb package is setup to create a virtual python environment.

src/web/install_dependencies.sh
ferrom-jaia: Yes, I will update and they are a thing.

michael-jaia (approved):
Comments resolved, tested in sim (no errors launching JCC, JDV and undo btn removed from help window)

PR #698, opened 2023-11-17
bug/false-bottom-detection-dive

ferrom-jaia:
Fixed issue with detecting bot diving by initially setting last depth to the current one


src/bin/mission_manager/machine.h
```diff
@@ -1627,6 +1627,8 @@ struct PoweredDescent
     goby::time::SteadyClock::time_point powered_descent_timeout_;
     // determines when the bot is diving
     bool is_bot_diving_{false};
+    // initial depth reading
```
michael-jaia: Not sure if we need this comment given the name of the variable

michael-jaia (changes requested):

src/bin/mission_manager/machine.h
ferrom-jaia: Changed comment to give more insight on what the variable is used for

PR #700, opened 2023-11-20
bug/run-flag-delete-does-not-work

michael-jaia:


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #701, opened 2023-11-20
task/remove-hover-over-for-custom-buttons

michael-jaia:
On the tablet, the hover color change would not revert to its original color after a hover. This pull request removes visual effects from the buttons located in the notification popups.


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #703, opened 2023-11-20
bug/remove-lag-from-waypoint-deletion

michael-jaia:


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #704, opened 2023-11-20
bug/remove-lag-when-moving-wpts

michael-jaia:


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #705, opened 2023-11-20
bug/deleting-all-runs-confirmations

ferrom-jaia:
Only showing delete all runs popup if there are runs available


src/web/command_control/client/components/CommandControl.tsx
```diff
@@ -1066,6 +1066,15 @@ export default class CommandControl extends React.Component {
 		return Object.keys(this.getRunList().botsAssignedToRuns).length > 0
 	}
 
+	/**
+	 * Check if there are runs in the mission panel
+	 * 
+	 * @returns True if there are runs and false if there are no runs
```
michael-jaia: @returns {boolean} ...

michael-jaia (approved):
Statically reviewed code + tested in sim...approving pr

PR #708, opened 2023-11-21
bug/select-logs-panel-needs-to-be-bigger-for-tablet

edsanville:
Fix width of the LogSelector to have 10% margins,to look good on a tablet (1024x768)


ferrom-jaia (approved):
Looks great! Tested on my computer.

PR #711, opened 2023-11-21
task/add-bots-not-avail-for-download-to-confirm-msg

michael-jaia:


ferrom-jaia (approved):
Looks good. Saw demo on Twomey's computer

PR #714, opened 2023-11-25
bug/weird-files-showing-in-jdv-select-panel

edsanville:
Restrict .goby and .h5 log patterns to eliminate detection and display of other types of files as log files in the LogSelector


src/web/jdv/server/jaialog_store.py
```diff
@@ -107,6 +88,37 @@ def get_title_from_path(path):
...
+@dataclass
+@dataclass_json
+class LogDescription:
+    '''Metadata pertaining to a log'''
+
+    bot: str
+    fleet: str
+    
+    filename: str
+    '''File stem for this log (without path, .goby or .h5 extension)'''
```
michael-jaia, line 100: Do we need docstring comments for class properties or should we just use # comments above the property name?

src/web/jdv/server/jaialog_store.py
```diff
@@ -107,6 +88,37 @@ def get_title_from_path(path):
...
+    '''Log duration (in microseconds).  Only present for a log that has been converted to HDF5 format.'''
+
+    size: int = 0
+    '''Log file size (in bytes)'''
+
+
+@dataclass
+@dataclass_json
+class LogDirectory:
+    """A list of available logs with their metadata, and the available space on the storage device"""
```
michael-jaia: Should we use single or double quotes for the docstrings? I think either is fine, we just need to be more consistent

src/web/jdv/server/jaialog_store.py
```diff
@@ -107,6 +88,37 @@ def get_title_from_path(path):
...
+
+    availableSpace: int
+    '''Available storage space (in bytes)'''
```
michael-jaia, line 118: Same thing as above with the comments for the class properties...is it better to use docstring or #?

src/web/jdv/server/moos_messages.py
```diff
@@ -7,59 +7,13 @@
...
-    else:
-        return float(f)
+from h5_tools import *
 
 
 MOOS_MESSAGE_PATH = '/jaiabot::moos/jaiabot.protobuf.MOOSMessage'
 
 
-def get_moos_messages(log_filenames, t_start, t_end) -> str:
+def get_moos_messages(log_filenames: List[str], t_start: int=None, t_end: int=None) -> str:
```
michael-jaia, line 16: Add args + returns to docstring

src/web/jdv/server/moos_messages.py
```diff
@@ -70,11 +24,14 @@ def get_moos_messages(log_filenames, t_start, t_end) -> str:
     csv_writer = csv.DictWriter(output_string_io, fieldnames=fieldnames)
     csv_writer.writeheader()
 
+    t_start = t_start or 0
```
michael-jaia, line 36: What do you think of setting the default values in the function declaration. I think we could eliminate these two lines if we replaced int=None with the default values you have here

src/web/jdv/server/moos_messages.py
```diff
@@ -120,3 +77,10 @@ def get_moos_messages(log_filenames, t_start, t_end) -> str:
             })
  
     return output_string_io.getvalue()
+
```
michael-jaia: Do we still need the test?

michael-jaia (changes requested):

src/web/jdv/server/jaialog_store.py
edsanville: I prefer docstring, because the IDE recognizes and shows them when you hover over the property of any object of that class.  Will do whatever we decide, though!

src/web/jdv/server/jaialog_store.py
edsanville: This is a good point.  The plugin I use only works with double quotes for some bizzare reason, but I prefer the look of single quotes better myself.

src/web/jdv/server/moos_messages.py
edsanville: You make a good point, but the only difference is if the caller wanted to pass None (no start date) for t_start, and also pass something in for t_end.  I suppose they could pass in 0 for t_start instead of None, but I felt the "None" makes it more explicit that they don't want any start time, just an end time.  I may be overthinking this, which is typical!

src/web/jdv/server/moos_messages.py
edsanville: You're right, it's probably better to delete the test, since it references a local log file for me.

src/web/jdv/server/moos_messages.py
michael-jaia: I didn't consider that. Yeah it gets tricky because there is sort of an implicit conversion from a start date to an index. I would just go for what you think is most clear and include a note about that in the doc string

src/web/jdv/server/jaialog_store.py
michael-jaia: It always comes down to trade-offs haha!

src/web/jdv/server/jaialog_store.py
michael-jaia: Oh cool! I think it makes a lot of sense to use the docstrings for properties then

edsanville:
All comments should be addressed now.

ferrom-jaia (dismissed):
Looks great! Tested in sim.

michael-jaia (approved):

PR #715, opened 2023-11-27
feature/undo-button

michael-jaia:


src/web/command_control/client/components/CommandControl.tsx
```diff
@@ -463,6 +465,16 @@ export default class CommandControl extends React.Component {
...
 
+		const emptyMission = {
+			id: '',
+			name: '',
+			runs: {},
+			runIdIncrement: 1,
+			botsAssignedToRuns: {},
+			runIdInEditMode: ''
+		}
+		this.updateMissionHistory(emptyMission)
```
ferrom-jaia, line 476: Why do we need an empty mission?

src/web/command_control/client/components/CommandControl.tsx
```diff
@@ -1052,11 +1064,33 @@ export default class CommandControl extends React.Component {
...
+
+	/**
+	 * Saves a deep copy of the current mission state so the operator can restore changes
+	 * 
+	 * @param {MissionInterface} mission Provides current mission data to append to mission history
+	 * @returns {void}
+	 * 
+	 * @notes
+	 * The maximum size of the missionHistory array is set to the variable missionHistoryMaxLen.
+	 * This is to prevent the mission history from causing performance issues.
```
ferrom-jaia: Maybe we can be more explicit why it would cause an issue:
This is to prevent the mission history from causing performance issues -> 
This is to prevent the mission history from growing unbounded and potentially causing performance issues

src/web/command_control/client/components/CommandControl.tsx
```diff
@@ -1225,10 +1259,10 @@ export default class CommandControl extends React.Component {
...
 	 * Reset mission planning
 	 * 
-	 * @param {MissionInterface} mission used to access the mission state
-	 * @param {boolean} needConfirmation does the deletion require a confirmation by the opertor?
-	 * @param {boolean} rallyPointMission is the mission a rally point mission?
-	 * @return {Promise<boolean>} did the deletion occur? If (yes) then (true)
+	 * @param {MissionInterface} mission Used to access the mission state
+	 * @param {boolean} needConfirmation Does the deletion require a confirmation by the opertor?
+	 * @param {boolean} rallyPointMission Is the mission a rally point mission?
+	 * @returns {Promise<boolean>} Did the deletion occur? If (yes) then (true)
```
ferrom-jaia: Does it return false if the deletion did not occur.
Can we remove the parentheses as well?

ferrom-jaia (changes requested):
Looks good. I would just disable/add a warning that there is not more history to undo if the array is empty. Probably mention that the max history saved is 5.

src/web/command_control/client/components/CommandControl.tsx
michael-jaia: Screenshot from 2023-11-27 11-42-33

In the case where I create a single waypoint run, then click the undo button, there needs to be an empty mission to pass to setRunList. We could possibly pass null? I don't love the placement of the creation of the empty mission. I'd rather import it from another file. What do you think?

src/web/command_control/client/components/CommandControl.tsx
michael-jaia: Yes, are you asking because there is only an example in the comments for the "yes" case?

ferrom-jaia: Yes, I think we should be explicit so others know what to expect.

src/web/command_control/client/components/CommandControl.tsx
ferrom-jaia: I think it is fine where it is now. We know that we have some optimization that we need to do for it anyways.

michael-jaia:
Used the info() message for the warning when there is no more history to undo. I chose this (along with not disabling the button) since it's not a mission critical feature to operating and clicking it when there is no more history to undo does not cause any harm. In short, I thought not disabling it would cause less confusion, but I can disable and increase the severity of the warning if you think that would be better for operators. I also added the max number of undos in the help window.

ferrom-jaia (approved):
Looks great! Tested in sim.

PR #719, opened 2023-11-28
bug/rc-dive-parameters-crashing-jcc

ferrom-jaia:
Fix bug with rc params local storage missing and crashing jcc

### DIFF
--- a/src/web/command_control/client/components/CommandControl.tsx
+++ b/src/web/command_control/client/components/CommandControl.tsx
@@ -2450,19 +2450,19 @@ export default class CommandControl extends React.Component {
 	}
 
 	/**
-	 * Used to initialize the rc dive parameters in local storage
+	 * Used to initialize the RC dive parameters in state
 	 * 
-	 * @param botId Used to find the bot's dive parameters
+	 * @param {number} botId Used to find the bot's dive parameters
 	 * @returns {void}
 	 */
-	initRCDivesStorage(botId: number) {
+	initRCDivesParams(botId: number) {
 		if (botId !== undefined) {
 			let newRCDives = cloneDeep(this.state.rcDives)
 				newRCDives[botId] = {
-				maxDepth: '0',
-				depthInterval: '0',
+				maxDepth: '10',
+				depthInterval: '10',
 				holdTime: '0',
-				driftTime: '0'
+				driftTime: '10'
 			}
 			this.setState({ rcDives: newRCDives })
 		}
@@ -3220,7 +3220,7 @@ export default class CommandControl extends React.Component {
 					weAreInControl={this.weAreInControl.bind(this)}
 					weHaveInterval={this.weHaveRemoteControlInterval.bind(this)}
 					setRCDiveParameters={this.setRCDiveParams.bind(this)}
-					initRCDivesStorage={this.initRCDivesStorage.bind(this)}
+					initRCDivesParams={this.initRCDivesParams.bind(this)}
 				/>
 			)
 		}
--- a/src/web/command_control/client/components/CommandControl.tsx
+++ b/src/web/command_control/client/components/CommandControl.tsx
@@ -136,7 +136,6 @@ interface State {
 	podStatusVersion: number
 	botExtents: {[key: number]: number[]},
 	lastBotCount: number,
-	areBotsLoadedJCC: boolean,
 
 	missionParams: MissionParams,
 	missionPlanningGrid?: {[key: string]: number[][]},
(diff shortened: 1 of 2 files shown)


src/web/command_control/client/components/CommandControl.tsx
```diff
@@ -2456,17 +2449,23 @@ export default class CommandControl extends React.Component {
...
-		for (let botId of botIds) {
-			newRCDives[Number(botId)] = {
-				maxDepth: '',
-				depthInterval: '',
-				holdTime: '',
-				driftTime: ''
+	/**
+	 * Used to initialize the rc dive parameters in local storage
+	 * 
+	 * @param botId Used to find the bot's dive parameters
```
michael-jaia: Need to add type here

src/web/command_control/client/components/CommandControl.tsx
```diff
@@ -2456,17 +2449,23 @@ export default class CommandControl extends React.Component {
...
-	initRCDivesStorage(botIds: string[]) {
-		let newRCDives = cloneDeep(this.state.rcDives)
-		for (let botId of botIds) {
-			newRCDives[Number(botId)] = {
-				maxDepth: '',
-				depthInterval: '',
-				holdTime: '',
-				driftTime: ''
+	/**
+	 * Used to initialize the rc dive parameters in local storage
```
michael-jaia: It's inconsistent all over the codebase...not sure if it matters but should we stick to either "rc" or "RC"?

src/web/command_control/client/components/RCControllerPanel.tsx
```diff
@@ -60,6 +61,12 @@ export default class RCControllerPanel extends React.Component {
         super(props)
         this.api = props.api
 
+        // Check to see if rc dive parameters are
+        // saved in local storage
```
michael-jaia: I think they are saved in state and not local storage, but I could be wrong here

src/web/command_control/client/components/RCControllerPanel.tsx
```diff
@@ -515,42 +522,46 @@ export default class RCControllerPanel extends React.Component {
...
-						<div>m</div>
-
-						<div>Hold Time:</div>
-						<input id='holdTime' className='rc-input' type='text' value={this.props?.rcDiveParameters?.holdTime} onChange={(evt: React.ChangeEvent<HTMLInputElement>) => this.handleTaskParamInputChange(evt)} autoComplete='off'/>
-						<div>s</div>
-
-						<div>Drift Time:</div>
-						<input id='driftTime' className='rc-input' type='text' value={this.props?.rcDiveParameters?.driftTime} onChange={(evt: React.ChangeEvent<HTMLInputElement>) => this.handleTaskParamInputChange(evt)} autoComplete='off' />
-						<div>s</div>
+		if (this.props?.rcDiveParameters !== undefined) {
```
michael-jaia: I don't think we need the question mark after props since rcDiveParameters is a required prop

src/web/command_control/client/components/RCControllerPanel.tsx
```diff
@@ -515,42 +522,46 @@ export default class RCControllerPanel extends React.Component {
...
-                    </Button>
-				</div>
-			</div>
-		)
+			)
+		}
 
-		// Set the remoteControlValues to the selected bot id
-		this.props.remoteControlValues.bot_id = this.props.bot.bot_id
+		if (this.props?.bot?.bot_id !== undefined) {
```
michael-jaia: Do we need all of these question marks?

michael-jaia (changes requested):

src/web/command_control/client/components/CommandControl.tsx
ferrom-jaia: Nice find!

src/web/command_control/client/components/CommandControl.tsx
ferrom-jaia: Let's start to use RC

src/web/command_control/client/components/RCControllerPanel.tsx
ferrom-jaia: Yup, they are! Changed the name of the function.

src/web/command_control/client/components/RCControllerPanel.tsx
ferrom-jaia: No, we still need the one before bot_id since it is an optional field currently

michael-jaia (approved):
Comments resolved...Matt clearly reproduced the bug and can no longer do so with these new code changes...approving pull request

PR #724, opened 2023-12-01
task/rc-command-send-speed-too-fast

ferrom-jaia:
The xbee radio cannot handle sending the rc command more than twice a second


ferrom-jaia:
Tested on the water. I was able to drive bots 3 and 4 back into shore from 100m out.

PR #725, opened 2023-12-03
Feature/jed tail bounds calibration

edsanville:
This is a first pass at the calibration GUI.  It's not yet merged into the main JED app, but you can access it like this:

[host]:40001/jed/calibration.html

Next, I will integrate it into the bottom of the main JED screen, while addressing any issues we find.


edsanville:
Side note:  It might be worth migrating JED to React at some point.

michael-jaia (changes requested):
Hey Ed, I just looked over the PR, and after doing the tail calibrations last week, I don't think the sliders are the best user experience. Input boxes for each bound are quicker because the bounds usually change in increments of 5 or 10 so it can be tricky to get the exact number with a slider. 

The second item is changing the word "Query" to "Get up-to-date bound values" or something along those lines because query can be a little vague as to the actual function it performs. Also I didn't see the bounds values in the Engineering Status message when I tried to query the updated values. 

We could also change the "Submit" button to "Submit bot bounds" (or something similar) to remove any ambiguity for when the production team uses the JED

edsanville:
Ok, it uses text inputs and the buttons have been renamed for maximum unambiguity

ferrom-jaia (approved):
Looks good. Twomey tested on fleet22.

PR #727, opened 2023-12-09
Added CSV download button to JCC

edsanville:


src/web/command_control/client/components/SettingsPanel.tsx
```diff
@@ -64,6 +67,11 @@ export function SettingsPanel(props: Props) {
         }
     }
     
+    const handleDownloadCSV = async (event: React.MouseEvent<HTMLButtonElement>) => {
```
michael-jaia, line 98: Add function comment

src/web/shared/CSVExport.ts
```diff
@@ -0,0 +1,51 @@
+import {TaskPacket} from './JAIAProtobuf'
+import {LogTaskPacket} from './LogMessages'
+
+function isoString(t_micros: number | undefined): string {
```
michael-jaia: Function comments

src/web/shared/CSVExport.ts
```diff
@@ -0,0 +1,51 @@
...
+    if (t_micros == null) {
+        return ''
+    }
+
+    return new Date(t_micros / 1000).toISOString()
+}
+
+export async function getCSV(taskPackets: (TaskPacket | LogTaskPacket)[]) {
```
michael-jaia, line 26: Function comments

src/web/shared/CSVExport.ts
```diff
@@ -0,0 +1,51 @@
...
+
+        csvText += rowData.join(',') + '\n'
+
+        id += 1
+    }
+
+    return csvText
+}
+
+export function getCSVFilename(taskPackets: TaskPacket[]) {
```
michael-jaia, line 69: Function comments

src/web/shared/CSVExport.ts
michael-jaia: Can we make t_micros tMicros

michael-jaia: I'm not sure we need "| undefined" because typescript will not crash if an undefined is passed to a variable with a type of  number. It is explicit, but if we do it here then we will have to do it for all function declarations which seems unnecessary since it is a default property. I could also be missing the reason for it, so just let me know your thoughts.

src/web/shared/CSVExport.ts
```diff
@@ -0,0 +1,51 @@
+import {TaskPacket} from './JAIAProtobuf'
+import {LogTaskPacket} from './LogMessages'
+
+function isoString(t_micros: number | undefined): string {
+    if (t_micros == null) {
```
michael-jaia: We can make this clearer by saying !== undefined

src/web/shared/CSVExport.ts
michael-jaia: or even microseconds

src/web/shared/CSVExport.ts
```diff
@@ -0,0 +1,51 @@
...
+        return ''
+    }
+
+    return new Date(t_micros / 1000).toISOString()
+}
+
+export async function getCSV(taskPackets: (TaskPacket | LogTaskPacket)[]) {
+    var csvText = 'id,bot,task,lat,lon,time,bottom dive,depth achieved (m),current strength (m/s),current heading (deg)\n'
```
michael-jaia: let

src/web/shared/CSVExport.ts
```diff
@@ -0,0 +1,51 @@
...
+    var id = 0
```
michael-jaia: let

src/web/shared/CSVExport.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    for (const taskPacket of taskPackets) {
+        if ('_scheme_' in taskPacket) {
+            if (taskPacket._scheme_ != 1) {
```
michael-jaia: !==

src/web/shared/LogMessages.ts
```diff
@@ -0,0 +1,31 @@
+import { Command, TaskPacket } from "./JAIAProtobuf"
+
+
```
michael-jaia: Remove extra blank line

src/web/shared/CSVExport.ts
```diff
@@ -0,0 +1,51 @@
...
+                // Skip DCCL messages
+                continue
+            }
+        }
+
+        const start_location = taskPacket.dive?.start_location ?? taskPacket.drift?.start_location
```
michael-jaia: camelCase

michael-jaia (changes requested):

src/web/shared/CSVExport.ts
edsanville: Do we want camelCase for this one?  The rule I've been using is that I use snake_case for any variables that are in the messages, so we never get confused about camelCase versus snake_case for those variables.  However, this one's a local variable, so maybe that's a different scenario.

edsanville: I'll go to camelCase in the meantime.

src/web/command_control/client/components/SettingsPanel.tsx
```diff
@@ -64,6 +67,18 @@ export function SettingsPanel(props: Props) {
         }
     }
     
+    
+    /**
+     * Event handler for the CSV dowload button.  Creates the CSV file and initiates
+     * the download.
+     *
+     * @async
```
michael-jaia: Do we want this async tag in the function comments...if so, where should it be placed among the function comments for standardization

edsanville: We can remove it.  It could be useful, but it's not in our current specs for function comments.  Too bad the IDE doesn't show async on hover, because that would be useful in practice.

ferrom-jaia (approved):
Looks good. Tested in sim.

ferrom-jaia (approved):
Looks good! Tested in sim.

PR #728, opened 2023-12-14
feature/update-salinity-script-to-show-output

edsanville:


michael-jaia:
Closing this pull request since we need to rethink the way we calibrate salinity sensors. The process in general needs work so it can be viable in larger scale production and performed smoothly by customers.

PR #735, opened 2023-12-27
task/web-apps-build-tree

edsanville:
Limited to only commits directly related to building all artifacts into the build/<arch>/intermediate directory.

There SHOULD be nothing generated in the src tree.


src/python/build_venv.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+TARGET_DIR=$1
+JAIA_DIR="$(pwd)/../../"
+
+# Build the venv
+echo 🟢 Building the python venv into ${TARGET_DIR}
+
+## Make target python directory
+mkdir -p ${TARGET_DIR}
```
michael-jaia: Do we want the -p switch here?

src/python/build_venv.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+TARGET_DIR=$1
+JAIA_DIR="$(pwd)/../../"
+
+# Build the venv
```
michael-jaia, line 14: Should the # be the same length (I'm not sure if there is a reason for #, ##, ###)

michael-jaia: Should we also make them start with a capital or lowercase letter?

src/web/CMakeLists.txt
```diff
@@ -1,40 +1,64 @@
...
   )
-add_custom_target(command_control
+add_custom_target(jcc
   ALL
-  DEPENDS ${command_control_outdir}/client.js ${command_control_outdir}/favicon.png ${command_control_outdir}/index.html ${command_control_outdir}/manifest.json
+  DEPENDS ${jcc_out_dir}/client.js npm_dependencies
   )
-add_dependencies(command_control npm_depdendencies)
+
+# Copy the JCC server
```
michael-jaia: Make jcc lowercase to be like other instances

src/web/command_control/webpack.config.js
```diff
@@ -1,129 +1,113 @@
...
+    // cache: true,
+    parallel : true,
+  }) ],
+  mangleWasmImports : true
+};
+
+module.exports = (env, argv) => {
+  webpackOptions =
+      [
+        // Client
```
michael-jaia: Do we need these comments?

src/web/install_dependencies.sh
```diff
@@ -1,45 +1,11 @@
...
-if [ "$(find "package-lock.json" -maxdepth 0 -printf '%u\n')" == "root" ]; then
-    echo "Running with root privileges to remove package-lock.json."
-    sudo rm -rf "${NODE_MODULES_DIR}package-lock.json"
-fi
-
-cd $NODE_MODULES_DIR
-npm install --no-audit --quiet
+pushd ${PACKAGE_JSON_DIR} > /dev/null
+    # Shut up unless there's an error!
+    npm install --no-audit --no-progress --silent || npm install --no-audit --no-progress --quiet
```
michael-jaia, line 10: Will this hide warnings about packages that have been depreciated?

src/web/jdv/client/test.webpack.config.js
michael-jaia, line 1: Now that we removed the "scripts" property from the package.json...can we remove this file too?

src/web/run.sh
```diff
@@ -5,22 +5,32 @@ set -e
...
 
 # Build Command Control
-pushd command_control
-    npm install --no-audit
-    npm run test
-popd
+## Install pre-requisites
+./install_dependencies.sh ./
+
+## copy the webpack.config.js file to the intermediate build directory, so webpack can import from node_modules
```
michael-jaia: Capitalize c in copy

src/web/server/app.py
```diff
@@ -44,12 +46,12 @@ def parseDate(date):
 app = Flask(__name__)
 
 ####### Static files
-root = '../command_control/dist/client/'
-jed = '../jed/'
+jcc: str = os.path.join(args.appRoot, 'jcc')
+jed = os.path.join(args.appRoot, 'jed')
```
michael-jaia: Should : str also be added after jed for consistency?

michael-jaia (changes requested):

src/python/build_venv.sh
edsanville: It needs to create all parent directories to the path, in case they don't exist, but...  now that I think of it, it might be a problem if the venv already exists and we want to clobber it, (removing all the installed packages).

edsanville: If we want to clobber any pre-existing venv directories, that would be an additional "rm -rf ${TARGET_DIR}" just before the mkdir.

src/python/build_venv.sh
edsanville: I was going for a sort of "nested list" thing, like in markdown files.  I wonder if there's a better way to do it.  We should stick with capitalized though, I think.

edsanville: Trying indentation for script structure, instead of multiple hashes...

src/web/command_control/webpack.config.js
edsanville: Nope!

src/web/install_dependencies.sh
edsanville: Oooh, interesting point.  I tailored this script for the build.sh process, so I wanted it to be silent unless errors were present.  I eventually moved this build logic into CMakeLists.txt.

Perhaps we should catch deprecated packages when we're building for the web, using the run.sh scripts?

ferrom-jaia (dismissed):
Looks great to me! Tested on my machine. 
1. Removed build directory
2. ran build.sh
3. ran the run.sh in the web dir
4. ran the all.launch

michael-jaia (approved):

michael-jaia:
Built in new Ubuntu 22.04 VM and my current WSL system without errors. Launched JCC and JDV to see runtime environment.

PR #737, opened 2023-12-27
task/add-custom-command-should-use-complete-list-of-depends

edsanville:
Add jcc and jfv dependencies on the npm_dependencies target Fixed npm_dependencies typo
Removed deletion of ol-layerswitcher (it was only causing build issues because of a race condition in the parallel build process, due to missing jcc dependence upon the npm_dependencies target)


JRoybot (dismissed):
All looks good

ferrom-jaia (approved):
Looks good! Tested on my machine:

1. Removed build
2. Ran build.sh
3. Started sim

PR #739, opened 2023-12-31
task/wmm-current-year

edsanville:


jason-jaia:
Is there an issue if the file doesn't have the current year? If you're running old code for a few years on end?

edsanville:
> Is there an issue if the file doesn't have the current year? If you're running old code for a few years on end?

The model itself gets updated every 5 years, and in the meantime it will project values using "secular variation coefficients,"  which are basically just linear rates of change.  These magnetic field lines change so slowly (for the most part), that I doubt it would be a huge issue even a decade or more out.  The compass most likely has a lot more error to it than the WMM model will have.

The WMM code doesn't bounds-check the year value, so feeding it a year too far into the future won't cause a runtime error.

jason-jaia:
Nice - I was hoping you would say something like that.

https://blinq.me/LvMWEUQ6kZag

michael-jaia (approved):
Ed, nice calculation to keep the year up to date. I added a small comment for clarity. Test: 1707417812 / (365.24219 * 24 * 60 * 60) + 1970 = 2024.1059 where 1707417812 is equal to the current time in seconds since the epoch.

PR #741, opened 2024-01-08
task/diagram-of-build

edsanville:
We can add more details as requested.


ferrom-jaia (changes requested):
Can we update so it says share instead of shared in the path for web.

edsanville:
> Can we update so it says share instead of shared in the path for web.

Oh yes, of course!

ferrom-jaia (approved):
Looks good

PR #742, opened 2024-01-09
feature/taskpacket-dive-show-depth-value-only-if-bottom-strike

edsanville:
Added a "taskPacket" property to dive features, so we can access all the relevant properties through that reference


src/web/shared/Styles.ts
```diff
@@ -600,10 +600,17 @@ export function getRallyStyle(rallyFeatureCount: number) {
  */
 export function divePacketIconStyle(feature: Feature, animatedColor?: string) {
     // Depth text
-    let text = feature.get('depthAchieved') ? feature.get('depthAchieved').toFixed(1) : null
-    if (text != null) {
-        text = text + 'm'
-    } else {
+    const taskPacket = feature.get('taskPacket') as TaskPacket
+    const dive = taskPacket.dive as DivePacket
+    if (dive == null) {
```
michael-jaia: We can make this more explicit by saying if (dive !== undefined) since JS handles returns undefined when a property does not exist on an object or we can shorten it to if (dive) {}

src/web/shared/Styles.ts
```diff
@@ -600,10 +600,17 @@ export function getRallyStyle(rallyFeatureCount: number) {
...
+        return null
+    }
+
+    let text
```
michael-jaia: if we initialize text to "", we can remove the else statement below

src/web/shared/Styles.ts
```diff
@@ -600,10 +600,17 @@ export function getRallyStyle(rallyFeatureCount: number) {
...
+    if (dive.bottom_dive == true && dive.depth_achieved != null) {
```
michael-jaia: We can be more explicit here (=== true) and (!== undefined). Or we can shorten it to if (dive.bottom_dive && dive.depth_achieved).

src/web/command_control/client/components/CommandControl.tsx
```diff
@@ -41,7 +41,7 @@ import { divePacketIconStyle, driftPacketIconStyle, getRallyStyle } from './shar
 import { createBotCourseOverGroundFeature, createBotHeadingFeature } from './shared/BotFeature'
```
michael-jaia: For consistency on the github page let start using the branch name as the pull request name "feature/taskpacket-dive-show-depth-value-only-if-bottom-strike"

michael-jaia (changes requested):

src/web/shared/Styles.ts
edsanville: The shorter one will return false if depth_achieved is 0, though.  Going with the first option here.

michael-jaia:
https://github.com/jaiarobotics/jaiabot/assets/109538793/755a2b01-5e81-4d59-b9ab-03635f0e9727

ferrom-jaia (approved):
Looks great! Nice and simple implementation.

michael-jaia (approved):
Modified Ed's code, tested in sim (see video below), and pinged Matt to review as well.

PR #745, opened 2024-01-11
feature/add-echo-task-option-in-jcc

marshalllele:
Adds Listen task


src/web/command_control/client/components/Settings.ts
```diff
@@ -95,6 +95,11 @@ export let GlobalSettings = {
         drift_time: 10
     }),
```
michael-jaia: Add a space between // and Default

src/web/command_control/client/components/Settings.ts
```diff
@@ -95,6 +95,11 @@ export let GlobalSettings = {
...
 
+    //Default depth parameter for echo task
+    listenParameters: Load<ListenParameters>('echoParameters', {
+        listen_depth: 3
```
michael-jaia: What are the units here...let's add that in a comment. It might also make sense to create a variable defaultListenDepth = 3...and place the comment about units above that line

src/web/shared/JAIAProtobuf.ts
```diff
@@ -645,11 +646,16 @@ export interface ConstantHeadingParameters {
     constant_heading_speed?: number
 }
 
+export interface ListenParameters {
+    listen_depth?: number
```
michael-jaia: Do we want listen_depth to be an optional property?

src/web/command_control/client/components/Settings.ts
```diff
@@ -95,6 +95,11 @@ export let GlobalSettings = {
         drift_time: 10
     }),
 
+    //Default depth parameter for echo task
+    listenParameters: Load<ListenParameters>('echoParameters', {
```
michael-jaia: What is the difference between echoParameters and listenParameters? If they are the same let's be consistent with the naming...if they are different, a comment explaining the difference could be helpful

michael-jaia (changes requested):

src/web/shared/Styles.ts
```diff
@@ -8,6 +8,7 @@ import { PortalBotStatus } from './PortalStatus';
 import { colorNameToHex } from './Color'
 
 // We use "require" here, so we can use the "as" keyword to tell TypeScript the types of these resource variables
+declare var require : any
```
michael-jaia: What does this line do?

src/web/command_control/client/components/Settings.ts
marshalllele: updated

src/web/command_control/client/components/Settings.ts
marshalllele: updated - they were the same. Matt changed his mind about and decided the feature should be called "Listen" instead of "Echo" while I was still finishing up the code so I must have missed this instance.

src/web/command_control/client/components/Settings.ts
marshalllele: Updated

src/web/command_control/client/components/Settings.ts
marshalllele: updated - they were the same. Matt changed his mind about and decided the feature should be called "Listen" instead of "Echo" while I was still finishing up the code so I must have missed this instance.

src/web/command_control/client/components/Settings.ts
marshalllele: Updated

src/web/shared/JAIAProtobuf.ts
marshalllele: We do not. Updated it to be a mandatory property.

src/web/shared/Styles.ts
michael-jaia: Let's figure out this TypeScript quirk before merging

michael-jaia:
To resolve the TypeScript require warning, we need a TS config file in the shared dir. It seems a little repetitive to have a TS config file for JCC, JDV, and now the shared dir. I imagine there is a way we can place the TS config file in src/web that so it can be shared  among the directories that need it.

michael-jaia (approved):
## Approving Review
Approving this pull request knowing there are outstanding changes need to the code structure. These pertain to code cleanup rather than functionality/bug fixes. Given this and the business requirements, I am going to approve and merge this pull request into 1.y so we can move forward with the on the water testing and ship the bots. Subsequent tickets will be created for the next sprint to construct a maintainable and readable infrastrucutre.

PR #746, opened 2024-01-12
feature/add-echo-option-to-turn-on-of-in-beta-engineering-panel

marshalllele:
Adds On/Off checkbox to JED page


src/web/command_control/client/components/Settings.ts
```diff
@@ -1,5 +1,5 @@
 // Saving and loading settings from browser's localStorage
-import { ConstantHeadingParameters, DiveParameters, DriftParameters, Speeds } from "./shared/JAIAProtobuf"
```
michael-jaia: Change the branch name to match the name on Jira (feature/add-echo-opt...)

src/web/jed/style.css
```diff
@@ -419,3 +419,69 @@ input.manualOrPidButton:hover {
   border-width: 3pt;
   border-color: red;
 }
+
+/* JED Task Settings */
```
michael-twomey, line 423: What is all this CSS used for? In JED, I just see a simple checkbox

src/web/jed/index.html
```diff
@@ -204,8 +204,21 @@ <h3>Rudder</h3>
       <image src="helpPane.png">
     </div>
 
+    <div>
+      <h2>Waypoint Task Settings</h3>
```
michael-jaia: What do you think about "Waypoint Settings"...seems a little crisper to me

src/web/jed/style.css
michael-jaia: What is all this CSS used for? In JED, I just see a simple checkbox

michael-jaia (changes requested):

PR #747, opened 2024-01-13
bug/bottom-detection-wave-detection-race-condition

edsanville:
I added a python threading.Lock object to try to limit access to the IMU.  If this doesn't solve the problem, then it's a frequency of query issue, which we can solve perhaps using a timer to limit query frequency.

Also fixed an issue where an irfft of a length-1 rfft would trigger an exception for one cycle.


ferrom-jaia (approved):
Looks great! Tested on the wave simulator machine.

PR #751, opened 2024-01-18
bug/catch-duplicate-task-packets

michael-jaia:
We were seeing duplicate Task Packets appearing in the JCC after a data offload. This is because of the difference in precision between Task Packet times sent over the radio and those written to the .taskpacket file. We first tried a 1 second epsilon value but saw edge cases where duplicates could appear, so we increased the epsilon to 10 seconds. 

This pull request was tested in sim by performing a task (at 1x warp speed) and moving the .taskpacket file to the /var/log/jaiabot/bot_offload dir to simulate a data offload. To verify the pull request works, there should be no Task Packets directly on top of each other.


edsanville:
It seems like the edge cases are being caused by the same task packet getting rounded up or down, depending on the lost precision?  If so, the increase to 10 seconds will decrease the frequency of the edge cases, but not eliminate them.

I suggest rounding all start_times down, but also checking for the existence of a taskpacket at the +/-1 interval keys.

Thoughts?

michael-jaia:
Yeah that is the reason we are seeing duplicate TaskPackets. With rounding down, couldn't that result in changes of the time greater than 10 seconds for comparison purposes?

src/web/server/jaia_portal.py
```diff
@@ -46,6 +46,49 @@ def protobufMessageToDict(message):
     return google.protobuf.json_format.MessageToDict(message, preserving_proto_field_name=True)
 
 
+def filterDuplicateTaskPackets(taskPackets: list[dict]):
```
ferrom-jaia: Tried to run in Sim and I got this Error:
Traceback (most recent call last):
  File "./app.py", line 22, in <module>
    import jaia_portal
  File "/home/ferrom/repos/jaiabot/src/web/server/jaia_portal.py", line 49, in <module>
    def filterDuplicateTaskPackets(taskPackets: list[dict]):
TypeError: 'type' object is not subscriptable

I have python version 3.8 and the bots do as well.

ferrom-jaia (changes requested):

src/web/server/jaia_portal.py
edsanville: Ah, yikes...  just need to import typing and use List instead of list.

edsanville: I should probably stick with that old way of type hinting until the world updates to the later 3.x versions of python...

michael-jaia:
Ed nice implementation. I moved the reduceTime function outside of the filter function to keep the filter code brief and easy to follow. Tested in sim by running two survey missions with dives and drifts. I downloaded the data to check for duplicate TaskPackets appearing in the JCC; I did not see any. To verify TaskPackets were coming from the bot_offload directory I used the Edit Dates feature to display past TaskPackets.

ferrom-jaia (approved):
Looks great! Witnessed Twomey test pull request on his machine.

PR #753, opened 2024-01-20
feature/simulator-seafloor-eggshell

edsanville:
Configurable egg crate function (sin(jx) * sin(ky)), to allow easy testing of the bottom dive and contour code in the simulator.


edsanville:
During simulator testing, I found that the bots wouldn't dive below 30m, (a timeout in the dive code triggered maybe?).  So, I set the default seafloor_depth to 15m, with eggshell wavelength 100m, and amplitude 5m.

PR #754, opened 2024-01-20
feature/simulator-seafloor-eggshell

edsanville:
Use egg crate function for seafloor depth in simulator, for easy contour map testing.


michael-jaia (changes requested):
Hey Ed, this is a cool pull request. What do you think of moving the calculation of seafloor depth to its own function so a function comment can be added explaining a little more about the calculation.

michael-jaia:
Screenshot (31)
Looks really cool Ed!

michael-jaia (approved):

PR #758, opened 2024-01-23
bug/reduce-time-to-process-task-packets

michael-jaia:
Aimed at resolving the hub slowdown we have been seeing this week when the number of TaskPackets approaches 1,000 and the hub has been running for over an hour. This pull request creates a list called all_task_packets to reduce the duplicated processing operations that came from storing offloaded TaskPackets and live TaskPackets in separate lists. It also prevents the *.taskpackets files from being read every 10 seconds and handles duplicates in one location to reduce the computation in the get_task_packet_count and get_task_packets functions.


src/web/server/jaia_portal.py
```diff
@@ -427,47 +417,41 @@ def process_active_mission_plan(self, bot_id, active_mission_plan):
...
+            return self.all_task_packets[start_index:]
         
         end_index = bisect.bisect_right(
-            list(map(lambda task_packet: int(task_packet['start_time']), task_packets_list)),
+            list(map(lambda task_packet: int(task_packet['start_time']),  self.all_task_packets)),
             utime(end_date)
         )
-        return task_packets_list[start_index: end_index]
+        
+        return self.all_task_packetst[start_index: end_index]
```
michael-jaia: Fix typo!

src/web/server/jaia_portal.py
```diff
@@ -56,33 +56,23 @@ class Interface:
     # Dict from bot_id => engineeringStatus
```
michael-jaia, line 98: Check offload config

michael-jaia:
When this PR is merged, we need to ping Josh Mickles to let him know the JCC has access to all of the TaskPackets on the hub

ferrom-jaia (approved):
Looks great! Tested in sim!

PR #759, opened 2024-01-24
task/setup-dynamic-radio-config

ferrom-jaia:
This code will read in the xbee config file to get the number of bots to dynamically setup up bins that allow the bots and hubs to talk together.


michael-jaia (approved):
Looks good, watched bench test with 29 bots where status ages matched the bin structure.

PR #760, opened 2024-01-24
task/detect-bot-roll

ferrom-jaia:
Added back the bot roll detection so the bot continues on the same path


michael-jaia (approved):
Looks good, watched Matt test on bench

PR #761, opened 2024-01-24
task/exclude-downloaded-taskpackets-to-reduce-server-load

ferrom-jaia:
Exclude downloaded taskpackets for now until we resolve server load issue


michael-jaia (approved):
Looks good, a needed change for testing and until we merge more efficient code for TaskPacket processing

PR #762, opened 2024-01-24
task/remove-dive-init-start-code

ferrom-jaia:
Removing initial dive constant throttle code


michael-jaia (approved):
Reviewed side by side with the original changes from task/change-reverse-start-bound-when-diving-to-match-dive-throttle #755

PR #763, opened 2024-02-01
feature/jdv-refresh-data-from-new-file

edsanville:


michael-jaia (changes requested):
Hey Ed, I just did some testing and I noticed the close button for individual graphs is no longer working. I'll attach a video below:

https://github.com/jaiarobotics/jaiabot/assets/109538793/5c1b5961-7abd-4c27-92b5-40a07cf0630a

michael-jaia (approved):
Re-tested with old log files. 1) The graphs clear out when new logs are opened.  2) The map displays the new log data (while keeping the map settings the same.) 3) The user is able to delete a single graph.

PR #768, opened 2024-02-06
task/clean-up-gitignores

michael-jaia:
After cleaning up the gitignores, I made changes in C++, Python, and TypeScript (JCC and JDV) and checked which changes were being tracked. Only the files I modified were being tracked. I also saw the same expected output after running the build script. Lastly, I ran the simulation to test git-ignoring the all.launch file.

Note: I added npx webpack to the build scripts to allow me to proceed with testing because I was getting a webpack command not found error. (These changes will be removed when we merge in Ed's pull request that deletes the jcc and jdv build scripts)


ferrom-jaia (approved):
Looks great!

PR #769, opened 2024-02-06
task/stop-auto-assign-with-degraded-comms

marshalllele:
TESTING: To test you can change the return statement in getBotStatusAge() to "return botId == 1 ? 32 : statusAge;" - this will simulate a status age > 30 for Bot 1, but return the true status age of each other bot in the simulator.


michael-jaia (approved):
Performed a pair code review with Nick which included a static review of the code + a stress test in sim.

PR #772, opened 2024-02-06
bug/mui-material-update-bug

ferrom-jaia:
Locking mui material node dependency until a fix is released for the new version 5.15.8


jason-jaia:
This one doesn't need a caret?

ferrom-jaia:
No longer an issue

PR #773, opened 2024-02-07
bug/update-docker-arm64-build-and-deploy-script

edsanville:
To avoid hitting disk cap, clear out the old code when we use docker_arm64_build_and_deploy.sh
Remove all jaiabot Deb packages when running docker_arm64_build_and_deploy.sh
- Before we copy new packages
- sudo apt remove *jaiabot*
Fix missing path for offload script 
Install correct metadata for version reported by jaiabot-status.sh
- Should be version of the deployed packages
Check the goby and dccl dependencies 
- Version in docker container match target


tsaubergine:
I think this is a good step in the right direction. I'm wondering if docker-arm64-build-and-deploy can directly leverage (call) the scripts from jaiabot-embedded and leverage debconf to avoid duplication of code (calling systemd.py, install python venv, etc):
- https://github.com/jaiarobotics/jaiabot-debian/blob/1.y/jaiabot-embedded.prerm 
- https://github.com/jaiarobotics/jaiabot-debian/blob/1.y/jaiabot-embedded.postinst
- https://github.com/jaiarobotics/jaiabot-debian/blob/1.y/jaiabot-python.postinst

edsanville:
> I think this is a good step in the right direction. I'm wondering if docker-arm64-build-and-deploy can directly leverage (call) the scripts from jaiabot-embedded and leverage debconf to avoid duplication of code (calling systemd.py, install python venv, etc):
> 
> * https://github.com/jaiarobotics/jaiabot-debian/blob/1.y/jaiabot-embedded.prerm
> * https://github.com/jaiarobotics/jaiabot-debian/blob/1.y/jaiabot-embedded.postinst
> * https://github.com/jaiarobotics/jaiabot-debian/blob/1.y/jaiabot-python.postinst

Possibly?  But since it's in a different repository, it would be more difficult to leverage I think.

I wonder if the cleanest way to deploy would be to actually generate the .deb files, rsync them, and then install on the remote.  Then, there would only be one (essentially) build and deploy pathway to maintain?

Edit:  on second thought, this would probably be super-slow, and the point of the deploy script is to quickly see results.

edsanville:
Deploying now writes the built version to /etc/jaiabot/software_version, which is read by jaiabot-status script.
This pull request should be used in conjunction with the jaiabot-debian PR of the same name.
The jaiabot-debian postinst script also writes the package version to the /etc/jaiabot/software_version file.

ferrom-jaia (approved):
Looks good! Tested on hub 1 fleet 1.

PR #774, opened 2024-02-07
task/change-hub-link-icons

JRoybot:


michael-jaia (commented):
Looks good Jeff. I showed it to Michael and Ian, and they recommended we use the wifi with a cog icon, so I made that change.

michael-jaia (approved):

PR #775, opened 2024-02-08
bug/authentication-required-to-restart-jaiabot-imu-py-service

JRoybot:
In response to https://jaia-innovation.atlassian.net/jira/software/projects/JAIAB/boards/2?selectedIssue=JAIAB-856

This change simply ignores any IMU Issues when running in Simulation.  I tested this with 4 bots at warp 15.  Even though the systems resources were pegged at 100% and the machine was very sluggish there were no blocking pop-ups and the simulation ran for the most part.  Because there were still gaps in reported health from the bots they would turn yellow or red at times.  I ran a mission with a few dives and station keeping and most of the tuns at least 2 of the bots could complete the mission.  I think this is worth putting in place as it prevents the system from getting locked by stacked pop up dialogs.

### DIFF
--- a/src/bin/health/app.cpp
+++ b/src/bin/health/app.cpp
@@ -190,8 +190,15 @@ jaiabot::apps::Health::Health()
             {
                 case protobuf::IMUIssue::STOP_BOT: break;
                 case protobuf::IMUIssue::RESTART_IMU_PY:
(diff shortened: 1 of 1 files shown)


ferrom-jaia (approved):
Looks good. Tested in sim. While the vehicles were running triggered an IMU issue and verified it in the logs.

PR #778, opened 2024-02-12
bug/playall-not-working-after-reassigning

ferrom-jaia:
## Summary:
This pull request fixes the setRcMode function so that it handles botIds that are equal to -1. -1 is used to identify runs that are not assigned. If you attempted to click the playall button with an unassigned run preceding assigned runs then it would not work as expected.

## Details:

Function: setRcMode(botId: number, rcMode: boolean)
Fixed setRcMode error when botId==-1

## Testing:

1. Create two runs 
2. Assign a bot to one of the runs 
3. Click play all button
4. Bot should execute the assigned run


michael-jaia (approved):
Nice fix, tested in sim, approving PR

PR #779, opened 2024-02-13
task/change-style-of-panel-headings

michael-jaia:
## Summary
Removes the white background from the titles of the JCC panels to create a cleaner, more consistent look and adds a top margin to the Hub Details panel to match the Bot Details. 

## Details
The panelHeading class in `src/web/command_control/client/style/CommandControl.less` controls the style for the panel headings (excluding the Hub and Bot Details). I added a top margin to the Hub Details title to match the spacing in the Bot Details panels. 

## Testing
Clicked through all of the panels to visually inspect the appearance of the titles on desktop and tablet screen-sizes.

## Notes
All of the changes to `src/web/command_control/client/components/Details.tsx` come from removing a duplicate div to correctly set the margin-top. Removing the duplicate div requires a layer of indentation to be removed as well which is why git detected so many changes in that file.

### DIFF
--- a/src/web/command_control/client/style/CommandControl.less
+++ b/src/web/command_control/client/style/CommandControl.less
@@ -52,7 +52,6 @@ td {
   div,
   label {
     font-family: inherit;
-    letter-spacing: 0;
     line-height: 1;
   }
 
@@ -1349,10 +1348,10 @@ div.closeButton {
   margin: 4pt 8pt 0 8pt;
   padding: 2pt;
   
-  background-color: #FFFFFF;
-
+  color: #FFFFFF;
   font-size: 1.1rem;
   font-weight: bold;
+  letter-spacing: 0.5px;
   text-align: center;
 }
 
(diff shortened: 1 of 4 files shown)


src/web/command_control/client/style/CommandControl.less
```diff
@@ -1349,10 +1348,10 @@ div.closeButton {
   margin: 4pt 8pt 0 8pt;
   padding: 2pt;
   
-  background-color: #FFFFFF;
-
+  color: #FFFFFF;
```
ferrom-jaia: Should we use a variable here for the color? Util.less is probably the spot for this

ferrom-jaia (commented):
Looks nice!

src/web/command_control/client/style/CommandControl.less
michael-jaia: Just added a variable...there is a little overlap with the variable names, but I think cc-text-color-light (I would prefer jcc-text-color-light but that is another ticket) is more reusable than the variable names specific for the panels (for instance there is @cc-accordian-text-color: white; but I find that a little confusing because it gets into the definition of an accordion (is it just the gray bar or does it include the drop down too which is black text)

ferrom-jaia (approved):
Looks nice!

PR #782, opened 2024-02-15
feature/update-mission-proto-to-include-srp-fields

michael-jaia:
## Summary
* Moves the BottomDepthSafetyParams message definition from engineering.proto to mission.proto in preparation of setting those values in the Optimize Mission Planning panel.

## Details
* Move the BottomDepthSafetyParams message to mission.proto
* Imports mission.proto into engineering.proto to prevent the engineering app from breaking
* Adds the bottom depth safety parameters check for the first fragment of the command message sent over the radio

## Testing
* Forced the SRP into action in sim by setting the seafloor depth to 5m, the dive depth to 7m, and safety depth to 10m (added the buffers to account for dive variability and the egg box shaped seafloor). After the dive, the bot entered its constant heading state as part of the SRP protocol.

### DIFF
--- a/src/bin/hub_manager/hub_manager.cpp
+++ b/src/bin/hub_manager/hub_manager.cpp
@@ -454,6 +454,12 @@ void jaiabot::apps::HubManager::handle_command(const jaiabot::protobuf::Command&
                 command_fragment.mutable_plan()->set_repeats(command.plan().repeats());
             }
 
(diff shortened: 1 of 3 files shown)


ferrom-jaia (commented):
I think you missed handling this in mission_manager

ferrom-jaia (approved):
Looks great!

PR #731, opened 2023-12-20
Feature/aws cloud

tsaubergine:
Implements CloudHub and VirtualFleet as envisioned in https://github.com/jaiarobotics/jaiabot/blob/c84c1faad5cff8156cb4c29650e406bd221e4a6d/src/doc/figures/jaia-fleet-in-the-cloud.png

Requires: https://github.com/jaiarobotics/jaiabot-rootfs-gen/pull/54
Related: https://github.com/jaiarobotics/private_jaia/pull/19
Requires: Goby 3.1.2 for IPv6 support which is in staging on packages.jaia.tech (and staging -> test only)

Video: https://jaiarobotics.slack.com/archives/C0350PYB22E/p1703057228553939
Cloud Markdown doc: https://github.com/jaiarobotics/jaiabot/blob/b94dbb7c6958d876316d0246dadccdec8f2f5bdb/src/doc/markdown/page56_cloud.md
Internal wiki: https://jaia-innovation.atlassian.net/wiki/spaces/PD/pages/1166802945/Cloud

Still To Do (I haven't made JIRA cards for these):

JCC
  - CloudHub/VirtualHub/regular hub visual distinction
  - Data offload button to CloudHub (and run jaiahub-dataoffload-cloudhub.sh)
  - Handoff between CloudHub / regular hub?
  - Links to JDV and Liaison prelaunch are hardcoded to 10.23 IP addresses. Need to change to use same hostname (or IP) that JCC is being accessed through.

JDV
 - On CloudHub needs to run goby_log_tool in a different directory from `/var/log/jaiabot/bot_offload` (`/var/log/jaiabot` would work fine) and move resulting h5 file into `/var/log/jaiabot/bot_offload`. This is due to the limits of s3fs and the permissions set (CloudHub can't delete data in the bucket, by design).

PR #783, opened 2024-02-15
Task/apt cacher

tsaubergine:
# Add hub cache mode for jaiabot/system package updates

First, see [updated documentation](https://github.com/jaiarobotics/jaiabot/blob/83e828369db3162deb9230d8497bf90ed5d12d28/src/doc/markdown/page20_build.md#alternative-debian-package-update-options).

This PR adds a third mode ("online with hub caching") to the existing "online" (normal) / "offline" (CD) update modes. It also fixes some bugs with the offline mode that had been introduced and not detected (since it has not regularly been used).

This new mode "online with hub caching" is for systems that can be internet connected but have low or metered bandwidth available, or are large fleets.  

I'm not convinced we should use "online with hub caching" with all fleets yet as it does introduce complication ("online" is definitely simpler - no proxies, no caching, etc.). If we get some data on the ground that suggests  "online with hub caching" is better and reliable, we can eventually consider making that the standard mode.

I initially tried using `apt-cacher-ng` but was running into annoying and non-reproducible issues (which the internet has also seen...) especially with concurrent hits to the same deb package (which we're doing all the time here). I then switched to `squid-deb-proxy` instead which seems rock solid so far in my testing.

## Related PRs

Required: https://github.com/jaiarobotics/jaiabot-debian/pull/51
Optional (but good for cleanliness): https://github.com/jaiarobotics/jaiabot-rootfs-gen/pull/56

## Changes

This PR is largely about modifying existing ansible playbook files:

- config/ansible/change-sources.yml: Expand the scope of this playbook to change the mode (online/offline/online with hub caching) and release version (only 1.y for now, but placeholder for future 2.y, 3.y, etc.) in addition to repo (continuous,release, etc.).
  - This playbook sets up the required configuration for the hub to act as an apt caching proxy (using [squid-deb-proxy](https://fabianlee.org/2018/02/08/ubuntu-a-centralized-apt-package-cache-using-squid-deb-proxy/)). The bot and hub sources.list are the same as for the normal "online", but a proxy (the hub) is added using the `/etc/apt/apt.conf.d/00aptproxy` file.
- config/ansible/install.yml: 
  - When using "online with hub caching", the jaiabot packages are downloaded on the hub first (to allow `squid` to cache these packages) then installed normally on the hub/bots (and the downloads are pulled from the `squid` cache).
  - When using "online with hub caching", the python packages in jaiabot-python's `requirements.txt` are cached to the hub at /var/www/html/updates (the same location as the "offline" mode when the CD is mounted) so they can be pulled with pip from the hub rather than each bot/hub pulling from pypi.org.
- config/ansible/upgrade-all.yml: 
  - Now a superset of config/ansible/install.yml to simplify development/debugging; first it installs jaiabot packages (same as config/ansible/install.yml)
  - When using "online with hub caching", the `apt-get upgrade` packages are downloaded on the hub first (to allow `squid` to cache these packages) 
  - Then it does an `apt-get upgrade`
  - Important: upgrade-all.yml no longer does "apt dist-upgrade" but rather does an "apt install" of the jaiabot packages and goby packages (which allows for add/remove of dependencies if necessary) and then does a normal "upgrade" - this should reduce the odds of breaking the customer's system (I found a case where an offline CD-based dist-upgrade removed openssh-server, which was annoying to say the least!).

Other changes:

- Ansible config set to use 25 forks so that up to 25 bots are upgraded in parallel (default is 5 in ansible). This may be able to be set higher (depends on memory/CPU of the hub). This change is cherry picked from https://github.com/jaiarobotics/jaiabot/pull/731 as it will greatly improve the upgrade speed of large fleets (as Ansible won't wait for 5 bots to complete before starting the next 5). Note that this will improve the upgrade speed in many cases for the normal "online" mode since it's not just the download takes time (package unpack/setup can be lengthy too, depending on the number of upgraded packages).
- Updated `requirements.txt` for jaiabot-python to include the full list of dependencies captured when installing the old `requirements.txt` on a focal machine and then using `pip3 freeze`. I'd like to ensure we keep this as reproducible as possible moving forward.
- Minor fix to src/web/jcc.conf to remove the trailing slash so that `http://hub0/updates` redirects correctly, in addition to `http://hub0/updates/`, which removes a potential point of confusion.

## Testing

I built this branch into the test/X.y debian repo.

I extensively tested this with 2 virtualbots and 1 virtualhub from an Aug 9, 2023 image (to ensure it had a a substantially out-of-date system running jaiabot 1.6). I have not tested this on Raspi bots/hubs, but I expect it will work OK.

- Tested 'online' -> 'online with hub cache' with large update
- Tested 'online' -> 'offline' with this same large update
- Tested 'offline' -> 'online with hub cache'

I do expect we may find edge cases switching between these modes, as much as I tried to eliminate them. I expect in most cases we can switch a customer to a particular update mode and stick to that.


ferrom-jaia (approved):
I read through the description and the documentation. Looks great and the testing by Toby seems to be sufficient to merge into 1.y. We realize there may be edge cases, but this puts us in a better spot than we currently are in for doing upgrades. Thanks Toby! Looks great!

PR #784, opened 2024-02-15
feature/add-srp-fields-in-omp

michael-jaia:
## Summary
* Allows operators to edit the Safety Return Path (SRP) fields from the Optimize Mission Planning panel. 

## Details
* Adds the SRP input to the MissionPlan property of the Command for each run when Apply is clicked in the Optimize Mission Planning panel.

## Testing
* Created a survey mission, edited the SRP fields, and pressed Play All. I then queried to status of the bots in the engineering panel to see that the SRP values updated
* Ran a survey mission with the seafloor set to 5m, the dive set to 7m, and safety depth set to 10m. This caused the bots to enter there constant heading state and then move to the last waypoint in the mission.

## Notes
* Pull Request #782 should be merged in prior to this PR review


src/bin/mission_manager/app.cpp
```diff
@@ -953,6 +947,27 @@ bool jaiabot::apps::MissionManager::handle_command_fragment(
     return false;
 }
 
+void jaiabot::apps::MissionManager::handle_bottom_dive_safety_params(
```
ferrom-jaia, line 959: Can we add a function comment here?

ferrom-jaia: If you need help finding the documentation: https://www.doxygen.nl/manual/docblocks.html

ferrom-jaia (approved):
Looks great! Tested in sim!

PR #787, opened 2024-02-16
task/jcc-button-layout

JRoybot:
Moved various map settings buttons around based on original description in JAIAB-865 and further direction from Matt.  Please provide a critical review.  This web stuff is all pretty new to me.


src/web/command_control/client/components/SettingsPanel.tsx
```diff
@@ -49,6 +57,16 @@ export function SettingsPanel(props: Props) {
         props.setClusterModeStatus(!props.isClusterModeOn)
     }
 
+    const handleTrackPodToggleClick = () => {
```
ferrom-jaia, line 67: Can we add a function comment for consistency moving forward: https://github.com/jaiarobotics/jaiabot/blob/1.y/src/doc/markdown/page110_coding_standards.md

ferrom-jaia (changes requested):
Nice job Jeff.

A couple of things to review and see what you think.

The Settings panel goes to the very bottom of the page:
image

I think we should limit it and give it some room away from the bottom of the page like the Mission Settings Panel:
image

Do you like the added space I am posting below between the trackbot toggle and the accordions?
image

ferrom-jaia (approved):
Looks good! Tested in sim.

PR #788, opened 2024-02-18
feature/first-wpt-task-option-omp

michael-jaia:
## Summary
* Allows an operator to set a start task when using the optimize mission planning tool. This start task takes place at the location of the start rally point.

## Details
* The implementation mostly mimics the structure of the end task option. The difference comes from setting the start task to occur at the start rally as opposed to the beginning of the survey line. (For comparison, the end task occurs at the last waypoint in the survey line rather than the end rally).

## Testing
* Planned a survey line with various start tasks (with and without a start rally point selected to check the error handling) and I did not produce any errors.
* Ran two survey missions with 2 bots. The first mission began with a constant heading and the second began with a dive. In both missions, the bots performed the expected steps.


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #789, opened 2024-02-20
bug/undo-button

marshalllele:
Undo button will undo changes made to task types and waypoints, but will not remove any changes made to task parameters.


JRoybot (commented):
I tested this branch on a clean VM.
Whenever Undo is used on the last waypoint it deletes the waypoint.  
This happens even if the waypoint is created, other work is done and then the last waypoint is edited to add a task (ie Dive), then Undo deletes the waypoint.

If a task is added to a waypoint that is not the last one the Undo button will just delete the task, not the waypoint.

Under all cases if I changed any task parameters and then used Undu the task will be deleted (or the entire waypoint if it's last).  Subsequently all new tasks of that type will have the parameters of the deleted task.  This seems global, across waypoints and across bots.

JRoybot (dismissed):
Reviewed and retested with latest commits.  All works as expected.

src/web/command_control/client/components/CommandControl.tsx
```diff
@@ -1090,39 +1090,23 @@ export default class CommandControl extends React.Component {
 	 * then the mission history is updated with the new task. This way, the undo button can work as expected. Does not track changes in
```
ferrom-jaia: Can we update the function comments to remove the param.

src/web/command_control/client/components/CommandControl.tsx
```diff
@@ -3160,7 +3144,12 @@ export default class CommandControl extends React.Component {
 		}
 
 		this.setState({ visiblePanel: panelType })
-		this.taskChangeTester(this.getRunList())
+		this.taskChangeTester()
+		/*
```
ferrom-jaia: Remove commented code

src/web/command_control/client/components/TaskSettingsPanel.tsx
```diff
@@ -272,7 +272,7 @@ export function TaskSettingsPanel(props: Props) {
 
         // Change to NONE
         if (newTaskType == TaskType.NONE) {
-            props.onChange(null)
+            props.onChange(undefined)
```
ferrom-jaia, line 338: What is the reason to change this from null to undefined?

src/web/command_control/client/components/CommandControl.tsx
ferrom-jaia: Also since we edited the function can we update it to match: 
https://github.com/jaiarobotics/jaiabot/blob/1.y/src/doc/markdown/page110_coding_standards.md

ferrom-jaia (changes requested):
Looks great!

Just a couple of things to add before merging in.

Tested multiple combinations to try and break it.

src/web/command_control/client/components/CommandControl.tsx
```diff
@@ -1079,6 +1079,38 @@ export default class CommandControl extends React.Component {
 		this.missionHistory.push(deepcopy(mission))
 	}
 
+	/**
+	 * Used to differentiate between a state change and a task/mission change. Called from onDoneClick() in Render(). Calls updateMissionHistory() if the task of a waypoint changes.
```
michael-jaia: What do you think of reducing this comment to just the last sentence?

src/web/command_control/client/components/CommandControl.tsx
```diff
@@ -1079,6 +1079,38 @@ export default class CommandControl extends React.Component {
...
 	}
 
+	/**
+	 * Used to differentiate between a state change and a task/mission change. Called from onDoneClick() in Render(). Calls updateMissionHistory() if the task of a waypoint changes. 
+	 * 
+	 * @param {MissionInterface} mission Provides current mission to test if task type has changed between the previous and latest missions
+	 * @returns {void}
+	 * 
+	 * @notes
+	 * Tests to see if there was a change in the type of task for each waypoint. If the done button was pressed in the goal settings panel,
```
michael-jaia: Something similar here, what do you of reducing this to the final sentence?

michael-jaia (changes requested):

src/web/command_control/client/components/CommandControl.tsx
marshalllele: Done! Removed the @param key.

src/web/command_control/client/components/CommandControl.tsx
marshalllele: Done. Removed commented code and updated the description following the Jaia Coding Standards.

src/web/command_control/client/components/TaskSettingsPanel.tsx
marshalllele: The taskChangeTester() function in CommandControl gets pretty messy when testing for both a TaskType of 'Undefined' (the value before the task is set) and 'Null' (the value if the task is changed back to 'None'). Changing this line from 'null' to 'undefined' makes it so that when the user changes the task to 'None', the value is represented as 'Undefined' again. 
I can change it back and modify the taskChangeTester() if that works better for the rest of the code. Just let me know!

src/web/command_control/client/components/CommandControl.tsx
marshalllele: I like it, done!

src/web/command_control/client/components/CommandControl.tsx
marshalllele: Done.

src/web/command_control/client/components/CommandControl.tsx
```diff
@@ -1078,6 +1078,34 @@ export default class CommandControl extends React.Component {
...
 	}
 
+	/**
+	 * Provides current mission to test if task type has changed between the previous and latest missions
+	 * @returns {void}
+	 * 
+	 * @notes
+	 * Does not track changes in task parameters, only in task type. 
+	 */
+	taskChangeTester() {
```
ferrom-jaia: Can we update this function name to be detectTaskChange instead of taskChangeTester.

ferrom-jaia (changes requested):

michael-jaia (approved):
Easy to follow implementation, nice work Nick. Tested in sim with various combinations of tasks, assigned runs, unassigned runs, duplicate runs, survey missions, and different edit mode states.

ferrom-jaia (approved):
Looks great!

PR #790, opened 2024-02-20
task/handle-retrofit-imu-reset

ferrom-jaia:
# Summary
Adding configuration for imu installation type
# Details
The approach for the imu installation type is similar to the imu type configuration 
# Testing
The GPIO pin for both the retrofit and the embedded IMU has been tested
# Notes
Please merge this pull request in first: https://github.com/jaiarobotics/jaiabot-debian/pull/53

### DIFF
--- a/src/python/adafruit_BNO085/jaia_firm_bno085_reset_gpio_pin.py
+++ b/src/python/adafruit_BNO085/jaia_firm_bno085_reset_gpio_pin.py
@@ -1,8 +1,19 @@
 import RPi.GPIO as GPIO
 import time
+from enum import Enum
+import argparse
 
-# BNO085 IMU reset pin
-bno085_pin = 10
+parser = argparse.ArgumentParser(description='Turn on/off bno085 imu', formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+parser.add_argument('--imu_install_type', choices=['embedded', 'retrofit', 'none'], help='If set, configure services for imu install type')
+args=parser.parse_args()
+
+# Set BNO085 IMU reset pin
+if args.imu_install_type == 'embedded':
+    bno085_pin = 10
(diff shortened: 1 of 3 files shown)


src/python/adafruit_BNO085/jaia_firm_bno085_reset_gpio_pin.py
```diff
@@ -1,8 +1,19 @@
...
+import argparse
 
-# BNO085 IMU reset pin
-bno085_pin = 10
+parser = argparse.ArgumentParser(description='Turn on/off bno085 imu', formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+parser.add_argument('--imu_install_type', choices=['embedded', 'retrofit', 'none'], help='If set, configure services for imu install type')
+args=parser.parse_args()
+
+# Set BNO085 IMU reset pin
+if args.imu_install_type == 'embedded':
```
michael-jaia: Not sure if it's important here, but could could we remove the elif by only checking for retrofit with the if and then everything else is 10

michael-jaia (commented):
Just left one comment...not sure if it needs to be changed just something I noticed. I also did not know variables defined within python if statements are not closed scope to within that block

ferrom-jaia:
> Just left one comment...not sure if it needs to be changed just something I noticed. I also did not know variables defined within python if statements are not closed scope to within that block

Ya, I think we can move the variable outside to make it more readable

src/python/adafruit_BNO085/jaia_firm_bno085_reset_gpio_pin.py
ferrom-jaia: Yes, I will update

michael-jaia (approved):
Comments resolved, looks good, approving PR into 1.y

PR #791, opened 2024-02-20
task/handle-data-offload-options

ferrom-jaia:
# Summary
Adding config for data offload type

# Details
The approach for the data offload type is similar to the imu type configuration

# Testing

This will need to be tested on a bot after merging in. 

# Notes
Please merge this pull request in first: https://github.com/jaiarobotics/jaiabot-debian/pull/54

### DIFF
--- a/config/gen/bot.py
+++ b/config/gen/bot.py
@@ -55,15 +55,10 @@ elif jaia_arduino_type == 'usb':
 else:
     jaia_arduino_dev_location="/dev/ttyAMA1"
 
+jaia_data_offload_ignore_type="NONE"
+
 if "jaia_data_offload_ignore_type" in os.environ:
     jaia_data_offload_ignore_type=os.environ['jaia_data_offload_ignore_type']
-
-if jaia_data_offload_ignore_type == "goby":
-    jaia_data_offload_ignore_type="GOBY"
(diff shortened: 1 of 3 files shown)


config/gen/bot.py
```diff
@@ -55,6 +55,16 @@
 else:
     jaia_arduino_dev_location="/dev/ttyAMA1"
```
michael-jaia, line 57: Do we need the duplicate checks if we just pass in GOBY, TASKPACKET, or NONE

michael-jaia (commented):
Same as last PR, not sure if it's necessary, but thought I'd leave the comment just in case. Also why do we need to adjust the both gen/bot.py and gen/systemd.py?

ferrom-jaia:
> Same as last PR, not sure if it's necessary, but thought I'd leave the comment just in case. Also why do we need to adjust the both gen/bot.py and gen/systemd.py?

We need systemd.py so our services get the correct configurations and it is also responsible for making the configurations environmental variables. We need the change in bot.py to update the mission manager configuration with the correct value from the environmental variable.

michael-jaia (approved):
Comments resolved, looks good, approving PR into 1.y

PR #792, opened 2024-02-21
bug/omp-spacing-input-render-and-validation

michael-jaia:
## Summary
* Increases the speed of input updates and allows 0 to displayed to make it easier for operators to enter values that do not start with the minimum value of 1.

## Details
* Passes a state setter through props and creates a copy of the missionParams to trigger a new state update when the input changes
* Adds a final input validation for the mission spacing values right before the preview is created to check for 0s. If a value equals 0, it is set to 1.

## Testing
* Quickly typed values into the mission spacing input boxes
* Typed in characters to make sure they default to 0
* Adjusted the point and line spacing values and watched them grow and shrink in the preview as expected
* Set the point and line spacing to 0 and saw no console errors and watched the preview use the default value of 1
* Played a mission when both the point and line spacing were set to 0

### DIFF
--- a/src/web/command_control/client/components/CommandControl.tsx
+++ b/src/web/command_control/client/components/CommandControl.tsx
@@ -3022,9 +3022,21 @@ export default class CommandControl extends React.Component {
 	//
 	//
 	/**
+	 * Triggers a state update for changes made to the MissionParams
+	 * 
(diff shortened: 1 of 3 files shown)


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #794, opened 2024-02-21
bug/rc-dive

ferrom-jaia:
## Summary
Removing interval for RC when the dive button is clicked

## Details
In RC mode if the operator was driving a bot and then proceeded to send the RC dive command the bot would get interrupted 2 times a second while comms were still good. To resolve this we removed the interval to send the RC commands when the dive command is clicked.

## Testing
1. To test start a simulation with 1 bot running at 1 warp
2. Open liaison: http://localhost:30002/?_=/scope
3. Watch engineering command messages and low control messages
4. Open JCC
5. Split the screen to watch both JCC and liaison
6. Put bot in RC mode and send a rudder command
7. You will see engineering messages being sent periodically
8. Then click on the dive button in the RC dive mode
9. These engineering messages should stop coming in


michael-jaia (approved):
Code looks good and I followed the testing instructions. I watched the engineering command come through in Manual Dual and then when I sent an RC dive command, the engineering command timestamp remained fixed until sent another command by using Manual Single to drive forward.

PR #796, opened 2024-02-21
Task/fix apt cacher

tsaubergine:
Fixes issues arising from https://github.com/jaiarobotics/jaiabot/pull/783:

- Ansible 2.9 (Ubuntu 20.04) does not properly understand "ansible.builtin.shell" (but "shell" works). Revert to using "shell"
- Add support for pulling offline / online with hub cache update from hub ID that is not hub0 (defaults to hub1 in Liaison drop-down menu).

image


ferrom-jaia (approved):
Looks good! Will test on fleet 1 02-22-2024.

PR #802, opened 2024-02-26
Task/data pull

tsaubergine:
## Switch bot to hub data offload to be a pull from the hub (rather than a push from the bots).

https://jaia-innovation.atlassian.net/browse/JAIAB-892?atlOrigin=eyJpIjoiNGYxN2U3MzI0ZDlkNDYwNGE4M2ExOGU4NzE2MGM5N2MiLCJwIjoiaiJ9

### Background

To support easier fleet configuration and management, **we want to switch to having a single Yubikey control runtime SSH access** from the hub to the bots, and remove the need for the bots to ever SSH into the hub. This means we can go from provisioning a large number of SSH keys per fleet to just one per hub.

### Implementation

See sequence diagram at https://github.com/jaiarobotics/jaiabot/blob/task/data-pull/src/doc/markdown/page60_mission_management.md#log-offload

In order to support this hub-only SSH scenario, we needed to switch the `rsync` command for data offload to be run on the hub (pull) rather the bots (push). This pull request implements this.

The **end result of the data offload process is largely identical** to before (files get put in /var/log/jaiabot/bot_offload on the hub and /var/log/jaiabot/archive on the bot, and old files get removed). The process is a bit different, where the **old "jaiabot-dataoffload.sh" script functionality is now split into three simpler scripts**:

- "jaia**bot**-predataoffload.sh" - Run on the bot (stage the desired files - typically to /var/log/jaiabot/staging).
- "jaia**hub**-dataoffload.sh" - Run on the hub (actually does the offload - typically to /var/log/jaiabot/bot_offload).
- "jaia**bot**-postdataoffload.sh" - Run on the bot (archive - typically to /var/log/jaiabot/archive - and remove old files).

While doing this, I also implemented some improvements to the data offload process by stopping logging entirely and closing the .goby log during data offload (and opening a new log when leaving data offload). **This removes the necessity to move/restore the various symlinks.**

State machine changes:
- Removed PostDeployment:DataProcessing as this placeholder state never ended up doing anything.
- Added new PostDeployment:Failed for cases when the data offload fails for some reason. From here EvRetryDataOffload can be posted (or EvShutdown can be sent).
- `goby_logger` stops logging when entering PostDeployment:DataOffload *and* closes the log. Thus, we can move around the log files without concern about whether one of them is being written to.
- While entering PostDeployment:DataOffload, the "jaiabot-predataoffload.sh" script is run to move the desired files to be offloaded to a "staging" directory (typically /var/log/jaiabot/staging).
- Once the hub_manager notes that a bot has entered PostDeployment:DataOffload, it queues up an offload for that bot, which is done in series. Each offload is run from the hub using the "jaiahub-dataoffload.sh" script (which runs rsync to move the files from "staging" to the offload directory on the hub, typically /var/log/jaiabot/bot_offload).
- When the hub_manager offload completes (or fails), it sends the new command type DATAOFFLOAD_COMPLETE (or _FAILED) to the bot, which then enters PostDeployment:Idle (or PostDeployment:Failed).
- As it exits PostDeplyment:DataProcessing, the bot runs the "jaiabot-postdataoffload.sh" script to move the files from staging in the archive directory (typically /var/log/jaiabot/archive) and removes any files in archive or log_dir that are older than a week (as before). `goby_logger` is also commanded to START_LOGGING (which creates a new .goby log).

Message changes:
- Removed `data_offload_percentage` from BotStatus
- Added `BotOffloadData` (including `data_offload_percentage`  and success/failure) to HubStatus
- Add DATA_OFFLOAD_COMPLETE / DATA_OFFLOAD_FAILED CommandTypes to Command message.

Minor additional changes:
- For commands from the hub, check in `jaiabot_hub_manager` to ensure the timestamp will be unique upon receipt at the bot so unique commands are not rejected as duplicate based on the timestamp. If the timestamp will not be unique, increment the timestamp by the required amount to make it unique (currently one second). This is important when sending commands automatically from hub_manager as the RECOVERED command can be sent very shortly before DATA_OFFLOAD_COMPLETE if there is little data to offload.
- I added support to the docker-arm64-build-and-deploy.sh script to allow it to deploy to a virtualbox fleet.
- Add support for different distros (focal/jammy), repos (e.g. continuous, release, beta) and versions (1.y, etc) to the scripts docker-arm64-build-and-deploy.sh and docker-build-build-system.sh. Defaults to focal/release/1.y if the applicable env vars are not set.
- Cleaned up CMakeLists.txt for install of various scripts.
- Added libjaiabot_moos_gateway_plugin.so.1 to the list of goby_logger libraries so these extensions get included in the .goby log files (I had to rename jaiabot.protobuf.jaiabot_config -> jaiabot.protobuf.jaiabot_gateway_config to avoid conflicting with the same name used by Liaison).

### Requirements

Requires Goby 3.1.4 (in packages.jaia.tech *staging* which points to *test* and *continuous* repos only). This version adds the ability for `goby_logger` to close an old log when STOP_LOGGING is sent. This allows us to stop logging *and* have the log file closed (and thus rotated when we start logging) during DATA_OFFLOAD.

### To do
- **Update JCC to use HubStatus values to track download progress wheels**
- Probably also want to have a way to send RETRY_DATA_OFFLOAD from JCC if bots are in PostDeployment:Failed.

### Testing performed

Used VirtualFleet with 2 bots, 1 hub, deployed both with Debian package and using docker-arm64-build-and-deploy.sh

```
export jaiabot_repo=continuous
export jaiabot_version=1.y
export jaiabot_machine_type=virtualbox
./docker-build-build-system.sh

jaiabot_systemd_type=bot ./docker-arm64-build-and-deploy.sh bot1-virtualfleet6 bot2-virtualfleet6
jaiabot_systemd_type=hub ./docker-arm64-build-and-deploy.sh hub1-virtualfleet6

```

Tested that:
- Data are offloaded into correct bot_offload directory in hub.
- Data are moved into correct archive directory on bots.
- Data are deleted if older than 7 days (fudged with `touch`)
- HubStatus correctly reports data offload percentage.
- Bots transition to PostDeployment:Idle on success
- Bots transition to PostDeployment:Failed on failure.
- The whole process is reproducible across several missions.
- .goby files can be converted into reasonable looking .txt files (`for i in *.goby; do goby log convert $i; done`)


tsaubergine:
@ferrom-jaia Merged in 1.y and resolved conflicts.

tsaubergine:
Merge conflicts resolved and retested using 2 bot / 1 hub VirtualFleet (VirtualBox).

Fixed installing jcc.conf using docker-arm64-build-and-deploy.sh

ferrom-jaia (approved):
Looks great! Tested on fleet 0 with hub 1 and bot 1.

PR #806, opened 2024-02-28
feature/no-forward-progress-state

tsaubergine:
## Overview

Implements: https://jaia-innovation.atlassian.net/browse/JAIAB-884

Logic (in `jaiabot_mission_manager`) is pretty straightforward:
- If the vehicle is pitch up and it should be moving (desired speed > threshold and in IVP control mode) we wait for a period of time and if this condition still persists, enter the new Pause::ResolveNoForwardProgress state.
- In Pause::ResolveNoFowardProgress, it just STOPS the motor and waits a bit and resumes.

Added to `jaiabot_simulator`:
- Sets a fixed pitch up (set to 85 degrees - please update to a more realistic value if you have one) when vehicle is not underway 
- Add StopForwardProgress option to SimulatorCommand which cuts simulated thrust for the commanded duration.

## Testing

Ran simulated mission with dive, drift, and stationkeep and did not see any false triggers into Pause::ResolveNoForwardProgress.

Manually injected StopForwardProgress when vehicle is moving to ensure it correctly detects the failure and enters  Pause::ResolveNoForwardProgress, waits, and resumes.

Command for manually injecting 30 second forward progress failure:

```
export GOBY_TOOL_LOAD_SHARED_LIBRARY=/home/toby/opensource/jaiabot/build/amd64/lib/libjaiabot_messages.so

goby zeromq publish jaiabot::simulator_command jaiabot.protobuf.SimulatorCommand 'stop_forward_progress { duration: 30 }' --interprocess 'platform: "bot1_fleet0"'
```

(also tested with duration: 120 to ensure the EvNoForwardProgress triggers multiple times).

Needs testing in the field, as the simulator pitch data (and root cause of the motor issue) will not accurately capture the problem.


ferrom-jaia (approved):
Looks great! Tested in sim following the directions on the PR. Also tested on the bench.

PR #815, opened 2024-03-14
task/rename-wpt-toggle

edsanville:
Used a script I had lying around :)


marshalllele (dismissed):
Everything looks good. No errors popped up while running through each toggle in JCC with the console open.

marshalllele (approved):
My mistake, everything looks good!

PR #818, opened 2024-03-14
bug/take-control-button-behind-rc-panel

marshalllele:
Brings the 'Take Control' panel to the very front of JCC


src/web/command_control/client/style/CommandControl.less
```diff
@@ -1319,6 +1319,7 @@ div.closeButton {
   margin: 8pt;
   padding: 8pt;
   border-radius: 4pt;
+  z-index: 2500; /** 2500 places 'Take Control' on top of any other panel so it's always visible */
```
michael-jaia: Let's move the comment above the z-index line and use the /* Text here */

config/gen/bot.py
```diff
@@ -60,6 +60,11 @@
 if "jaia_data_offload_ignore_type" in os.environ:
```
michael-jaia, line 60: Hey Nick can you start this branch from 1.y so it does not mark the code from the branch you built from as new. Also can you make the B in the pr name lowercase

michael-jaia (changes requested):

PR #819, opened 2024-03-15
bug/take-control-button-behind-rc-panel

marshalllele:
Places the 'Take Control' panel in front of all other panels.


ferrom-jaia (approved):
Looks good! Works as expected.

PR #823, opened 2024-03-27
task/goby-jaiabot-package-path

JRoybot:
Update path to goby and jaibot packages in build scripts

### DIFF
--- a/scripts/setup-tools-build-nodocker.sh
+++ b/scripts/setup-tools-build-nodocker.sh
@@ -1,12 +1,5 @@
 #!/usr/bin/env bash
 
-# sudo may not be present, if in docker
-apt-get -y update && apt-get -y install sudo
-
-echo -e "deb http://packages.jaia.tech/ubuntu/continuous/1.y/ focal/\ndeb http://packages.jaia.tech/ubuntu/gobysoft/continuous/1.y/ focal/" >> /etc/apt/sources.list.d/jaiabot_release_1.y.list && \
-  apt-key adv --recv-key --keyserver hkp://keyserver.ubuntu.com:80 954A004CD5D8CF32 && \
-  apt-key adv --recv-key --keyserver hkp://keyserver.ubuntu.com:80 19478082E2F8D3FE
-
 # Add packages.gobysoft.org mirror to your apt sources
 echo "deb http://packages.jaia.tech/ubuntu/gobysoft/continuous/1.y/ `lsb_release -c -s`/" | sudo tee /etc/apt/sources.list.d/gobysoft_release.list
 # Install the public key for packages.gobysoft.org
--- a/scripts/setup-tools-build-nodocker.sh
+++ b/scripts/setup-tools-build-nodocker.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # Add packages.gobysoft.org mirror to your apt sources
-echo "deb http://packages.jaia.tech/ubuntu/gobysoft/continuous/1.y/ `lsb_release -c -s`/" | sudo tee /etc/apt/sources.list.d/gobysoft_release.list
+echo "deb http://packages.jaia.tech/ubuntu/gobysoft/continuous/1.y/ `lsb_release -c -s`/" | sudo tee /etc/apt/sources.list.d/gobysoft_continuous.list
 # Install the public key for packages.gobysoft.org
 sudo apt-key adv --recv-key --keyserver hkp://keyserver.ubuntu.com:80 19478082E2F8D3FE
 # Update apt
--- a/scripts/setup-tools-build-nodocker.sh
+++ b/scripts/setup-tools-build-nodocker.sh
@@ -3,12 +3,12 @@
 # sudo may not be present, if in docker
 apt-get -y update && apt-get -y install sudo
 
-echo -e "deb http://packages.jaia.tech/ubuntu/release/1.y/ focal/\ndeb http://packages.jaia.tech/ubuntu/gobysoft/release/1.y/ focal/" >> /etc/apt/sources.list.d/jaiabot_release_1.y.list && \
+echo -e "deb http://packages.jaia.tech/ubuntu/continuous/1.y/ focal/\ndeb http://packages.jaia.tech/ubuntu/gobysoft/continuous/1.y/ focal/" >> /etc/apt/sources.list.d/jaiabot_release_1.y.list && \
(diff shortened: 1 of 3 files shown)


ferrom-jaia (approved):

tsaubergine:
Why are we switching from release to continuous here? A bit more description of the need and problem this solves would be great.

ferrom-jaia:
> Why are we switching from release to continuous here? A bit more description of the need and problem this solves would be great.

The issue we experienced is gobysoft release packages not compatible continuous 1.y jaiabot software. 

Jaia tool was merged in but it needed the newest gobysoft packages. 

Following the process to point at the right gobysoft packages we had a mismatch with what our local machines were using.

The mismatch also led to errors when deploying code.

Build error for goby.ev optional field. 

```bin/tool/config.proto:38:25: Error while parsing option value for "field": Unknown enumeration value of "DEVELOPER" for field "action".
bin/tool/config.proto:85:24: Error while parsing option value for "field": Message type "goby.GobyFieldOptions.ConfigurationOptions" has no field named "position".
bin/tool/config.proto:42:19: Option "(goby.ev)" unknown. Ensure that your proto definition file imports the proto which defines the option.
bin/tool/config.proto:52:21: Option "(goby.ev)" unknown. Ensure that your proto definition file imports the proto which defines the option.
bin/tool/config.proto:57:18: Option "(goby.ev)" unknown. Ensure that your proto definition file imports the proto which defines the option.
bin/tool/config.proto:68:22: Option "(goby.ev)" unknown. Ensure that your proto definition file imports the proto which defines the option.
bin/tool/config.proto:71:18: Option "(goby.ev)" unknown. Ensure that your proto definition file imports the proto which defines the option.
bin/tool/config.proto:31:12: Option "(goby.msg).cfg" unknown. Ensure that your proto definition file imports the proto which defines the option.
gmake[2]: *** [src/CMakeFiles/jaiabot_config.dir/build.make:193: include/bin/tool/config.pb.cc] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:734: src/CMakeFiles/jaiabot_config.dir/all] Error 2
gmake: *** [Makefile:136: all] Error 2```

**Comment by tsaubergine** [comment]
Ok, well this PR as written will break the CircleCI deb builds for beta and release as it will build against the continuous repo. 

The relevant lines are here: https://github.com/jaiarobotics/jaiabot/blob/1.y/.circleci/config.yml#L68-L72, which adds "newer" repos as needed. This only works though if the one in the Docker container version is older than any added (release < beta < continuous < test).

So if you switch the Dockerfile to default to continuous, this will always take precedence for packages installed, even on beta/release builds.

**Comment by tsaubergine** [comment]
I think the Dockerfile needs to take `repo` as an argument (as it takes `distro` already), defaulting to "release", and the local build scripts can set it to "continuous" when they build the Dockerfile locally.

And while we're at it, might as well set `version` as well (1.y, 2.y, etc.) to set ourselves up for forthcoming 2.y

**Comment by ferrom-jaia** [comment]
> Ok, well this PR as written will break the CircleCI deb builds for beta and release as it will build against the continuous repo.

Interesting. I wonder why CircleCi was not breaking then using the release goby packages to build against the continuous repos.

Nice Catch!

**Review by ferrom-jaia — CHANGES_REQUESTED** [review]
Based on feedback from Toby

**Review by tsaubergine — CHANGES_REQUESTED** [review]
repo (continuous, release, beta, test) and version (1.y, 2.y) need to be arguments to the Dockerfile. 

Most of this work has been already been done in https://github.com/jaiarobotics/jaiabot/pull/802

See from that PR: [amd64/Dockerfile](https://github.com/jaiarobotics/jaiabot/pull/802/files#diff-11649f50ff4a2ffffe8f32f3f4fcce7eca0e92d570190d051e9c1913cadd5455), [arm64/Dockerfile](https://github.com/jaiarobotics/jaiabot/pull/802/files#diff-30f3e74d39ac204ba1d58cb234bccd9f07f4c3bf76e5b365a367c6b16a0b69ec),[scripts/docker-build-build-system.sh](https://github.com/jaiarobotics/jaiabot/pull/802/files#diff-e68771609d1478f51d2de9cf8c44d966e7f0a9351af3e0843fe43171856ece57),
[scripts/docker-arm64-build-and-deploy.sh](https://github.com/jaiarobotics/jaiabot/pull/802/files#diff-7495e4096889159525e54fbae4c801bf80ed83e51bf9c089de738d819a60368f)

Please either review and merge that PR or cherry-pick the required build system changes into this PR.

The only part that PR doesn't fix I think is scripts/setup-tools-build-nodocker.sh (which CircleCI doesn't use, so I don't much care what it contains).

File: scripts/setup-tools-build-nodocker.sh [hunk 1]
```diff
@@ -1,14 +1,7 @@
...
-# sudo may not be present, if in docker
-apt-get -y update && apt-get -y install sudo
-
-echo -e "deb http://packages.jaia.tech/ubuntu/release/1.y/ focal/\ndeb http://packages.jaia.tech/ubuntu/gobysoft/release/1.y/ focal/" >> /etc/apt/sources.list.d/jaiabot_release_1.y.list && \
-  apt-key adv --recv-key --keyserver hkp://keyserver.ubuntu.com:80 954A004CD5D8CF32 && \
-  apt-key adv --recv-key --keyserver hkp://keyserver.ubuntu.com:80 19478082E2F8D3FE
-
 # Add packages.gobysoft.org mirror to your apt sources
-echo "deb http://packages.jaia.tech/ubuntu/gobysoft/1.y/ `lsb_release -c -s`/" | sudo tee /etc/apt/sources.list.d/gobysoft_release.list
+echo "deb http://packages.jaia.tech/ubuntu/gobysoft/continuous/1.y/ `lsb_release -c -s`/" | sudo tee /etc/apt/sources.list.d/gobysoft_release.list
```
tsaubergine: Please also change file name to "gobysoft_continuous.list" if changing the repo to avoid confusion.

ferrom-jaia:
Let's remove the docker configuration from this pull request and just have the fix for the local development environments. Toby has a pull request that addresses the docker issues. This way developers can continue to make progress.

scripts/setup-tools-build-nodocker.sh
ferrom-jaia: Updated

tsaubergine (approved):

ferrom-jaia (approved):
After consulting with Toby, we can approve and merge this in.

PR #826, opened 2024-03-29
bug/spurious-depth

edsanville:
Tested in simulation, and no spurious depth readings came through to the map for a drift task or a station keep task.


ferrom-jaia (approved):
Looks good! Tested in my environment.

PR #831, opened 2024-04-06
task/separate-imu-drivers-from-wave-calculations

edsanville:
Use the pyjaia.waves module from both the adafruit BNO055 and BNO085 directories.


ferrom-jaia (approved):
Looks good. Tested on a bot with a bno085 imu and a bot that has a bno055 imu.

PR #832, opened 2024-04-12
bug/nproc-in-build-script

edsanville:
Add some documentation to README.md about the environment variables that control the build process.


tsaubergine:
I like it!

tsaubergine (dismissed):

tsaubergine:
I wonder if it makes more sense to use MemAvailable rather than MemTotal? (Sometimes I have some VMs running that uses a lot of memory)

edsanville:
> I wonder if it makes more sense to use MemAvailable rather than MemTotal? (Sometimes I have some VMs running that uses a lot of memory)

Good point!

ferrom-jaia (approved):

PR #834, opened 2024-04-16
task/button-icon-OMP-on-left-side-toolbar

Moriarty777:
- The OMP and Settings icons have been changed.
- All the icons are now imported from Mdi library.
- Removed unused import and code cleanup.


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #836, opened 2024-04-16
bug/missing-dependency-dccl4-apps

ferrom-jaia:
### Description
Added dccl apps to setup-tools-build-nodocker.sh. Needed to run dccl and get the hash of the different dccl messages we send.

### Testing
- Run setup-tools-build.sh
- Run the ./build.sh
- Verify the package installed: `dpkg-query -W <package name>`


marshalllele (approved):
Looks good! Tested using the given instructions and the package was installed.

PR #839, opened 2024-04-17
bug/jdv-not-loading-logs

ferrom-jaia:
### Description
The group id for bot status is now the api version number. Using 0 as a default will not work. This pull request gets bot id from file name. Also updated to use goby log convert instead of goby_log_tool.

### Testing
- Start simulation with 2 bots
- Copy goby files to /var/log/jaiabot/bot_offload directory
- Start jdv
- Select logs to convert
- Select data to view


jason-jaia (approved):
I tested locally and it worked great.

Thank you.

PR #840, opened 2024-04-18
Fix weirdness with jaia_mode settings

tsaubergine:
Updated preseed.goby to more explicitly set jaia_mode based on expected settings. This fixes the bug where running "./all.launch" puts the logs in the runtime directory (rather than the simulation directory), as the all.launch run of preseed.goby was setting the runtime log directory, which is then not overwritten by the child bot.launch/hub.launch preseed.goby.

- If you're running a script in simulation directory, you expect jaia_mode == simulation. This is now explicitly set in simulation/preseed.goby.
- If you call config/preseed.goby directly, you must now set jaia_mode == simulation or jaia_mode == runtime. The only other place this is used (as far as I know) is in gen/systemd.py, where jaia_mode is explicitly set prior to sourcing preseed.goby
- Add default for jaia_bot_type to HYDRO if unset.


ferrom-jaia (approved):

PR #842, opened 2024-04-20
bug/after-performing-a-task-mission-manager-breaks

ferrom-jaia:
### Description
Need to pass a default publisher to send task packet over intervehicle. The group is inferred from jaiabot::INTERVEHICLE_API_VERSION and possibly the already-set bot_id field we don't need to set anything using the group function. However Goby requires one to exist for using non-broadcast groups, so we define a no-op group_func
### Testing
- Start simulation
- Send a run to a bot with a task
- The bot should still be in good health after the run is completed


jason-jaia (approved):
Tested locally on two different machines. Works nicely.

edsanville:
Works for me.

PR #844, opened 2024-04-22
task/hide-echo-field-if-not-bots-are-echos

michael-jaia:
### Summary
Hides the acoustic modem toggle in the JCC if none of the bots in the pod are JaiaBot PAMs

### Description
Loops through the bot types as the bot status messages arrive on the client side. If one of the bot status messages includes the bot type ECHO (in the future we will change this to PAM), then a new state property `enableEcho` is set to true. This is passed down to the panels that render the toggle.

### Testing
Tested in sim with:
2 HYDROs
2 ECHOs
1 HYDRO, 1 ECHO
1 ECHO, 1 HYDRO
For each combination of bots, I created individual runs and verified I could use the toggle on dive and drift tasks. I also created a survey mission to verify the toggle appeared correctly in that panel.


ferrom-jaia (approved):
Looks good. I saw Twomey testing this.

PR #845, opened 2024-04-25
feature/show-bot-path

edsanville:


src/web/jcc/client/components/CommandControl.tsx
```diff
@@ -801,6 +807,70 @@ export default class CommandControl extends React.Component {
 		}
 	}
 
+	
+	/**
+	 * Use JaiaAPI to poll the botPath data.
+	 */
+	pollBotPaths() {
+		console.log(this.lastBotPathPointUtime)
```
marshalllele: This console.log()  gunks up the console a good bit

marshalllele (changes requested):

src/web/jcc/client/components/CommandControl.tsx
edsanville: Removed!

src/web/jcc/client/components/CommandControl.tsx
```diff
@@ -74,6 +74,7 @@ import '../style/CommandControl.less'
 // Utility
 import cloneDeep from 'lodash.clonedeep'
 import { HelpWindow } from './HelpWindow'
+import { getBotPathColor } from './shared/BotPathColors'
```
marshalllele: It looks like this is an unused import

marshalllele (changes requested):

marshalllele (approved):
Looks good! Tested both paths in sim and in JDV.

PR #847, opened 2024-04-26
feature/confirmation-for-sending-mission-out-of-range

marshalllele:
If a user tries to run a mission containing a run that will finish outside of comms range of the hub (250 meters), a dialog box will pop up asking for confirmation before the run is sent. 

A warning will also show up if the user attempts to send a mission while the hub has no location data.

Testing:
Out of Comms Range:
- Create a mission with run(s) ending outside of comms range
- Ensure that a confirmation message appears for each run ending out of range with a bot assigned to it

No Hub Location Data:
- Change line 1255 of CommandControl  (`if (hub.location.lat === undefined || hub.location.lon === undefined)` in `runMissions()`) to `if (hublat === undefined || hublon === undefined)`
- Insert the following code just above this line
`let hublat = undefined`
`let hublon = undefined`

This should force the warning to come up.


src/web/jcc/client/components/CommandControl.tsx
```diff
@@ -74,6 +74,7 @@ import '../style/CommandControl.less'
 // Utility
 import cloneDeep from 'lodash.clonedeep'
 import { HelpWindow } from './HelpWindow'
+import { skip } from 'node:test'
```
ferrom-jaia: Are we using this?

src/web/jcc/client/components/CommandControl.tsx
```diff
@@ -1240,6 +1241,7 @@ export default class CommandControl extends React.Component {
 			const commDest = this.determineAllCommandBots(true, false, false, false)
 			const botIdsAssignedToRuns: number[] = []
 			const runs = mission.runs
+			const commsRange = 250
```
ferrom-jaia: Can we add a const at the top of the file for the magic number?

src/web/jcc/client/components/CommandControl.tsx
```diff
@@ -1272,29 +1274,79 @@ export default class CommandControl extends React.Component {
 				CustomAlert.alert(commDest.poorHealthMessage + commDest.idleStateMessage + commDest.downloadQueueMessage + commDest.disconnectedMessage + notAssignedMessage)
 			} 
 			else {
-
+				const hub = Object.values(this.state.podStatus.hubs)[0]
```
ferrom-jaia: Can we create a function for this: getHubStatus. Add some checks to insure there is a hub.

src/web/jcc/client/components/CommandControl.tsx
```diff
@@ -1272,29 +1274,79 @@ export default class CommandControl extends React.Component {
...
-								if (runs[key].id === this.getRunList().runIdInEditMode) {
-									const runList = this.getRunList()
-									runList.runIdInEditMode = ''
-									this.setRunList(runList)
+							const run = runs[key]
+							const lastWptIdx = run.command.plan.goal.length - 1
+							const lastWptLoc = run.command.plan.goal[lastWptIdx].location
+				
+							// If distance is greater than 250 meters, verify that user understands the last waypoint is out of comms range before sending the run to the bot
+							let distance = OlGetDistance([hub.location.lat, hub.location.lon], [lastWptLoc.lat, lastWptLoc.lon])
```
ferrom-jaia: Do we need a check here to see if the hub has a lat lon?

src/web/jcc/client/components/Details.tsx
```diff
@@ -274,27 +275,52 @@ function issueCommandForHub(api: JaiaAPI, hub_id: number, commandForHub: Command
     })
 }
 
-function issueRunCommand(api: JaiaAPI, bot: PortalBotStatus, botRun: Command, setRcMode: (botId: number, rcMode: boolean) => void, disableMessage: string) {
+function issueRunCommand(api: JaiaAPI, bot: PortalBotStatus, hub: PortalHubStatus, botRun: Command, setRcMode: (botId: number, rcMode: boolean) => void, disableMessage: string) {
     takeControlFunction(() => {
+        const commsRange = 250
```
ferrom-jaia: Can we add a const at the top of the file for the magic number?

src/web/jcc/client/components/Details.tsx
```diff
@@ -274,27 +275,52 @@ function issueCommandForHub(api: JaiaAPI, hub_id: number, commandForHub: Command
...
+            warning(warningString)
+        }
 
-                api.postCommand(botRun).then(response => {
-                    if (response.message) {
-                        error(response.message)
+        CustomAlert.confirmAsync("Are you sure you'd like to play this run for Bot: " + bot.bot_id + '?', 'Play Run').then(async (confirmed) => {
+            if (confirmed) {
+                // If distance is greater than 250 meters, verify that user understands the last waypoint is out of comms range before sending the run to the bot
+                let distance = OlGetDistance([hub.location.lat, hub.location.lon], [lastWptLoc.lat, lastWptLoc.lon])
```
ferrom-jaia: Do we need a check here to see if the hub has a lat lon?

ferrom-jaia (changes requested):

src/web/jcc/client/components/Details.tsx
```diff
@@ -277,34 +278,53 @@ function issueCommandForHub(api: JaiaAPI, hub_id: number, commandForHub: Command
...
+            hubLat = hub.location.lat
+            hubHasPosData = true
         }
 
         CustomAlert.confirmAsync("Are you sure you'd like to play this run for Bot: " + bot.bot_id + '?', 'Play Run').then(async (confirmed) => {
             if (confirmed) {
                 // If distance is greater than 250 meters, verify that user understands the last waypoint is out of comms range before sending the run to the bot
-                let distance = OlGetDistance([hub.location.lat, hub.location.lon], [lastWptLoc.lat, lastWptLoc.lon]) 
-                if (distance > commsRange) {
+                let distance = OlGetDistance([hubLon, hubLat], [lastWptLoc.lat, lastWptLoc.lon])
```
ferrom-jaia: Can we add the check here for lat lon:
```
let userConfirmed = null
if (hub.location.lat === undefined || hub.location.lon === undefined) {
   userConfirmed = await new Promise((resolve) => { 
    // Wait for user to confirm that they know the hub has no location data
    CustomAlert.confirm("WARNING: The hub does not currently have position data. Make sure the last waypoint is not out of comms range from the hub.", "Confirm",
        () => resolve(true),
        () => resolve(false)
     ) 
  })
} else {
   let distance = OlGetDistance([hub.location.lat, hub.location.lon], [lastWptLoc.lat, lastWptLoc.lon]) 
   if (distance > commsRange) {
       userConfirmed = await new Promise((resolve) => { 
         // Wait for user to confirm that they know the run will finish out of comms range
         CustomAlert.confirm(`WARNING: The last waypoint is out of comms range from the hub. Are you sure you want to send bot ${bot.bot_id} on its run?`, "Confirm",
              () => resolve(true),
              () => resolve(false)
         ) 
       })
   }
}
if (userConfirmed === true) {
....
```

src/web/jcc/client/components/Details.tsx
```diff
@@ -277,34 +278,53 @@ function issueCommandForHub(api: JaiaAPI, hub_id: number, commandForHub: Command
...
 
         // Exit if we have a disableMessage
         if (disableMessage !== "") {
             CustomAlert.alert(disableMessage)
             return
         }
-
+        
+        // Checks for hub lat/lon. If none are found, they are set to MAX_SAFE_INTEGER so distance calculation doesn't break
         if (hub.location.lat === undefined || hub.location.lon === undefined) {
```
ferrom-jaia: Can we remove this logic block. Example in following comment.

src/web/jcc/client/components/CommandControl.tsx
```diff
@@ -1324,6 +1326,21 @@ export default class CommandControl extends React.Component {
...
+	/**
+	 * Gets the array of all hubs in the pod
+	 * 
+	 * @param {void}
+	 * @returns {PortalHubStatus[]} Array of all hubs
+	 */
+	getHubStatus() {
+		while (!Object.values(this.state.podStatus.hubs)) {
+			warning("No hub found")
+		}
```
JRoybot, line 1338: Is this a potential infinite loop?

src/web/jcc/client/components/CommandControl.tsx
```diff
@@ -1367,31 +1383,100 @@ export default class CommandControl extends React.Component {
...
-									const runList = this.getRunList()
-									runList.runIdInEditMode = ''
-									this.setRunList(runList)
+							const run = runs[key]
+							const lastWptIdx = run.command.plan.goal.length - 1
+							const lastWptLoc = run.command.plan.goal[lastWptIdx].location
+				
+							// If distance is greater than 250 meters, verify that user understands the last waypoint is out of comms range before sending the run to the bot
+							let distance = OlGetDistance([hubLat, hubLon], [lastWptLoc.lat, lastWptLoc.lon]) 
+							if (distance > COMMS_RANGE && botIndex !== -1 && commDest.botIds.includes(botIndex)) {
```
JRoybot, line 1423: This is a very complex if-then-else structure.  Suggest adding plain english comments at the key if and else statements to help reader follow logic.  Cyclomatic complexity of this block is extremely high. (measure of all possible ways through the code)  

Hard to break down existing logic but since the if on 1423 is nearly identical to the one on 1456 other than the COMMS_Range check AND the logic at the end of both conditional blocks look the same (1444-1450 & 1459-1466) it is likely the code would be easier to follow if check for 

botIndex !== -1 && commDest.botIds.includes(botIndex)

as a first check and then inside that conditional block check if distance > COMMS_RANGE and do the special logic for the user confirmation

Then finish the outer block with the common logic cited above.  Anything to make this simpler would be good.

JRoybot (commented):
Generally looks good.  A few comments but no issues with code.  Testing went well after rebuilding with a clean build dir.

edsanville:
> Note the issues I had with the web build and the inconsistent edit mode are not due to this branch. I did not see the recent note to delete current build directory and recreate. Once I did that the web build errors went away and I have not been able to recreate the edit mode issue. Sorry.

Yep, that's definitely because of the leftover softlink that linked `src/web/jed` to `build/web_dev/jed`.

We moved from soft-linking JED, to building it with webpack.  Since the old softlink was still there, it was essentially trying to build back into the source directory, which triggered another build, _ad infinitum_.

The issue should be gone now that the `build` directory was cleared.  We could add a check to make sure the destination directory isn't a softlink.  That would prevent anyone from experiencing this issue again, but it would add a bit of weird complexity to the `run.sh` script.

michael-jaia:
Closing PR due to new edge cases found that would hurt the user experience such as when the hub does not have GPS every run would prompt a second confirmation to play.

src/web/jcc/client/components/Details.tsx
edsanville: Please pardon my side note:
You can use the CustomAlert.confirmAsync method like this:
```
userConfirmed = await CustomAlert.confirmAsync("WARNING: The hub does not currently have position data. Make sure the last waypoint is not out of comms range from the hub.", "Confirm")
```

edsanville: Come to think of it... we should probably only have the confirmAsync method, and use await if we want it to be synchronous.  I didn't think of that when I wrote that class.

PR #849, opened 2024-04-29
bug/bot-details-should-not-close-when-toggling-edit

michael-jaia:
## Testing Cases

- Toggle edit mode in the details panel of a bot
- Toggle edit mode with the goal settings panel open
- Toggle edit mode from the mission panel


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #852, opened 2024-05-02
task/change-check-for-old-times-to-repeat-times

michael-jaia:
## Summary
Switch checks from ignoring old commands and task packets to ignoring commands and task packets with the same timestamps.

## Testing
Function `handle_task_packet`

1. Set start time for all task packets to be the same to simulate repeat task packets
    1. Get the current UNIX timestamp from: https://www.unixtimestamp.com/
    2. Hard code the task packet start time to the UNIX timestamp on line 320 of `machine.cpp`
    3. Plan a single waypoint run with a dive and repeat it twice
        1. Should see glog output `Repeat taskpacket received! Ignoring...` for the logging of the second task packet
    4. **PASS**
2. Do not interfere with start time for task packets (restore the changes to `machine.cpp`)
    1. Add this line: `glog.is_debug1() && glog << "Publishing taskpacket..." << std::endl;` to 420 of `hub_manager` 
    2. Plan a single waypoint run with a dive and repeat it twice
        1. Should see glog output `Publishing taskpacket...` twice
    3. **PASS**

Function `handle_command`

1. Set the time for all commands to be the same to simulate repeat commands
    1. Get the current UNIX timestamp from: https://www.unixtimestamp.com/
    2. Hard code the return value of the `now` function in `jaia_portal.py` to the UNIX timestamp
    3. Activate the bots
    4. Run a mission
    5. Should see glog output `Repeat command received! Ignoring...`
    6. **PASS**
2. Do not interfere with start time for commands
    1. Add the line `glog.is_debug1() && glog << "Processing command..." << std::endl;` to line 717 of `handle_command` in mission_manager/app.cpp
    2. Should see the glog output `Processing command...` two times when two commands are sent
    3. **PASS**


ferrom-jaia (approved):
Looks good. I followed the test plan.

PR #854, opened 2024-05-03
task/reinstate-auto-assign-feature-with-omp

michael-jaia:
## Testing
1. Create more runs than bots in OMP, apply the mission
2. Create the same amount of runs as bots in OMP, apply mission
3. Create less runs than bots in OMP, apply mission


ferrom-jaia (approved):
Looks good! Tested in sim.

JRoybot (approved):

PR #855, opened 2024-05-06
bug/120-second-jed-dive

edsanville:
…m 100 seconds.


edsanville:
The DCCL message had max value of 100 for the timeout value.  This was increased to 120.  If we want to support longer timeouts, we can increase this even further.

edsanville:
We may want to consider whether it's actually worth changing the DCCL message.  The other fix would be to change the GUI to only allow up to 100 second dives, to match the DCCL max value.

PR #858, opened 2024-05-10
bug/imu-detection-reboot

ferrom-jaia:
## Description
The imu detection entered a loop where it would keep restarting. This left the vehicle unable to drive using the autonomy. Moved the imu publish to restart into the imu drivers instead of fusion. Updated the imu restart state to 15 seconds.

## Testing
* Deployed code to bot 1 fleet 0
* Stopped the python driver for the imu
* Check for the error message in liaison
* Check to see if the error goes away and imu data comes back


michael-jaia (approved):
Watched Matt test on flat bot + statically reviewed code...approving PR

PR #860, opened 2024-05-10
feature/station-keep-for-a-certain-time

marshalllele:
Allows the user to choose a time for station keep tasks to be executed for before moving onto the next task. 

Testing:
* Ensure that when a time is given, bots actually only station keep for that given time
* Ensure that bots move onto the next task after station keep timer is up
* If the task is on the last wpt, the state should update from TASK_STATION_KEEP to RECOVERY_STATION_KEEP after the time is up


src/web/jcc/client/components/TaskSettingsPanel.tsx
```diff
@@ -85,6 +85,20 @@ function TaskOptionsPanel(props: Props) {
         props.onChange(newTask)
     }
 
+    function onChangeStationKeepParameter(evt: React.ChangeEvent<HTMLInputElement>) {
```
michael-jaia, line 94: Hey Nick, can you add a function comment here and switch the `var` to `let`. I know all the other related functions use `var` but lets stop that trend.

michael-jaia (changes requested):
Hey Nick, I saw in your description that you do not want the last wpt to allow a station keep time to be set. I noticed a single wpt run allows a time to be added for station keep. Is that the behavior you want or should a single wpt run no allow an operator to set a time?

ferrom-jaia:
> Hey Nick, I saw in your description that you do not want the last wpt to allow a station keep time to be set. I noticed a single wpt run allows a time to be added for station keep. Is that the behavior you want or should a single wpt run no allow an operator to set a time?

It is okay if the user sets the last wpt as a station keep task with a duration. The vehicle will go into recovery state and will station keep at the last wpt.

michael-jaia (dismissed):
Statically reviewed code + tested in simulator with single waypoint mission, survey mission, and a classic run. When testing on the classic run (4 wpts), I added multiple station keep tasks to make sure the time set on one waypoint did not overwrite the time saved on another.

ferrom-jaia (approved):
Looks good! Tested in sim.

PR #864, opened 2024-05-14
feature/update-hub-location-sim

edsanville:
Do we want the hub location(s) to persist?  That's another task, I think.


src/web/jcc/client/components/SetHubLocationPanel.tsx
```diff
@@ -0,0 +1,188 @@
...
+        return destroySelectOnMapInteraction
+    }, [])
+
+
+    return (
+        <div className="panel" key="set-hub-location">
+            <div className='panel-heading' style={{color: "black"}}>Set Hub Location</div>
+            <div className='mission-settings-panel-container'>
+
+                { /* this.hubIdSelectionElement() */ }
```
marshalllele: Is this meant to be commented out? If so, it doesn't look like we need the hubIdSelectionElement() function

src/web/jcc/client/components/SetHubLocationPanel.tsx
```diff
@@ -0,0 +1,188 @@
...
+    const selectOnMapInteractionRef = useRef(null)
+
+    const hubLocation = props.hubs[hub_id].location
+
+    /**
+     * A Select element for choosing the hub_id.
+     *
+     * @returns {*} The Select element.
+     */
+    function hubIdSelectionElement() {
```
marshalllele: It looks like this function isn't used

src/web/jcc/client/components/SetHubLocationPanel.tsx
```diff
@@ -0,0 +1,188 @@
...
+/* eslint-disable react/sort-comp */
+/* eslint-disable no-unused-vars */
+import React, { useState, useEffect, useRef } from 'react'
+import Button from '@mui/material/Button';
+import {JaiaAPI} from '../../common/JaiaAPI'
+import { PortalHubStatus } from './shared/PortalStatus';
+import { CommandForHub, GeographicCoordinate, HubCommandType } from './shared/JAIAProtobuf';
+import { getElementById, getGeographicCoordinate } from './shared/Utilities';
+import { Map } from 'ol';
+import { Interaction } from 'ol/interaction';
```
marshalllele: Unused import

src/web/jcc/client/components/SetHubLocationPanel.tsx
```diff
@@ -0,0 +1,188 @@
...
+    map: Map
+}
+
+
+export default function SetHubLocationPanel(props: Props) {
+    const [ hub_id, set_hub_id ] = useState(Number(Object.keys(props.hubs)[0]) ?? 1)
+    const [ selectingOnMap, setSelectingOnMap ] = useState(false)
+    const selectOnMapInteractionRef = useRef(null)
+
+    const hubLocation = props.hubs[hub_id].location
```
marshalllele: Unused variable

marshalllele (changes requested):
Functionality works well in sim! Deployed to hub and verified that the Simulation accordion only shows up if in a simulator.

src/web/jcc/client/components/SetHubLocationPanel.tsx
edsanville: It is commented out, because as @michael-jaia noted, the sim doesn't allow multi-hub yet.  I kept the function around in case we want it someday.

src/web/jcc/client/components/SetHubLocationPanel.tsx
edsanville: (see above)

marshalllele (dismissed):
Everything looks good! Testing in sim and deployed to a hub to make sure the Simulation panel only appears in a simulation.

src/lib/messages/jaia_dccl.proto
```diff
@@ -101,10 +101,15 @@ message CommandForHub
...
         SHUTDOWN_COMPUTER = 42;
+
+        // simulator commands
+        SET_HUB_LOCATION = 80;
     }
 
     required HubCommandType type = 10;
     optional uint32 scan_for_bot_id = 11;
+
+    optional GeographicCoordinate set_hub_location = 80;
```
michael-jaia: Could `set_hub_location` be confusing the way protocol buffers generates setters. Would `hub_location` work here?

src/web/jcc/client/components/SetHubLocationPanel.tsx
```diff
@@ -0,0 +1,189 @@
+/* eslint-disable jsx-a11y/label-has-for */
```
michael-jaia: Can these eslint lines be removed?

src/web/jcc/client/components/SetHubLocationPanel.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+interface Props {
+    hubs: {[key: number]: PortalHubStatus}
+    api: JaiaAPI
+    map: Map
+}
+
+
+export default function SetHubLocationPanel(props: Props) {
+    const [ hub_id, set_hub_id ] = useState(Number(Object.keys(props.hubs)[0]) ?? 1)
```
michael-jaia: Can we make this `hubID` or `hubId` and `setHubID` or `setHubId` to keep up with the JS style

src/web/jcc/client/components/SetHubLocationPanel.tsx
```diff
@@ -0,0 +1,189 @@
+/* eslint-disable jsx-a11y/label-has-for */
+/* eslint-disable jsx-a11y/label-has-associated-control */
+/* eslint-disable react/sort-comp */
+/* eslint-disable no-unused-vars */
+import React, { useState, useEffect, useRef } from 'react'
```
michael-jaia: Can we group these imports based on React, Jaia, OpenLayers, and style?

src/web/jcc/client/components/SetHubLocationPanel.tsx
```diff
@@ -0,0 +1,189 @@
...
+    const selectOnMapInteractionRef = useRef(null)
+
+    const hubLocation = props.hubs[hub_id].location
+
+    /**
+     * A Select element for choosing the hub_id.
+     *
+     * @returns {*} The Select element.
+     */
+    function hubIdSelectionElement() {
```
michael-jaia: Can we update the return type to be an array of HTML elements? Also maybe we should add a @notes section to the comment to say would be used in a simulator with multiple hubs

src/web/jcc/client/components/SetHubLocationPanel.tsx
```diff
@@ -0,0 +1,189 @@
...
+    
+    /**
+     * Gets the hub location from the text inputs.
+     *
+     * @returns {GeographicCoordinate}
+     */
+    function getInputHubLocation(): GeographicCoordinate {
+        const lat = Number(getElementById<HTMLInputElement>('set-hub-location-latitude').value)
+        const lon = Number(getElementById<HTMLInputElement>('set-hub-location-longitude').value)
+        if (lat == null || lon == null) {
```
michael-jaia: Can we make these `===` since we know getElementById returns null if the element is not found

src/web/jcc/client/components/SetHubLocationPanel.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+        return [ labelRow, hubIdSelectionRow ]
+    }
+
+
+    
+    /**
+     * Gets the hub location from the text inputs.
+     *
+     * @returns {GeographicCoordinate}
```
michael-jaia: We should make this reflect the fact the function can return null

src/web/jcc/client/components/SetHubLocationPanel.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+
+    /**
+     * Calls the API to submit a location change for a certain hub.
+     *
+     */
+    function submitHubLocation() {
+        const hubLocation = getInputHubLocation()
+
+        if (hubLocation == null) {
```
michael-jaia: Let's make this `===`

src/web/jcc/client/components/SetHubLocationPanel.tsx
```diff
@@ -0,0 +1,189 @@
...
+        return {
+            lat: lat,
+            lon: lon
+        }
+    }
+
+
+    /**
+     * Calls the API to submit a location change for a certain hub.
+     *
```
michael-jaia: We need the `@returns` line here

src/web/jcc/client/components/SetHubLocationPanel.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+        props.api.postCommandForHub(hubCommand)
+    }
+
+    
+    /**
+     * Initiate a PointerInteraction to select a new hub location with a click or tap.
+     */
+    const toggleSelectOnMapInteraction = () => {
+        if (selectOnMapInteractionRef.current != null) {
```
michael-jaia: Can we make this `!==`?

src/web/jcc/client/components/SetHubLocationPanel.tsx
```diff
@@ -0,0 +1,189 @@
...
+            set_hub_location: hubLocation
+        }
+
+        props.api.postCommandForHub(hubCommand)
+    }
+
+    
+    /**
+     * Initiate a PointerInteraction to select a new hub location with a click or tap.
+     */
```
michael-jaia: Can we update this function comment to include the return value

src/web/jcc/client/components/SetHubLocationPanel.tsx
```diff
@@ -0,0 +1,189 @@
...
+    const toggleSelectOnMapInteraction = () => {
+        if (selectOnMapInteractionRef.current != null) {
+            destroySelectOnMapInteraction()
+            return
+        }
+
+        selectOnMapInteractionRef.current = new PointerInteraction({
+            handleEvent: (evt) => {
```
michael-jaia: How do the boolean return values impact the `handleEvent` property?

src/web/jcc/client/components/SetHubLocationPanel.tsx
```diff
@@ -0,0 +1,189 @@
...
+        props.map.addInteraction(selectOnMapInteractionRef.current)
+        props.map.getTargetElement().style.cursor = 'crosshair'
+    }
+    
+
+    /**
+     * Destroy the hub location selection interaction.
+     */
+    const destroySelectOnMapInteraction = () => {
+        if (selectOnMapInteractionRef.current != null) {
```
michael-jaia: Can we make this `!==`

src/web/jcc/client/components/SettingsPanel.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { ReactElement, useState } from 'react'
```
michael-jaia: Ed, what do you think of wrapping the SetHubLocationPanel with the mui accoridon components, renaming that panel to HubLocationAccordion and then all we need to do is import that component in this file as one unit and call it as `<HubLocationAccordion props={...} />`

src/web/jcc/client/components/SetHubLocationPanel.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+
+    // Destroy the map select interaction on unmount, if present.
+    useEffect(() => {
+        return destroySelectOnMapInteraction
+    }, [])
+
+
+    return (
+        <div className="panel" key="set-hub-location">
```
michael-jaia: Can we make the contents centered within the accordion details?

michael-jaia (changes requested):

src/lib/messages/jaia_dccl.proto
edsanville: Good point on this one, done.

src/web/jcc/client/components/SetHubLocationPanel.tsx
edsanville: I decided to go with snake_case, only because hub_id is a field name from a protobuf message.  Our general principle in JS has been to use the snake_case only for protobuf stuff, so those fields have unambiguous names across JS, C++, and python.

@ferrom-jaia what do you think on this one?

src/web/jcc/client/components/SetHubLocationPanel.tsx
edsanville: This one actually has no return value.

src/web/jcc/client/components/SetHubLocationPanel.tsx
edsanville: From the OpenLayers docs:

```Method called by the map to notify the interaction that a browser event was dispatched to the map. The function may return false to prevent the propagation of the event to other interactions in the map's interactions chain.```

So, I'm preventing the click event from falling through to the other interactions on the map's interaction stack.  I added some comments to clarify the reasoning on this.

File: src/web/jcc/client/components/SettingsPanel.tsx (hunk 18 above)
Comment (edsanville) [reply src/web/jcc/client/components/SettingsPanel.tsx]: We could do that, but then we'd need to pass `isOpenAccordionTab` and `handleAccordionTabClick` as props to the HubLocationAccordion component.  I'm not sure it's worth it, but the Accordion JSX probably does need some refactoring...  it looks like there's a lot of code duplication there.  We probably need some brainstorming to come up with a way to make it more concise and manageable.

File: src/web/jcc/client/components/SetHubLocationPanel.tsx (hunk 15 above)
Comment (michael-jaia) [reply src/web/jcc/client/components/SetHubLocationPanel.tsx]: Yes I meant to say we should include the `@returns {void}` tag here

File: src/web/jcc/client/components/SettingsPanel.tsx (hunk 18 above)
Comment (michael-jaia) [reply src/web/jcc/client/components/SettingsPanel.tsx]: Ok, I am onboard with keeping the Accordion wrapper in the settings panel. One thing I did notice when reviewing the props: what do you think of importing the instance of JaiaAPI in the file and not passing it as props?

File: src/web/jcc/client/components/SetHubLocationPanel.tsx [hunk 20]
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+
+    // Destroy the map select interaction on unmount, if present.
+    useEffect(() => {
+        return destroySelectOnMapInteraction
+    }, [])
+
+
+    return (
```
michael-jaia: Hey Ed, can we move the inline css into its own css file. A SetHubLocation.css or .less file in style/components. A less file would probably be better so we can use the nesting to avoid naming collisions with other files.

src/web/jcc/client/components/SettingsPanel.tsx
edsanville: I'm onboard with that idea.  In general, importing the api and the map, rather than passing as props would probably make sense app-wide, really.

src/web/jcc/client/components/SetHubLocationPanel.tsx
edsanville: Ok, gotcha!

src/web/jcc/client/components/SetHubLocationPanel.tsx
edsanville: Yeah, we can certainly do that.  I did it in an ad-hoc way here, only because I had to finagle it to get things to center properly.  It's kind of a code smell for our css, that we need to massage the styles so much to get them to look good.  I was trying to localize those tweaks, so they didn't affect any other elements... but yes they're probably better off in a less file for that panel.

src/web/jcc/client/components/SetHubLocationPanel.tsx
edsanville: Ok, switched to hubId

src/web/jcc/client/components/SetHubLocationPanel.tsx
```diff
@@ -0,0 +1,196 @@
...
+        if (selectOnMapInteractionRef.current !== null) {
+            destroySelectOnMapInteraction()
+            return
+        }
+
+        selectOnMapInteractionRef.current = new PointerInteraction({
+            handleEvent: (evt) => {
+                if (evt.type == "click") {
+                    const clickedLocation = getGeographicCoordinate(evt.coordinate, evt.map)
+                    getElementById<HTMLInputElement>('set-hub-location-latitude').value = clickedLocation.lat.toFixed(6)
```
ferrom-jaia: Can we use a React way to do this? This works, but it looks different than most of the jcc code with how we interact with text fields.

src/web/jcc/client/components/SetHubLocationPanel.tsx
```diff
@@ -0,0 +1,196 @@
...
+import '../style/components/SetHubLocationPanel.less'
+
+
+interface Props {
+    hubs: {[key: number]: PortalHubStatus}
+    api: JaiaAPI
+    map: Map
+}
+
+export default function SetHubLocationPanel(props: Props) {
```
ferrom-jaia: Instead of SetHubLocationPanel can we use SetHubLocation for both the function name and the css name since it is not a panel

src/web/jcc/client/style/components/SetHubLocationPanel.less
```diff
@@ -0,0 +1,21 @@
+#set-hub-location-panel {
```
ferrom-jaia: If we are changing the name of the function this should change as well

ferrom-jaia (commented):
Nice work Ed, I left a couple of comments that we should review and discuss before merging in.

ferrom-jaia:
Can we update the pull request image or remove it since it does not match the implementation.

src/web/jcc/client/components/SetHubLocationPanel.tsx
edsanville: Ah yes...  will do this.

edsanville: Ok, I used React Refs to track the two fields as uncontrolled elements.  We could use React state controlled elements, but that would be more complex code...

ferrom-jaia (approved):
Looks great!

PR #867, opened 2024-05-14
task/refactor-hub-details

michael-jaia:
### Summary
First pass at integrating React Context into the web applications. The HubDetails component now relies on context for the data it needs to correctly render the panel. The markdown file included in this pull request outlines the implementation.

### Testing
* Launched the simulator
* Opened the HubDetails panel
* Clicked through all of the accordions
* Closed the panel using the X button
* Deselected the hub by clicking the hub tab in the toolbar
* Clicked rapidly through all of the accoridons
* Clicked rapidly on the hub tab in the toolbar
* Opened and closed BotDetails panels with HubDetails open and closed


src/doc/markdown/page111_developing_with_react_context.md
```diff
@@ -0,0 +1,162 @@
...
+* ContextProviderProps
+    * Satisfies TypeScript's requirements for accepting a React Component as a prop
+```
+// Example
+interface GlobalContextProviderProps {
+    children: ReactNode
+}
+```
+
+#### 3. Constant Variables and Enums
```
ferrom-jaia, line 92: Do we want to add coming soon for this section?

src/doc/markdown/page111_developing_with_react_context.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+#### 5. Reducer Function
+A reducer function is called by a dispatch function that comes from "using" the dispatch context. The only required property of the `action` object is `type` which is a string describing the type of action to dispatch. In some cases, you will want to pass data to the reducer function to set state to your desired value. If that is the case, modify the `Action` interface, so you can pass that data in with the `action` object.   
+```
+function globalReducer(state: ContextType, action: Action) {
+    let mutableState = {...state}
+    switch (action.type) {
+        case 'SAVED_CLIENT_ID':
+            return saveClientID(mutableState, action.clientID)
```
ferrom-jaia, line 109: Do we need to ensure action.clientID has data or is available to access?

ferrom-jaia: Should we have a rule or a uniform way of passing data? 

For example:

SAVE CLIENT ID type

Action has a client object that includes a clientID
action.client.id, action.client.etc

ferrom-jaia: Also, how do we know if a action type requires an input. Is this documented or does the developer just review the reducer?

src/doc/markdown/page111_developing_with_react_context.md
```diff
@@ -0,0 +1,162 @@
...
+    import { GlobalDispatchContext } from '../context/GlobalContext'
+
+    const globalDispatch = useContext(GlobalDispatchContext)
+
+    function handleClosePanel() {
+        globalDispatch({ type: 'CLOSED_HUB_DETAILS' })
+    }
+```
+
+#### 6. Reducer Helper Functions
```
ferrom-jaia, line 133: Love it!

src/doc/markdown/page111_developing_with_react_context.md
```diff
@@ -0,0 +1,162 @@
...
+    return mutableState
+}
+
+function handleControlTaken(mutableState: GlobalContextType) {
+    mutableState.controllingClientID = mutableState.clientID
+    return mutableState
+}
+```
+
+#### 7. Context Provider Component
```
ferrom-jaia, line 147: Love it!

src/web/context/GlobalContext.tsx
```diff
@@ -0,0 +1,231 @@
...
+ * 
+ * @param {GlobalContextType} state Holds the most recent reference to state 
+ * @param {Action} action Contains data associated with a state update 
+ * @returns {GlobalContextType} A copy of the updated state
+ */
+function globalReducer(state: GlobalContextType, action: Action) {
+    let mutableState = {...state}
+    switch (action.type) {
+        case 'SAVED_CLIENT_ID':
+            return handleSavedClientID(mutableState, action.clientID)
```
ferrom-jaia, line 73: Should we check if clientID exists

src/web/context/GlobalContext.tsx
```diff
@@ -0,0 +1,231 @@
...
+            return handleClosedHubDetails(mutableState)
+        
+        case 'CLICKED_HUB_TAB':
+            return handleClickedHubTab(mutableState)
+
+        case 'CLICKED_BOT_TAB':
+            return handleClickedBotTab(mutableState)
+
+        case 'CLICKED_HUB_ACCORDION':
+            return handleClickedHubAccordion(mutableState, action.hubAccordionName)
```
ferrom-jaia, line 97: Should we check if hubAccordionName exists

src/web/context/GlobalContext.tsx
```diff
@@ -0,0 +1,231 @@
...
+}
+
+/**
+ * Handles the interplay between selecting the hub and clicking the HubTab
+ * 
+ * @param {GlobalContextType} mutableState State object ref for making modifications
+ * @returns {GlobalContextType} Updated mutable state object
+ */
+function handleClickedHubTab(mutableState: GlobalContextType) {
+    const isHubSelected = mutableState.selectedPodElement !== null && mutableState.selectedPodElement.type === PodElement.HUB
```
ferrom-jaia, line 160: This could be condensed with this using ternary operator:

const isHubSelected = mutableState.selectedPodElement !== null && mutableState.selectedPodElement.type === PodElement.HUB;
mutableState.showHubDetails = !isHubSelected;

src/web/context/GlobalContext.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+    if (isHubSelected) {
+        mutableState.showHubDetails = false
+    } else {
+        mutableState.showHubDetails = true
+    }
+
+    if (mutableState.showHubDetails) {
+        mutableState.selectedPodElement = { type: PodElement.HUB, id: 1 }
```
ferrom-jaia: Should id always be 1?

src/web/context/GlobalContext.tsx
ferrom-jaia: Readability in your solution is probably better.

src/web/context/GlobalContext.tsx
```diff
@@ -0,0 +1,231 @@
...
+ * 
+ * @param {GlobalContextType} mutableState State object ref for making modifications
+ * @returns {GlobalContextType} Updated mutable state object
+ * 
+ * @notes
+ * This function only unselects the hub and closes the HubDetails panel. 
+ * It does not handle bot selection yet.
+ */
+function handleClickedBotTab(mutableState: GlobalContextType) {
+    if (mutableState.selectedPodElement !== null && mutableState.selectedPodElement.type === PodElement.HUB) {
```
ferrom-jaia: Can we be consistent with this conditional in terms of setup. See above function.

src/web/context/HubContext.tsx
```diff
@@ -0,0 +1,96 @@
...
+ * 
+ * @param {HubContextType} state Holds the most recent reference to state 
+ * @param {Action} action Contains data associated with a state update 
+ * @returns {HubContextType} A copy of the updated state
+ */
+function hubReducer(state: HubContextType, action: Action) {
+    let mutableState = {...state}
+    switch (action.type) {
+        case 'HUB_STATUS_POLLED':
+            return handleHubStatusPolled(mutableState, action.hubStatus)
```
ferrom-jaia, line 41: Should we check if hub status exists?

src/web/jcc/client/components/BotListPanel.tsx
```diff
@@ -35,21 +37,37 @@ export function BotListPanel(props: Props) {
...
         let tracked = bot.bot_id == props.trackedBotId ? 'tracked' : ''
         let disconnected = Math.max(0.0, bot.portalStatusAge / 1e6) > 30 ? 'disconnected' : ''
 
+        /**
+         * Triggers the handling logic in CommandControl and dispatches the event to GlobalContext
+         * 
+         * @returns {void}
+         * 
+         * @notes
+         * This is an instance where we are begininning to migrate from CommandControl state to context
```
ferrom-jaia: typo begininning

src/web/jcc/client/components/BotListPanel.tsx
```diff
@@ -58,21 +76,41 @@ export function BotListPanel(props: Props) {
...
+                selected = 'selected'
+        }
+
+        /**
+         * Triggers the handling logic in CommandControl and dispatches the event to GlobalContext
+         * 
+         * @returns {void}
+         * 
+         * @notes
+         * This is an instance where we are begininning to migrate from CommandControl state to context
```
ferrom-jaia: typo begininning

src/web/jcc/common/JaiaAPI.ts
```diff
@@ -90,10 +90,24 @@ export class JaiaAPI {
...
+  getClientId() { return this.clientId }
+
   getMetadata() { return this.get('jaia/metadata') }
 
   getStatus() { return this.get('jaia/status') }
 
+  /**
+   * Gets most recent hub status
+   * 
+   * @returns {{[key: string]: PortalHubStatus}} Object containing most recent hub status
```
ferrom-jaia: hub statuses

ferrom-jaia (changes requested):
Great work Twomey! This is going to be so helpful moving forward. I added some comments for review.

src/doc/markdown/page111_developing_with_react_context.md
michael-jaia: I didn't add a description because I didn't think it would add much more than what the title already says, but I can add a short blurb if you would like for consistency with the other sections

src/doc/markdown/page111_developing_with_react_context.md
michael-jaia: I can add some more documentation around this. Essentially if you need a piece of data to update state the way you want, you would append an optional property to the action interface and save the data to that property. Only the case (in the switch statement) for the action.type you call will be looking for that property which is why I did not use a question mark because that case is expecting that data.

src/doc/markdown/page111_developing_with_react_context.md
```diff
@@ -0,0 +1,162 @@
...
+    * `hubAccordionStates`
+#### `PodContext`
+* Encapsulates `HubContext` and `BotContext`
+* Provides a location to store pod-wide data
+##### `HubContext`
+* Stores the `PortalHubStatus`
+##### `BotContext`
+* Coming soon
+
+## General Context Strcuture
```
edsanville: Typo 'Structure'

src/doc/markdown/page111_developing_with_react_context.md
edsanville: This is an interesting point.  We'll probably have to document which fields are required for each action type.  Preferably, IMO, in a way where it'll show up by hovering in an IDE like VSCode.

src/web/context/GlobalContext.tsx
```diff
@@ -0,0 +1,252 @@
...
+}
+
+interface HubAccordionStates {
+    quickLook: boolean,
+    commands: boolean,
+    links: boolean
+}
+
+export interface GlobalAction {
+    type: string,
```
edsanville, line 29: What if type is an enum, so we can document each action type:

```
export enum GlobalActionType {
    
    /**
     * @param clientID The client id that was saved.
     */
    SAVED_CLIENT_ID = 1,
    TAKE_CONTROL_SUCCESS = 2
    ...
}

export interface GlobalAction {
    type: GlobalActionType,
    clientID?: string,
    hubAccordionName?: string
}
```

edsanville: If we used an enum like the above code, then `@param` shows up when hovering over the type enum:

image

src/web/components/HealthStatusLine.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+    return (
+        <tr>
+            <td>Health</td>
+            <td>
+                <div className={healthClassName}>{props.healthState}</div>
+            </td>
+        </tr>
+    )
+}
```
edsanville, line 26: I love the idea of keeping reusable components in this components directory, and containers in the containers directory.

src/web/context/GlobalContext.tsx
edsanville: It makes me wonder whether the action/reducer pattern is solving the same problem as the delegate pattern solves.  Except that with a delegate object, the actions would be methods, and you'd be able to document, enforce, and type-check all of the parameters for each method.  There may be some difference between the two patterns that I'm not seeing, though...

edsanville (dismissed):
Tested in simulator, and everything works as expected.

src/web/context/GlobalContext.tsx
edsanville: The enum wouldn't have to be integers either, it could be strings.

src/web/context/GlobalContext.tsx
michael-jaia: I think there is some overlap. I think with React we need to stick with the action/reducer pattern since under the hood the reducer updates state with the value it returns. I suppose we could make the action object expect functions, but it might add an extra challenge to maintainability and onboarding since we deviate from the uses shown in the React docs.

src/doc/markdown/page111_developing_with_react_context.md
ferrom-jaia: No that is fine

ferrom-jaia (approved):
Thank you for reviewing comments! Looks great! Tested on my machine.

PR #868, opened 2024-05-15
Fix intermittent crash of goby_liaison_standalone on boot

tsaubergine:
https://jaia-innovation.atlassian.net/browse/JAIA-1192

Wt is throwing an exception when trying to start with bind address 0::0 (IPv6 localhost).

0::0 _should_ always be available. This must be a bug in how Wt is calling Asio domain name query.

It appears that if we wait until a non-loopback interface is configured with any IPv6 address (just a fe80:: link-local address is sufficient) that this issue goes away. I chose wlan0 since this is configured in all real and virtual bots. This PR updates the systemd service to wait up to 60 seconds for an IPv6 address to be assigned to wlan0. This means that in systems without wlan0 (should they exist), this service will eventually start (after a minute).

Tested by replicated in a VirtualBox hub and then tested 5-6 reboots after fixing and the issue does not recur.


ferrom-jaia (changes requested):
Can we add this for the jaiabot_goby_liaison service as well?

tsaubergine:
> Can we add this for the jaiabot_goby_liaison service as well?

I think it's unnecessary as even under the most optimistic scenario the jaiabot_goby_liaison won't start until after IPv6 has been provisioned.

That said I updated the PR to apply to both Liaison services.

ferrom-jaia (approved):
Looks great! Tested on fleet 1 hub and bot 5.

PR #869, opened 2024-05-15
feature/custom-overlay-script-improvements

ted-at-rtr:
Document palette bands. Improve geotiff_scan/format: recognize palette bands and print warning if image does not contain extent/projection.

Matt: This is the exact script update I sent you a few days ago.   I realized that I did not change the documentation to talk about palette bands, so I fixed that too.


ferrom-jaia (approved):
Looks great! Saw updated script output. Verified the images showing in JCC.

PR #874, opened 2024-05-21
feature/jcc-retry-offload

michael-jaia:
## Notes
* Data offload did not work in sim because of permission/key issues with rsync trying to access my local machine as if was a remote system. The solution I added excludes using the localhost IP address.
* Added a couple of question marks to prevent breaks to the system when the mission state comes across as undefined.
* We know the download queue needs some improvements. This is not the ticket to tackle those.
## Testing
* Tested with 3 bots where the data offload proceeded as expected.
* Tested with 3 bots where all of them failed which showed that the retry data offload button was enabled, and I could see the command being sent in via the browser console.


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #881, opened 2024-05-24
Bug/fix cloud regressions and add `jaia ping` tool action

tsaubergine:
Issues I found when setting up the Cloud demo:

- Minor issue with Ansible playbooks on Cloudhub (which uses a new Ansible to support the EC2 instances tasks) - `warn: false` has been removed from  Ansible. I removed it and this will now work but will show a minor warning when running these tasks on the command line.
- Have JDV listen on 0::0 so it works on IPv6 networks.

I found `jaia ssh` needed a companion `jaia ping` so I created it as well.


tsaubergine:
Updated with changes from https://github.com/jaiarobotics/jaiabot/pull/885

ferrom-jaia (approved):
Looks good! Tested on flat fleet

PR #882, opened 2024-05-24
Minor renaming of src/bin directories

tsaubergine:
For consistency, renamed three app directories that are differently named:

`src/bin/jaiabot_xyz` -> `src/bin/xyz`


ferrom-jaia (approved):
Looks good!

PR #885, opened 2024-05-29
bug/goby-liaison-fails-inet6-timeout

tsaubergine:
The workaround from https://github.com/jaiarobotics/jaiabot/pull/868 fails if the bots aren't connected to wifi within 60 seconds. 

Rather than add complexity, this PR reverts the Liaison bind address to 0.0.0.0 (IPv4) for all real fleets and 0::0 (IPv6) for Virtual fleets.

We will need to resolve this issue fully if/when we move the bots to IPv6, but that can part of that task, hopefully with a newer version of Wt.

https://jaia-innovation.atlassian.net/browse/JAIA-1263?atlOrigin=eyJpIjoiOTcxOTgzYjExMjc3NDNmN2EzZjNiNzg0MjExYjQ4ZWMiLCJwIjoiaiJ9

Tested on VirtualBox fleet (should be close to actual bots / hubs), and AWS Cloud VirtualFleet. Needs verification on actual bots but as this reverts the bind address for real bots/hubs to that prior to this regression (0.0.0.0), I do not expect issues.


ferrom-jaia (approved):
Looks good! Tested on flat fleet.

PR #886, opened 2024-05-31
bug/data-offload-percent-missing-from-details-panel

michael-jaia:
## Summary
I forgot to add update the data offload percentage in the details panel. It now grabs the offload percent from the hub status message.

## Testing
Watched the percentage come through correctly in the details panel while working with Fleet 2 (Bots 1 and 2)


ferrom-jaia (approved):
Looks great! Twomey showed demo.

PR #890, opened 2024-06-05
Task/Support x.y

tsaubergine:
This PR supports the creation of 2.y (and 3.y in the future).

- Copy jaiabot-debian as jaiabot/debian (we should archive jaiabot-debian after merging this PR).
- Copy jaiabot-rootfs-gen as jaiabot/rootfs (we should archive jaiabot-rootfs-gen after merging this PR).
  - Merge CircleCI config into jaiabot/.circleci/config.yml. Set filter template for rootfs builds to default to 1.y only (not every commit as these take a while to build). Can be set temporarily in a feature branch to do rootfs builds on that branch just as we can set test Debian package builds now.
- Rename test/X.y to test/1.y everywhere (with the expectation of the new 2.y will use test/2.y).
- Update packages.jaia.tech to have 2.y repos and configuration.
- Update docs at src/doc/markdown/page19_repository.md
- Add scripts/release_branch file that all scripts now use to determine the current release branch (1.y, 2.y, etc.) rather than defaulting to 1.y.
- Very minor changes to support Ubuntu Noble.


ferrom-jaia (approved):
Looks good!

PR #893, opened 2024-06-06
bug/install-web-dependencies-clears-pre-commit-hooks

michael-jaia:
## Summary
* Bug: Installing the dependencies in the run script, cleared the pre-commit hook script.
* Solution: Copy the pre-commit hook script for husky execution after the web dependencies are installed in `run.sh`

## Testing
* Executed the run script with the pre-commit script cleared and watched it copy after the web dependencies were installed


ferrom-jaia (approved):
Looks good!

PR #895, opened 2024-06-07
bug/race-condition-fleet2-radio-jaia-497

ferrom-jaia:
## Description
The xbee 250 radios require a bypass command to be entered before getting into command mode. Sometimes the radio would not send the correct menu back. This caused the xbee driver to hang. This implementation will retry until the xbee driver sends the correct menu or it times out. The timeout is for the 900 xbee radio which does not require a bypass menu.

## Testing
1. Deployed to fleet 0 bot 1 and hub 1
2. Ensure the radios are able to talk


src/lib/comms/xbee/xbee.cpp
```diff
@@ -439,23 +440,128 @@ size_t jaiabot::comms::XBeeDevice::bytes_available()
 
 void jaiabot::comms::XBeeDevice::enter_command_mode()
 {
-    // We need to send bypass command if we are
-    // are using the new radio.
-    // (Need a delay of 5 seconds between the bypass commands)
-    // Old radio still functions the same.
+    // Start trying to enter command mode
```
michael-jaia: I think the function name is descriptive enough where we do not need the comment

src/lib/comms/xbee/xbee.cpp
```diff
@@ -439,23 +440,128 @@ size_t jaiabot::comms::XBeeDevice::bytes_available()
...
+ * @brief Keep trying to enter command mode until the expected result
+ * 
+ */
+void jaiabot::comms::XBeeDevice::try_enter_command_mode()
+{
+    std::string buffer;
+    std::string deliminiter = "";
+    int timeout_seconds = 5;
+    
+    //Triggers menu to appear for bypass
```
michael-jaia: Nit: Add space between // and Triggers

src/lib/comms/xbee/xbee.h
```diff
@@ -117,9 +119,16 @@ class XBeeDevice
     std::string read_until(const std::string& delimiter);
     size_t bytes_available();
     void read(void* ptr, const size_t n_bytes);
+    void async_read_with_timeout(std::string& buffer, const std::string& delimiter,
+                                 int timeout_seconds,
+                                 std::function<void(const std::string&)> handler);
+    
+    //convert string to hex
```
michael-jaia: Nit: Match with other comment format

src/lib/comms/xbee/xbee.cpp
```diff
@@ -439,23 +440,128 @@ size_t jaiabot::comms::XBeeDevice::bytes_available()
...
-    write("+++");
+    // Set the expiration time of the timer
+    timer->expires_from_now(boost::posix_time::seconds(timeout_seconds));
 
-    sleep(1);
-    // Read until we get the OK, in case some binary data comes through and interferes
-    read_until("OK\r");
+    // Set up the timer's asynchronous wait operation
+    timer->async_wait([&](const boost::system::error_code& ec) {
+        if (!ec)
```
michael-jaia, line 524: Is it possible that asyc_wait produces an error code. If so would the try_enter_command_mode handle indefinitely?

src/lib/comms/xbee/xbee.cpp
```diff
@@ -439,23 +440,128 @@ size_t jaiabot::comms::XBeeDevice::bytes_available()
...
+                handler("read_error");
+            }
+        });
+    
+    // Wait for either the read operation to complete or the timeout to occur
+    io->run();
+}
+
+/**
+ * @brief Function to convert the string to hexadecimal format and return as a string
```
michael-jaia: Nit: "return as a string "may not be necessary here. It may be more concise to say "...to a hexadecimal formatted string" if you want to include the word "string" in the brief.

src/lib/comms/xbee/xbee.h
```diff
@@ -117,9 +119,16 @@ class XBeeDevice
...
+    std::string convertToHex(const std::string& str);
 
     // Command mode stuff
     void enter_command_mode();
+    void try_enter_command_mode();
     void assert_ok();
     void exit_command_mode();
```
michael-jaia: General: What is the format the docstring comments should follow. Some have return statements some do not.

src/lib/comms/xbee/xbee.cpp
```diff
@@ -439,23 +440,128 @@ size_t jaiabot::comms::XBeeDevice::bytes_available()
 
 void jaiabot::comms::XBeeDevice::enter_command_mode()
```
michael-jaia, line 440: Do we need `enter_command_mode` to wrap `try_enter_command_mode`? Could we either call `try_enter_command_mode` directly or rename `try_enter_command_mode` to `enter_command_mode`

michael-jaia (changes requested):

src/lib/comms/xbee/xbee.h
ferrom-jaia: Hmm, do you think we should be consistent and add return void?

src/lib/comms/xbee/xbee.cpp
michael-jaia: No, if async_wait produces an error code, it will be handled in the else statement of lambda passed to async_read_with_timeout

michael-jaia (approved):
Code tested on Fleet 0, Bot 1 (XBee 250MW) and Fleet 1, Bot 1 (XBee 900MW).

xbee-900
xbee-250

PR #896, opened 2024-06-10
Task/hub yubikey/jaia-68

tsaubergine:
# Overview 

This PR implements changes to fleet-config.sh to use a Yubikey installed in each Hub to SSH into the bots rather than a normal file-based SSH key pair. This greatly improves security as the compromise of the private file key alone is not sufficient to access the bot (as the Yubikey must be present as well).

# Hub Yubikey 

The Hub Yubikey SSH key pair is configured with "no-touch-required" so that the Hub is able to log into the Bots without any physical interaction with the Yubikey. This is weaker than typical Yubikey usage (which requires the touch / physical presence) but is necessary to provide the same embedded JCC data offload option we currently have. Future upgrades could add touch if such a level of security is required by a customer.

# Multiple Authorized Keys files

SSH authorized_keys are now split across several files (rather than just /home/jaia/.ssh/authorized_keys):

- /home/jaia/.ssh/authorized_keys - empty for customer use
- /etc/jaiabot/ssh/root_authorized_keys - Authorized Root (Master) Yubikeys
- /etc/jaiabot/ssh/hub_authorized_keys - Authorized Hub Yubikeys (this file exists on the bots only)
- /etc/jaiabot/ssh/tmp_authorized_keys - Authorized Temporary Yubikeys (employees, repair shop, etc.) - this is a placeholder to be used in a future PR.

This is to better segregate keys for different purposes and lower the chance of accidental removal of required keys.

# Fleet config

fleet-config.sh is rewritten to be primarily based around a new set of Ansible playbooks (in `jaiabot/config/ansible/fleet`). 

## ssh_key_setup

The original ssh_key_setup function is now 3 ansible playbooks that perform slightly different actions based on whether it's a virtual or real fleet:

- fleet/ssh-create-hub-keys.yml (Only requires Hub(s) to be connected):
   - Real: Copy pre-generated Yubikey SSH key pair to hub
   - Virtual: Generate SSH key pair on hub (same as previous behavior)
- fleet/ssh-copy-hub-keys.yml: Copy hub keys to bots
   - Real: Copy pre-generated Yubikey SSH public key to bots (Only requires Bot(s) to be connected)
   - Virtual: Copy SSH key pair from hub to bots (Requires both Hub(s) and Bot(s) to be connected).
- fleet/ssh-update-hubs.yml: (Requires both bot and hub to be connected): 
  - Real and virtual: Update hub /etc/hosts and known_hosts with Bot(s). 

This is split into three playbooks so we can eventually run these tasks atomically when required. 

See jaia-ssh-use-cases

We still have outstanding work to do to fully implement the second and third use cases on this diagram, see https://jaia-innovation.atlassian.net/browse/JAIA-1301, https://jaia-innovation.atlassian.net/browse/JAIA-1302, https://jaia-innovation.atlassian.net/browse/JAIA-1303.

## Retrofit existing fleets

A new task was added to fleet-config.sh (Retrofit SSH config) that runs the playbook "fleet/retrofit-ssh-config.yml" to reconfigure existing images to support this new configuration. This can be run on already deployed bots and hubs to allow them to transition to this new SSH configuration without reflashing the SD cards. This needs to be paired with the "Setup SSH Keys amongst fleet" action to fully reconfigure an existing fleet.

## Rename cloud playbooks

For cleanliness, the existing cloud playbooks have been moved to `jaiabot/config/ansible/cloud`).

This has been fully tested by creating a new Cloudhub and VirtualFleet to test.


ferrom-jaia (approved):
Looks great! Tested with Toby. (Fleet 0)

PR #898, opened 2024-06-10
task/add-adafruit_bno08x-repo-into-jaiabot-repo/jaia-1270

edsanville:
Use AdaFruit BNO08x submodule forked from the Adafruit repos rather than the pip version.


ferrom-jaia (changes requested):
The Adafruit_CircuitPython_BNO08x directory is there, but empty. You must run two commands: git submodule init to initialize your local configuration file, and git submodule update to fetch all the data from that project. 

I think we need to add this to the Dockerfile so that CircleCi packages things correctly. We should probably add instructions how to do this or incorporate this in the build.sh script.

tsaubergine:
We will need to include the submodule in the `git archive` output: https://github.com/jaiarobotics/jaiabot/blob/task/add-adafruit_bno08x-repo-into-jaiabot-repo-jaia-1270/.circleci/config.yml#L296-L304

 Something like this may work: https://ttboj.wordpress.com/2015/07/23/git-archive-with-submodules-and-tar-magic/

tsaubergine:
> We will need to include the submodule in the `git archive` output: https://github.com/jaiarobotics/jaiabot/blob/task/add-adafruit_bno08x-repo-into-jaiabot-repo-jaia-1270/.circleci/config.yml#L296-L304
> 
> Something like this may work: https://ttboj.wordpress.com/2015/07/23/git-archive-with-submodules-and-tar-magic/

This should do it (still need to test `apt install`): https://github.com/jaiarobotics/jaiabot/pull/898/commits/aa0f2e513e5c082d32ce847af864f6c7ef87f13e

edsanville:
> The Adafruit_CircuitPython_BNO08x directory is there, but empty. You must run two commands: git submodule init to initialize your local configuration file, and git submodule update to fetch all the data from that project.
> 
> I think we need to add this to the Dockerfile so that CircleCi packages things correctly. We should probably add instructions how to do this or incorporate this in the build.sh script.

Looks like you can do all of that in one command like so:
```git checkout -f --recurse-submodules```

I love the idea of submodules, but this was my only real issue with git.  They overcomplicate them so much.  I just want to check out the repository and make my directory the same as it was when committed, git!  Sheesh.

michael-jaia:
Ed, I used the command `git submodule update --init` because it also initializes the submodule (if it has not been)

PR #899, opened 2024-06-10
bug/moving-waypoint-with-edit-mode-off/jaia-1248

edsanville:
We probably want to refactor the waypoint movement into either:
a)  an Interaction that is added to the map's interaction stack, or
b)  a DragInteraction that allows waypoints to be moved, (only if the run is in Edit Mode, of course)

Ultimately, I think there should be two toolbar buttons:
a) Add Waypoint, (when in this mode, clicking the map adds a new waypoint to the active run, otherwise clicking does nothing)
b) Move Waypoints, (when in this mode, the waypoints become draggable, otherwise dragging only drags the map around)

Thoughts welcome!


ferrom-jaia (approved):
Looks good! I agree. We need a refactor at some point.

PR #902, opened 2024-06-11
bug/slow-first-boot/JAIA-1288

tsaubergine:
Problem: first-boot is slow on Raspi with no Ethernet cable connected to a DHCP server.

Solution:
- Use 192.168.123.45/24 dummy address for eth0 (this will get overwritten by jaiabot-embedded after that's installed).
- Update cloud init user data to write DHCP for cloud hub (as it's required here)


michael-jaia:
I removed the static address code and saw the same low-end first boot times (~3 mins) and did not see the output hang on the Wait for Network job. Since no clear difference came from changing the addresses to static, I'm only going to keep the code that removes the cloud packages from the raspi image.

tsaubergine:
OK cool.

PR #907, opened 2024-06-16
bug/jdv-bot-wpts-not-appearing-jaia-1284

edsanville:
BotStatuses and Commands weren't being loaded (from the h5 files) with the correct bot_ids.  This was due to a change in the group name format for those two messages.  Fixed the group name formats to be broader, and extract the bot_ids from the messages themselves instead of from the group names.


src/web/jdv/client/src/JaiaMap.ts
```diff
@@ -569,6 +569,8 @@ export default class JaiaMap {
             return;
         }
 
+        console.log(this.command_dict);
```
JRoybot: Were these log statements intended to stay or were they just needed for debugging?

src/web/jdv/server/jaia_h5.py
```diff
@@ -159,18 +159,23 @@ def commands(self):
...
                     
                     command_path = hub_command_path + '/jaiabot.protobuf.Command'
                     commands = jaialog_get_object_list(log_file[command_path], repeated_members={"goal"})
 
+                    try:
+                        bot_id = commands[0]['bot_id']
+                    except IndexError: # No commands
+                        continue
+
+                    if bot_id not in results:
```
JRoybot: results declared as empty dictionary on 156, is this check necessary?

JRoybot (commented):
A few comments, otherwise looks good

src/web/jdv/client/src/JaiaMap.ts
edsanville: Fixed!

src/web/jdv/server/jaia_h5.py
edsanville: Fixed!

JRoybot (dismissed):

src/web/jdv/client/src/JaiaMap.ts
edsanville: Also removed some other console.log statements.

ferrom-jaia (approved):
Looks good! Tested in sim.

PR #912, opened 2024-06-24
task/jcc-use-minimize-jaia-1280

edsanville:
Do not use devtool: eval-source-map option when building JCC for production, (reduces 35 MB product down to 7 MB!)

### DIFF
--- a/src/web/run.sh
+++ b/src/web/run.sh
@@ -27,16 +27,13 @@ pushd ${JAIA_DIR}/scripts/git-hooks/init/pre-commit/ > /dev/null
 popd > /dev/null
 
 
-mkdir -p ${BUILD_DIR}
-echo 🟢 Building JCC and JED into ${BUILD_DIR}
-npx webpack --mode production --env OUTPUT_DIR=${BUILD_DIR} --progress
-
-
 # Start server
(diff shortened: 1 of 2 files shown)


JRoybot (dismissed):
Looks good to me but did not understand it all

edsanville:
> Looks good to me but did not understand it all

I did do some refactoring of the webpack configuration.  

* We have the baseConfig, which is all the settings that should be shared regardless of mode (production or development) and target (JCC or JED).
* We have the productionConfig and developmentConfig, which have the settings for the build modes.
* We have the jccConfig and jedConfig, which have the settings for the target app.

The configuration is built by using the last line:

```    
    return [
        Object.assign({}, baseConfig, modeConfig, jedConfig),
        Object.assign({}, baseConfig, modeConfig, jccConfig),
    ];
```

This just takes a new empty object {}, and then updates it in place with the baseConfig, then the modeConfig, then the config for the target app...  basically fusing the three configs together.  I will add some comments to that effect.

JRoybot (approved):
Thanks for the explanation and additional comments

ferrom-jaia (approved):
Looks good! Tested on hub 1 fleet 2.

PR #913, opened 2024-06-24
task/move-rally-button-to-right-toolbar-jaia-1249

edsanville:
Moved Add Rally Point button to right toolbar.


JRoybot (approved):
Tested, with sim, all looks good.

michael-jaia (approved):
Launched simulator, new button location looks good. I like the flow it creates by being above the OMP button.

PR #914, opened 2024-06-26
task/remove-venv-when-removing-jaiabot-python-jaia-1281

edsanville:
Remove the python venv for both "remove" and "purge"


ferrom-jaia (approved):
Looks good!

PR #915, opened 2024-07-01
feature/collapse-expand-rc-panel-jaia-1256

edsanville:
Clicking anywhere in the header will actually toggle the min/max, but there are two buttons that act as indicators to the user that you can do this.


ferrom-jaia (approved):
Looks great! Tested in  sim.

PR #916, opened 2024-07-02
Parse ntpq for peer status code rather than waiting ntpstat

tsaubergine:
This will drive down our time to first bot status from cold power-on substantially. `ntpstat` seems excessively picky about declaring synchronization.

Needs testing on a bot/hub.


ferrom-jaia (approved):
Looks good! Tested on hub1 fleet 2.

PR #919, opened 2024-07-03
bug/jdv-map-headings-off

edsanville:
Do not filter out zeroes from lat/lon series on the backend, so the heading series always has the same length as the lat/lon series.


ferrom-jaia (approved):

PR #922, opened 2024-07-03
task/remove-gps-outliers-from-jdv-map

michael-jaia:


ferrom-jaia (approved):
Looks good. Tested in sim.

PR #923, opened 2024-07-06
bug/bno085-calibration-status-jaia-1271

edsanville:
Report an Accuracies message within the IMUData message, that describes the accuracy of the mag, gyr, and acc, as reported by the IMU.


edsanville:
@ferrom-jaia @michael-jaia 
Do you think we should report the Accuracies object in BotStatus?  I didn't do that, because it goes over intervehicle and would consume precious xbee bandwidth, as well as requiring an update to the intervehicle API version number.

I figure, if we get an IMU calibration issue on the water, we can plant a flag at that timestamp, and then take a look at the IMUData messages on the actual bot after we download the logs from the bot.

Let me know if we should add the Accuracies to the BotStatus message, though.

edsanville:
Using a side (non-main) branch for the Adafruit fork now.

ferrom-jaia (approved):
Looks good! Ed and I both built and deployed the software to bot 2 flee 1. I removed the deployed jaiabot folder on bot 2 to ensure environments were consistent.

PR #924, opened 2024-07-06
task/move-data-offload-button-to-top-toolbar

edsanville:
Move Download All button to top toolbar.


src/web/jcc/client/components/DownloadPanel.tsx
```diff
@@ -116,11 +106,6 @@ export default function DownloadPanel(props: Props) {
             <div className="panel-heading">Download Panel</div>
 
             <div className="download-panel-inner-container">
-                <div className="toolbar">
-                    <Button id="downloadAll" className={`button-jcc`} onClick={handleDownloadAll}>
-                        <Icon path={mdiDownloadMultiple} title="Download All" />
```
JRoybot, line 121: remove import of mdiDownloadMultiple on line 14

src/web/jcc/client/components/CommandControl/CommandControl.tsx
```diff
@@ -3122,6 +3123,9 @@ export default class CommandControl extends React.Component {
                 >
                     <Icon path={mdiPlay} title="Run Mission" />
                 </Button>
+                <Button id="downloadAll" className={`button-jcc`} onClick={this.processDownloadAllBots.bind(this)}>
```
JRoybot, line 3108: why is bind needed here? (for my own understanding)

JRoybot (changes requested):
Looks good.  Built and tested branch.  1 minor change requested

src/web/jcc/client/components/CommandControl/CommandControl.tsx
edsanville: It's because we need to bind the method to the value of "this" at that point.  onClick just expects a function with 1 argument (an Event object) or 0 arguments, and wouldn't know what "this" was without the bind.

JRoybot (approved):

ferrom-jaia (approved):
Looks good! Tested in sim.

Issue #887, opened 2024-06-04
NVM install directory issue


seantfish:
An EACCES error preceded by this error `./setup-tools-build-nodocker.sh: line 34: nvm: command not found` can occur if the install location of nvm is not `$HOME/.nvm`. Checking `XDG_CONFIG_HOME` may reveal the nvm install location to be `$HOME/.config/nvm`. This is caused by a behavior in the [install script](https://raw.githubusercontent.com/creationix/nvm/master/install.sh).

OS: Fedora Silverblue in Ubuntu 22.04 container

seantfish:
The reasoning is provided [here](https://github.com/npm/rfcs/issues/389)

ferrom-jaia:
Thank you Sean! We will add a ticket and fix.

seantfish:
Thank you!

PR #926, opened 2024-07-10
Setup fixes for NVM install directory and submodule URL

seantfish:
See #887 for NVM issue.

The new Adafruit submodule will not initialize on systems without an ssh key setup. This is not in the build documentation so it can cause an unexpected error for anyone who has not set one up. Switching this URL to https solves this issue and does not introduce credential issues as the submodule repository is public.


ferrom-jaia (approved):
Looks great!

PR #927, opened 2024-07-10
task/add-gps-outliers-to-jdv-map

seantfish:
Removed JDV outlier filter due to spurious filtering and to implement a more informative outlier filter upstream in the data


michaelrockJAIA (approved):
I tested the branch locally

PR #928, opened 2024-07-15
feature/color-map-for-contour-layer-jaia-715

edsanville:
Use Polygon objects with fill colors in the GeoJSON returned by the contour endpoints.


edsanville:
Whoops... forgot to document a few functions there!  Will come back to this one...

edsanville:
Ok, function comments in place!

src/python/pyjaia/pyjaia/contours.py
```diff
@@ -68,137 +99,179 @@ def geojson(features):
...
         fraction = 0
     else:
-        fraction = (value - pt0[2]) / (pt1[2] - pt0[2])
+        fraction = (depth - dive0.depth) / (dive1.depth - dive0.depth)
 
     return [
-        pt0[0] + fraction * (pt1[0] - pt0[0]),
-        pt0[1] + fraction * (pt1[1] - pt0[1])
+        dive0.lon + fraction * (dive1.lon - dive0.lon),
+        dive0.lat + fraction * (dive1.lat - dive0.lat)
```
JRoybot: Real edge case but shouldn't lat and lon be constrained (-90 - 90) (-180 - 180)

JRoybot (commented):
Didn't see anything wrong in the code

JRoybot (dismissed):
Code looks good.  Issues I found in testing appear to be existing bugs and not related to these changes.  Approving

src/python/pyjaia/pyjaia/contours.py
edsanville: Oh wow, good point!

JRoybot (dismissed):

ferrom-jaia (changes requested):
Before approving can we make the contour map non selectable. I can see an operator getting confused by this.

edsanville:
It turns out that the Translate interaction (for dragging Rally Points?) filtered by the `disableDrag` property, which made draggability opt-out in JCC.  

So, it turns out every feature was draggable, but we didn't notice because most features are being updated by the status polling!  It turns out the Rally Points were opting out anyway, so they were the only feature that were NOT draggable.

I converted it to `enableDrag` to make draggability opt-IN, instead.  If we want to make features draggable, we can enable this property, and work on the hooks for the Translate interaction to do the actual work of re-positioning stuff.

JRoybot (approved):
Nice find on the dragability!

ferrom-jaia (approved):
Looks great! Tested in sim.

PR #931, opened 2024-07-18
task/report_mid_battery_2

OverArchMC:
Reports 'vmid_voltage' data inside the BotStatus message, read from the midpoint of the two batteries onboard a Jaiabot.


ferrom-jaia:
Closing until this is tested and the priority for this is escalated.

PR #934, opened 2024-07-23
tasks / Support updating root yubikeys and security labels on PRs

tsaubergine:
# Update root yubikeys with software release update

## Overview 

This PR makes updates to so that the root yubikey authorized_keys (/etc/jaiabot/ssh/root_authorized_keys) can be changed with normal updates to jaiabot now, in the event that the root yubikeys are lost, stolen or need to be replaced for other reasons (e.g., newer version of the keys supporting new features).

## Additional PR labeling for verification 

To ensure this is done securely (and improve the likelihood that we will catch malicious code inserted to key parts of the repository), this PR adds additional labels (using .github/workflows/pr-labeler.yml) that are automatically attached to pull requests that modify files that are most likely to cause harm. This will be an extra hint to reviewers to pay attention to ensure these changes are correct (and not malicious):

- **rootfs**: Files in the rootfs generation directory (this can modify that core features of *newly* created fleets). This also includes changes to ssh key access.
- **debian**: Files in the debian directory (this can script changes that occur on each `apt update`)
- **github-actions**: Files in the github-actions (including the action that flags the PRs!).
- **circleci**: Files in .circleci which govern the automatic deployment of code
- **systemd**: Files in the systemd generation which govern how the various services are run.

I merged the existing PR labeling behavior into this, using the official `labeler@v5` (https://github.com/actions/labeler) action.

## Further considerations:

Another tool we can use is CODEOWNERS: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-and-branch-protection

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -20,3 +20,13 @@ rootfs:
 github-actions:
   - changed-files:
       - any-glob-to-any-file: '.github/**'
+
+circleci:
+  - changed-files:
+      - any-glob-to-any-file: '.circleci/**'
+
+systemd:
+  - changed-files:
+      - any-glob-to-any-file:
+          - 'config/gen/systemd*'
+          - 'config/templates/systemd/**'
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,8 @@
+debian:
+  - 'jaiabot/debian/**'
+
+rootfs:
+  - 'jaiabot/rootfs/**'
+
+actions:
+  - 'jaiabot/.github/workflows/**'
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -17,6 +17,6 @@ rootfs:
   - changed-files:
       - any-glob-to-any-file: 'jaiabot/rootfs/**'
 
-actions:
+github-actions:
   - changed-files:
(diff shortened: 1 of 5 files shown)


jason-jaia:
I didn't dig too deeply, but found that the tag protection rules are being deprecated soon.  https://github.blog/changelog/2024-05-29-sunset-notice-tag-protections/  Furthermore, the documentation for CODEOWNERS states that the rule sets are superior to the tag protection rules. I understand that the tag protection rules will get migrated, but I feel we should probably use the non-deprecated solution from the start.

tsaubergine:
> I didn't dig too deeply, but found that the tag protection rules are being deprecated soon. https://github.blog/changelog/2024-05-29-sunset-notice-tag-protections/ Furthermore, the documentation for CODEOWNERS states that the rule sets are superior to the tag protection rules. I understand that the tag protection rules will get migrated, but I feel we should probably use the non-deprecated solution from the start.

Oops wrong concept - I was thinking we could protect labels from being deleted. For some reason my brain conflated this with tags, which are obviously different.

ferrom-jaia (approved):
Looks great! Thank you for the in depth description!

PR #937, opened 2024-07-24
bug/jdv-zoom-on-vertical-axis-jaia-1513

edsanville:
If a plot zoom has no xrange, then set timeRange to null.


JRoybot (dismissed):
Code looks good.  Built and tested branch.  Could not reproduce original bug

ferrom-jaia (approved):
Looks great! Tested in sim.

PR #938, opened 2024-07-24
bug/jammy-image-does-not-boot

tsaubergine:
The `flash-kernel` script has gotten fancier in Ubuntu 22.04/jammy about detecting if it should run. 

As we are building the image on a EFI system in a chroot, we need a bit more coaxing to get it to run during image creation.

This small change fixes the create_raspi_base_image.sh script so it correctly installs the kernel and thus can be booted on the Raspberry Pi.

Tested on Raspberry Pi 4 hardware.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -186,6 +186,7 @@ filter-template-rootfs-builds: &filter-template-rootfs-builds
     branches:
       only:
         - "1.y"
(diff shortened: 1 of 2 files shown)


michael-jaia:
Tested on Hub Raspberry Pi
IMG_5325

PR #925, opened 2024-07-10
Feature/app for api/jaia-1460, jaia-1517, jaia-1550, jaia-1459, jaia-1514, jaia-1516

tsaubergine:
# Web endpoints

This PR implements a new REST API for accessing fleet data and sending commands. It is intended to supercede the preliminary API in the JCC server, which has been renamed to "/jaia/v0" to avoid conflict when running the new REST API and JCC on the same port, that is "/jaia/status" -> "/jaia/v0/status", etc.

This PR implements https://jaia-innovation.atlassian.net/browse/JAIA-1463 and most child issues, except as noted below in Follow-on work

## Merged PRs

This PR merges in the following separate PRs for final review:

- Auto-doc of the API and addition of the `jaia doc` subtool: https://github.com/jaiarobotics/jaiabot/pull/941
- Task Packets: https://github.com/jaiarobotics/jaiabot/pull/939
- pyjaia split into pyjaia and pyjaiaprotobuf

## Follow-on work

- Add read-only API key option.
- Add protobuf metadata for labeling message fields GUARANTEED so we can implement https://jaia-innovation.atlassian.net/browse/JAIA-1462 (but the fields set in this PR are notional - they need review as part of that task).
- Need autolinter for these GUARANTEED fields.
- JCC should be updated to use /jaia/v1 API endpoints whenever possible, with the goal of deprecating and removing the /jaia/v0 endpoints.

## Hosting

The Apache config has been updated to serve /jaia using this new REST API for ports 80 and 443 (https), and /jaia/v0 and / with the JCC server on port 80 (http only).

## REST API - Overview

The new API directly introspects from the new APIRequest Protobuf message for input and APIResponse for output. It accepts GET requests for the subset of requests that only have singular (non-repeated) non-recursive (no submessages) data. This is generally suitable for simpler requests (e.g. Stop all bots, query status, etc) but not for more complicated MISSION_PLAN type commands.

It also accepts POST for all requests, using JSON (which will be required for  the more complicated messages).

api_key can be omitted for local connections, if we want.

There are 2 variants of the API:

- "long" variant (e.g., 10.23.1.11:9092/jaia/v1/status/all) where the action and the target(s) are specified in the url and any data may be passed via GET arguments (subject to limitations mentioned above). The "long" version can be used for POST requests as well. Here the POST data (if any) contains the "action" submessage contents
[...]

PR #939, opened 2024-07-29
task/add-task-packets-to-rest-api-jaia-1517

edsanville:
Implemented the task packet API endpoint.

### DIFF
--- a/src/lib/messages/rest_api.proto
+++ b/src/lib/messages/rest_api.proto
@@ -9,8 +9,8 @@ package jaiabot.protobuf;
 
 message TaskPacketsRequest
 {
-    required uint64 start_date = 1 [(jaia.field).rest_api.presence = GUARANTEED];
-    required uint64 end_date = 2 [(jaia.field).rest_api.presence = GUARANTEED];
+    required uint64 start_time = 1 [(jaia.field).rest_api.presence = GUARANTEED];
+    required uint64 end_time = 2 [(jaia.field).rest_api.presence = GUARANTEED];
 }
 
 message CommandResult
--- a/src/lib/messages/rest_api.proto
+++ b/src/lib/messages/rest_api.proto
@@ -9,9 +9,8 @@ package jaiabot.protobuf;
 
 message TaskPacketsRequest
 {
-    required string start_date = 1
-        [(jaia.field).rest_api.presence = GUARANTEED];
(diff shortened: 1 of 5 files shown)


tsaubergine:
I changed the base branch to https://github.com/jaiarobotics/jaiabot/tree/feature/app-for-api/jaia-1460 to simplify review of this PR (as it builds on that work).

src/web/rest_api/common/time.py
```diff
@@ -7,3 +7,20 @@
 
 def utc_now_microseconds():
     return int(datetime.datetime.utcnow().timestamp() * 1e6)
+
+def iso_date_microseconds(iso_date: str):
```
tsaubergine: Do we still need this now that we're using the UNIX microseconds? If not I'd suggest removing (I want to keep the lines of code count in the REST API as low as possible)

src/lib/messages/rest_api.proto
```diff
@@ -9,9 +9,8 @@ package jaiabot.protobuf;
 
 message TaskPacketsRequest
 {
-    required string start_date = 1
-        [(jaia.field).rest_api.presence = GUARANTEED];
-    required string end_date = 2 [(jaia.field).rest_api.presence = GUARANTEED];
+    required uint64 start_date = 1 [(jaia.field).rest_api.presence = GUARANTEED];
```
tsaubergine: I'd prefer "start_time" and "end_time" for a unix timestamp field.

JRoybot (approved):

tsaubergine (changes requested):
Please add filtering by `jaia_request.target` - Right now I always get all the task packets, whereas if I send, for example:

`http://localhost:9092/jaia/v1/task_packets/b1?start_date=1722177986720894&end_date=1722350786720894`

I just want task packets from bot 1

edsanville:
> Please add filtering by `jaia_request.target` - Right now I always get all the task packets, whereas if I send, for example:
> 
> `http://localhost:9092/jaia/v1/task_packets/b1?start_date=1722177986720894&end_date=1722350786720894`
> 
> I just want task packets from bot 1

Whoops, yes this was an oversight.

tsaubergine:
Merging into https://github.com/jaiarobotics/jaiabot/pull/925; final review can happen there.

PR #945, opened 2024-08-07
task/rest-api-use-minimal-venv

edsanville:
The run.sh script in this branch uses a minimal venv, which only installs the pyjaiaprotobuf python package as discussed.


ferrom-jaia:
Closing until we decide we need this.

PR #946, opened 2024-08-07
bug/clicking-hub-button-no-gps-jaia-1520

edsanville:
-     Add type information to the HubContext variable.
-     Change the type of HubContextType.hubStatuses variable to be a hash of hub_id to PortalHubStatus.
-     Added the optional chaining operator to hubStatus?.location?.lat and lon
-     Change name of HubContextType.hubStatus to hubStatuses to indicate that it's a map onto PortalHubStatus objects.
-     Changed "Octal" to "Octet"
-     Don't return early before all hooks are generated (React doesn't like this and produces a console warning).

### DIFF
--- a/src/web/containers/HubDetails.tsx
+++ b/src/web/containers/HubDetails.tsx
@@ -38,10 +38,6 @@ export function HubDetails() {
 
     const hubContext = useContext(HubContext);
 
-    if (hubContext === null || !globalContext.showHubDetails) {
-        return <div></div>;
-    }
-
     const [accordionTheme, setAccordionTheme] = useState(
         createTheme({
             transitions: {
@@ -53,6 +49,10 @@ export function HubDetails() {
     useEffect(() => {
         addDropdownListener("accordionContainer", "hubDetailsAccordionContainer", 30);
     }, []);
+
+    if (hubContext === null || !globalContext.showHubDetails) {
+        return <div></div>;
+    }
 
     const firstKey = Number(Object.keys(hubContext.hubStatuses)[0]);
     const hubStatus = hubContext.hubStatuses[firstKey];
--- a/src/web/containers/HubDetails.tsx
+++ b/src/web/containers/HubDetails.tsx
@@ -54,8 +54,8 @@ export function HubDetails() {
         addDropdownListener("accordionContainer", "hubDetailsAccordionContainer", 30);
     }, []);
 
-    const firstKey = Object.keys(hubContext.hubStatus)[0];
-    const hubStatus = hubContext.hubStatus[firstKey];
+    const firstKey = Number(Object.keys(hubContext.hubStatuses)[0]);
+    const hubStatus = hubContext.hubStatuses[firstKey];
 
     /**
      * Dispatches an action to close the HubDetails panel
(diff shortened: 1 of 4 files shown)


edsanville:
Regarding the `HubContext` variable:  should be change this to `hubContext` to keep in line with the convention of only capitalizing types and not variables?

src/web/containers/HubDetails.tsx
```diff
@@ -54,8 +50,12 @@ export function HubDetails() {
         addDropdownListener("accordionContainer", "hubDetailsAccordionContainer", 30);
     }, []);
 
-    const firstKey = Object.keys(hubContext.hubStatus)[0];
-    const hubStatus = hubContext.hubStatus[firstKey];
+    if (hubContext === null || !globalContext.showHubDetails) {
+        return <div></div>;
```
JRoybot, line 54: Why do we need to return an empty element?

src/web/containers/HubDetails.tsx
```diff
@@ -196,11 +196,11 @@ export function HubDetails() {
                                     <HealthStatusLine healthState={hubStatus?.health_state} />
                                     <tr>
                                         <td>Latitude</td>
-                                        <td>{formatLatitude(hubStatus?.location.lat)}</td>
+                                        <td>{formatLatitude(hubStatus?.location?.lat)}</td>
                                     </tr>
                                     <tr>
                                         <td>Longitude</td>
-                                        <td>{formatLongitude(hubStatus?.location.lon)}</td>
+                                        <td>{formatLongitude(hubStatus?.location?.lon)}</td>
```
JRoybot, line 203: Rather than leaving lat/lon blank would it be more useful to display "N/A"?

JRoybot (commented):
A few questions but looks good... will test branch later and re-review

JRoybot (dismissed):
Built and tested branch

src/web/containers/HubDetails.tsx
edsanville: I believe we can return `false` here, (but `null` is recommended against):

https://medium.com/@davidkelley87/stop-using-return-null-in-react-a2ebf08fc9cd

src/web/containers/HubDetails.tsx
edsanville: Indeed!

src/web/containers/HubDetails.tsx
edsanville: Ok, so if I use `return false`, then TypeScript complains that React doesn't want the element to return a boolean value.  So, we're limited to either `null` or an empty div.  Given the warnings in the above article, perhaps an empty `div` actually is the safe way to go right now.

src/web/containers/HubDetails.tsx
edsanville: Ah, the formatLongitude and formatLatitude functions currently return "?" if they are passed an undefined or null value.  I forgot about that.  I will change it to "N/A".

michael-jaia (commented):
Statically reviewed the code...these changes definitely make the code clearer! Going to deploy to hub that does not have GPS connection.

michael-jaia:
* Deployed to Fleet 1, Hub 1
* Tested on hub with and without a GPS fix (no crashes, and expected values displayed)
* Clicked through HubDetails panel checking for undefined crashes (no crashes)

michael-jaia (dismissed):

ferrom-jaia (approved):
Looks great! Tested in sim.

PR #948, opened 2024-08-13
Bug/jcc bottom dive toggle/jaia 1568

JRoybot:
This is just the bug fix for the Bottom Dive bug fix sub-task.  Still trying to work out the integration tests to verify the functionality.

The code re-structure is in place to support the testing, which will go in the __tests_ directory

The bug fix was verified manually using Chrome and examining the new tasks using breakpoints.

The unit tests will be reviewed separately as a different sub-task of the original bug task


src/web/jcc/client/components/TaskSettingsPanel/TaskSettingsPanel.tsx
```diff
@@ -290,7 +290,7 @@ function TaskOptionsPanel(props: Props) {
                 </tr>
             );
         }
-        return <div></div>;
+        return <tr></tr>;
```
JRoybot, line 293: This was creating a warning because <div> is not allowed after <tbody>

src/web/jcc/package.json
```diff
@@ -5,9 +5,8 @@
     "description": "",
     "main": "server/index.js",
     "scripts": {
-        "build": "webpack --mode production",
-        "test": "webpack --mode production --watch --progress"
-    },
+        "build": "webpack --mode production"
```
JRoybot, line 8: Note this file prevents us from running the npm test from the jcc directory, need to run them at src/web.  For now just removed the stale "test" target, eventually would like to see the files merged so we can run tests at lower levels

src/web/jcc/client/components/TaskSettingsPanel/TaskSettingsPanel.tsx
```diff
@@ -39,7 +39,7 @@ function fmod(a: number, b: number) {
     return Number((a - Math.floor(a / b) * b).toPrecision(8));
 }
 
-interface Props {
+export interface Props {
```
JRoybot, line 42: needed to support tests

PR #950, opened 2024-08-13
bug/jcc-bottom-dive-toggle-fix/JAIA-1568

JRoybot:
Review of just the bug fix, follow on subtask will include unit and integration tests


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #959, opened 2024-08-20
bug/check-subscription-file-exists-before-deletion

michael-jaia:


ferrom-jaia (approved):
Looks good! Tested on hub and bot.

PR #960, opened 2024-08-21
task/update_salinity_sensor_calibration

OverArchMC:
Updates the calibration process for the Jaiabot's salinity sensor. The new version simplifies the process and automatically detects when the sensor's EC values begin to plateau. Attached are the new calibration instructions, both for customers and for in-house calibration. 
[In-house salinity sensor calibration procedure.pdf](https://github.com/user-attachments/files/16692662/In-house.salinity.sensor.calibration.procedure.pdf)
[Customer salinity sensor calibration instructions.pdf](https://github.com/user-attachments/files/16692663/Customer.salinity.sensor.calibration.instructions.pdf)


ferrom-jaia:
Closing pull request for now because we want to go a different direction.

michael-jaia:
Need to modify so calibration data is not cleared after each saved calibration point.

PR #967, opened 2024-08-26
feature/endpoint-plot-customer-data-jaia-1595

edsanville:
The endpoint is `jaia/v0/annotations`, and supports GET, POST, and DELETE (currently deletes all).

When POSTing a new point, use a GeoJSON Feature, with properties.data being a key-value mapping.

The Feature uses Mapbox `simplestyle` properties `marker-color` and `marker-size` to customize the feature appearances.


edsanville:
To test, you can:

1) start a simulation, from the `config/launch/simulation` directory
2) run the `src/web/run.sh` script
3) run the test script in `src/web/server/tests/run_tests.py`.  This will POST a single Point Feature to the annotations endpoint.

The endpoint is `http://localhost:40001/jaia/v0/annotations`, and the body of the message must be a GeoJSON feature, with the data in the `properties` of the Feature.  For example, the test script will submit something like the following:


```
{"type": "Feature", "id": 674839, "geometry": {"type": "Point", "coordinates": [144.939856, 3.963564]}, "properties": {"title": "Tada!", "marker-size": "large", "marker-color": "green", "data": {"depth": 50000, "salinity": 10}}}
```

src/web/jcc/client/components/Annotations.ts
```diff
@@ -0,0 +1,112 @@
...
+    constructor() {
+        setInterval(this.pollAnnotations.bind(this), 1000);
+    }
+
+    /** Polls the backend for new annotations, updating the feature source if there are new ones. */
+    pollAnnotations() {
+        jaiaAPI
+            .getAnnotations(this.version)
+            .then((response) => {
+                if (response.status == 304) {
```
JRoybot: Suggest using enumerated values for the status responses or at least annotate them with comments similar to app.py

JRoybot (dismissed):
All looks good, built and tested with sim.  One non-critical suggestion

src/web/jcc/client/components/Annotations.ts
edsanville: Good point!  Weirdly, it seems these constants aren't present in the JS standard libs.  Of all the languages to have those constants, I would expect JS to be at the top, but I guess not!

JRoybot (approved):

JRoybot (approved):

ferrom-jaia:
I am going to close this pull request until we are confident this needs to be added into the repo. This was a one off at a demo. Thank you for your effort in getting this spun up Ed! I will leave branch if we want to revisit.

PR #968, opened 2024-08-28
task/hub-comms-limit-default-on

michaelrockJAIA:
The hub comms limit layer has been updated to be on by default. 

I did not check what happens if the Hub does not have GPS.

We still need to address the potential for the user to what to show these limits even if the hub is not getting GPS but the user knows the hubs location


marshalllele (approved):
Verified that no errors occur and nothing breaks even if the hub has no GPS. Tested in sim.

michael-jaia:
## Without Hub GPS
image

## With Hub GPS

PR #969, opened 2024-08-28
feature/jdv-goby-conversion-directory-jaia-929

edsanville:
Use the parent of bot_offload as a temporary directory for h5 conversion, to avoid the s3 data deletion issue.


tsaubergine:
Code looks OK to me but I think the description is wrong? (parent of bot_offload, not tempfile module)

JRoybot (approved):

PR #973, opened 2024-09-19
bug/ip-incorrect-for-jdv-liaison-when-using-vpn-jaia-532

edsanville:
The client JS checks the location's hostname and matches it against "hXfX" or "172.23.X.X,"  and if either one matches then it assumes you're on the VPN and uses the "172.23" IP prefix in that case.

### DIFF
--- a/src/web/containers/HubDetails.tsx
+++ b/src/web/containers/HubDetails.tsx
@@ -32,11 +32,32 @@ import {
     mdiWrenchCog,
 } from "@mdi/js";
 
+/**
+ * Returns the IP prefix (two octets), corresponding to either the LAN or the VPN access.
+ *
+ * @param {string} hostname
+ * @returns {("172.23" | "10.23")} The IP prefix needed to create proper links to JDV, the router, upgrade, etc.
+ */
+function getIPPrefix(hostname: string) {
+    // For the VPN
+    // Return 172.23 for hostnames matching:
+    // * 172.23.x.x
+    // * bXfX
+    //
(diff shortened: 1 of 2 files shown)


michael-jaia:
Will we need to update this for use with the cloudhub? I have not used the web apps running in the cloud, but I believe they use IPv6. I'm also thinking this function might be more suited for a utility file so we do not cloud UI code with logic to resolve IP addresses.

edsanville:
> Will we need to update this for use with the cloudhub? I have not used the web apps running in the cloud, but I believe they use IPv6. I'm also thinking this function might be more suited for a utility file so we do not cloud UI code with logic to resolve IP addresses.

Good point for both.  I haven't used CloudHub and know nearly nothing about it.  Perhaps it's time to get that short introduction to that way of doing things, and I can come back and add IPv6 support.

In the meantime, I'll move this function off to a utility file.

ferrom-jaia (approved):
Looks good. I am going to add the task for the cloud hub links.

tsaubergine:
> Looks good. I am going to add the task for the cloud hub links.

Did this task ever get created? I still get a bad link (10.23.xxx.yyy) when clicking through to JDV from JCC in VirtualHub/CloudHub

ferrom-jaia:
I checked the backlog and I did not find it. I added the issue: https://jaia-innovation.atlassian.net/jira/software/c/projects/JAIA/boards/15/backlog?selectedIssue=JAIA-1835

PR #975, opened 2024-09-23
Fix upload of large files to CloudHub s3fs (/var/log/jaiabot/bot_offload)

tsaubergine:
## Background

s3fs uses /tmp to cache data to upload. /tmp is a tmpfs that is 50% of RAM (512M on t3a.micro). On Cloudhub, filling /tmp was causing OOM problems, and thus hanging.

## Solution

Short term: use /var/log/jaiabot as temporary directory for s3fs. Also fix size of /tmp to 256M to avoid OOM issues.

Long term (2.y): Switch to using physical disk for /tmp, rather than tmpfs: https://jaia-innovation.atlassian.net/browse/JAIA-1711?atlOrigin=eyJpIjoiMzg5ODZiMDI4OTNkNGYzNTk0OTg2ZTc3M2Y0MGI0YTUiLCJwIjoiaiJ9

### Retrofit existing CloudHub

```
sudo overlayroot-chroot
emacs /etc/fstab
# append ",size=256M" to end of options line for /tmp
# append ",tmpdir=/var/log/jaiabot" to end of options line for jaia--cloudhub-data--fleetN mount
# save and reboot
```

## Also

Substantially reduce size of filesystem by using linux-image-virtual rather than linux-image-generic.


tsaubergine:
Ready for review but leaving as draft as I'd like to test on AWS before merging.

ferrom-jaia (approved):
Looks good. Tested fleet 0 cloud hub data offload with 800MB file.

PR #977, opened 2024-09-25
bug/hub-cache-did-not-succeed/jaia-1708

tsaubergine:
The offline bundle pulls the files only from requirements.txt so this needs to be a complete list of all requirements.

At least 'setuptools' was missing, and so I update the list. I also added a script `sync_requirements.sh` that can be run to add all the dependencies to requirements.txt (given the current requirements.txt), so this should be run each time requirements.txt is updated or the pyjaia/pyjaiaprotobuf package dependencies are updated.

I don't have a quick way to test this updated bundle but I can if you need that.


edsanville:
It looks good to me! The script runs no problem with no updates to `requirements.txt` at the moment.

edsanville (approved):

PR #985, opened 2024-10-05
bug/contour-map-holes-jaia-1717

edsanville:
interpolate function returns None if the two dive depths are equal, and the calling function needs to use the appropriate default
Make sure the maxValue of depth makes it into the contourValues list


ferrom-jaia (approved):
Looks good! Tested in sim with the taskpackets we got from Michael Rock.

PR #986, opened 2024-10-08
Task/clean up existing npm packages update ol/jaia 1678/jaia 1712

JRoybot:
This Pull request supercedes #978 
It included updates for both JAIA-1678 Clean Up existing NPM Packages & JAIA-1712 Update Open Layers and Turf
It also fixes the error identified in JAIA-1558, which will be deleted

To test
remove existing build directory
remove jaiabot/src/web/node_modules directory
remove src/web/package-lock.json

rerun
jaiabot/scripts/setup-tools-build.sh
jaiabot/scripts/setup-tools-runtime.sh

rebuild
jaiabot/build.sh

Launch Simulator
jaiabot/config/launch/simulation/all.launch

Launch JCC/JED
jaiabot/src/web/run.sh

Launch JDV
jaiabot/src/web/run.sh

Verify all launch and run without errors.

Verify Node.js version is v20.17.0
~/jaiabot/src/web$ nvm ls
       v18.12.1
->     v20.17.0
         system
default -> v20.17.0
iojs -> N/A (default)
unstable -> N/A (default)
node -> stable (-> v20.17.0) (default)
stable -> 20.17 (-> v20.17.0) (default)
lts/* -> lts/iron (-> v20.17.0)
lts/argon -> v4.9.1 (-> N/A)
lts/boron -> v6.17.1 (-> N/A)
lts/carbon -> v8.17.0 (-> N/A)
lts/dubnium -> v10.24.1 (-> N/A)
lts/erbium -> v12.22.12 (-> N/A)
lts/fermium -> v14.21.3 (-> N/A)
lts/gallium -> v16.20.2 (-> N/A)
lts/hydrogen -> v18.20.4 (-> N/A)
lts/iron -> v20.17.0

Verify npm version is 10.9.0
~/jaiabot/src/web$ npm --version
10.9.0

PR #987, opened 2024-10-10
Task/common environment versions file/jaia 1723

tsaubergine:
## Rationale

Previously we had the NPM and Node versions explicitly specified in 5 or so different files. Also, we didn't explicitly call out the Arduino CLI or NVM versions to install.

This PR will make it easier to update these versions, and in the case of Arduino or NVM, makes sure we are always installing the same known version for consistency.

## Updates

- New `scripts/common-versions.env` file captures versions of everything we're using (except requirements.txt which already captures Python dependency versions): 
```
jaia_version_release_branch=1.y
jaia_version_ubuntu=20.04
jaia_version_ubuntu_codename=focal
jaia_version_nodejs=v20.17.0
jaia_version_arduino_cli=v1.0.4
jaia_version_nvm=v0.40.1
jaia_version_npm=10.9.0
jaia_version_webpack=5.95.0
jaia_version_webpack_cli=5.1.4
```
- Removed release-branch file as this is contained now in the new common-versions.env
- Updated all scripts to use this new file
- Moved all Dockerfile to Dockerfile.in and configure using CMake variables (read from common-versions.env).
- Changed build output directory for ./docker-arm64-build-and-deploy.sh so it is now dependent on distro and release version (`build/jammy-1.y-amd64-vbox` rather than `build/amd64-vbox` and `build/jammy-1.y-arm64-vbox` instead of `/build/arm64`) to allow for multiple different builds (jammy/1.y, focal/1.y and/or noble/2.y) to happily coexist on a dev machine.
- `package.json` renamed to `package.json.in` and substitutions for webpack and webpack-cli versions are performed using the existing CMake configure_file step.

## Testing performed

- Ran `./docker-arm64-build-and-deploy.sh` successfully on a focal/1.y VirtualBox bot and hub
- Rebuilt and pushed Jammy docker images to DockerHub and tested that CircleCI build runs sucessfully.
- Built and tested sim-docker


scripts/common-versions.env
```diff
@@ -0,0 +1,9 @@
+jaia_version_release_branch=1.y
```
JRoybot, line 1: We'll need to coordinate how this merges with https://github.com/jaiarobotics/jaiabot/pull/986#issue-2573624542

My changes to the existing installs include the following versions.
nodejs v20.17.0
npm 10.9.0
webpack5.95.0 
webpack-cli 5.1.4

JRoybot: Note that webpack & webpack-cli versions are specified in src/web/package.json, which will need to be kept in sync with everything else.

JRoybot (commented):
Great job consolidating all of the versions.  I left a comment with the version in my open pull request 
https://github.com/jaiarobotics/jaiabot/pull/986#issue-2573624542

I am not sure how to use env variables in the package.json file, but assume there is a way.

I think it should be pretty easy to merge the 2 PRs if #986 gets merged first as I did not change the file structure

scripts/common-versions.env
tsaubergine: I'm fine with #986 merging first, then we can update this PR to those version numbers.

For the package.json, we could do what I did with the Dockerfile to create a Dockerfile.in and then have it pre-processed by CMake to the Dockerfile (in this case package.json.in -> package.json).

JRoybot: Sounds like a good plan. :-)

tsaubergine: I've updated this to reflect changes in #986 and to preprocess package.json.

tsaubergine:
Ready for review again after incorporating 1.y changes from #986

src/web/CMakeLists.txt
```diff
@@ -6,12 +6,14 @@
 # Get the web source tree into the intermediate directory
 set(intermediate_web_dir ${project_INTERMEDIATE_DIR}/web)
 file(MAKE_DIRECTORY ${intermediate_web_dir})
-configure_file(${CMAKE_CURRENT_SOURCE_DIR}/package.json ${intermediate_web_dir}/ COPYONLY)
+
+include(ReadCommonVersions)
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/package.json.in ${intermediate_web_dir}/package.json @ONLY)
```
JRoybot: Does this create the package.json file in src/web dir needed by run.sh?

JRoybot: realized it ends up in ~/jaiabot/build/intermediate/web, can we put a copy in jaiabot/src/web as well?

tsaubergine: All set now that I used a similar approach to the Dockerfiles. ./run.sh is updated to generate the package.json in jaiabot/src/web using the same CMake code.

JRoybot (approved):
Clean build and tested with simulation

tsaubergine:
> I noticed after approving the branch that on my local repo `src/web/package-lock.json ` had been modified. The diff looks like only minor version updates. I suspect it may be because I removed the lock file and the node_modules directory before running src/web/run.sh.
> 
> If others find that building this branch updates the lock file we may want to include the new version in the merge. If not it is likely just due to me deleting the file and the directory.
> 
> I put the lock file under CM control in the NPM update PR #986 based on documentation I read on maintaining node.js applications. The reason was by putting it under CM we will if we or any clients are running a different environment.

Are we using `npm ci` to make sure that the package-lock.json is actually used? If not I'd prefer to not have it in Git: https://stackoverflow.com/questions/52499617/what-is-the-difference-between-npm-install-and-npm-ci

edsanville:
From the linked page:

> Essentially, npm install reads package.json to create a list of dependencies and uses package-lock.json to inform which versions of these dependencies to install. If a dependency is not in package-lock.json it will be added by npm install.

This indicates to me that, if we commit our `package-lock.json` file, and that file contains all of the packages we use, then `npm install` will install those locked versions, (without the time-consuming clean installation of `npm ci`).

Therefore, it seems to me that `npm ci` isn't that useful for us, unless we're seeing some kind of package problem that needs them to be cleanly installed.

edsanville:
If we're committing `package-lock.json`, then we should copy it to `build/<arch>/intermediate/web` during the build process, to ensure that we get our locked versions.

tsaubergine:
Can we merge this and move the package-lock.json discussion / development to another Task?

JRoybot (approved):

PR #989, opened 2024-10-11
task/update-global-protobuf-extension-number

tsaubergine:
Google finally accepted my PR that includes our global option extension number (to allow global compatibility with other users of Protobuf option extensions):

https://github.com/protocolbuffers/protobuf/commit/46f7879cae6a986ca0f780811dc8414ae14d18d5

Given their delay they had to update the tag index to 1252, which is our final allocated value. This PR updates the extension IDs to this accepted value.

This change should have no effect on our codebase other than ensuring we are compatible with all potential users of Protobuf.

### DIFF
--- a/src/lib/messages/option_extensions.proto
+++ b/src/lib/messages/option_extensions.proto
@@ -32,17 +32,20 @@ message EnumValueOptions
     optional RestAPI rest_api = 1;
 }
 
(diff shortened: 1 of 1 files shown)


ferrom-jaia (approved):
Looks good.

PR #990, opened 2024-10-15
bug/salinity-calibration-high-low-jaia-1707

edsanville:
This PR involves the script: `src/python/atlas_scientific_ezo_ec/calibrate_atlas_oem.py`, which needs to be run on a physical bot with salinity probe attached.

* A "Set temperature compensation" option was added, which allows you to set the temperature compensation value (in degrees)
* A bug was fixed, where the high and low calibration numbers were reversed!
* The high and low calibration process prints live raw sensor values, so the user can see when the reading has settled down before committing the calibration value.

To test, run the `src/python/atlas_scientific_ezo_ec/calibrate_atlas_oem.py` script on a physical device with attached salinity probe.  This will bring up a menu, from which you can set the temp calibration value, as well as run through the high and low calibration process.

The true test will, of course, be an actual salinity sensor calibration on a real ocean-going device.


src/python/atlas_scientific_ezo_ec/calibrate_atlas_oem.py
```diff
@@ -65,6 +78,21 @@ def doCalibration(description: str, type: int):
     else:
         value = 0
 
+    while True:
```
JRoybot, line 81: Not part of the update but I did see a warning that line 76 above should be 
`if description != 'DRY':`

JRoybot: Dont think "is" tests should be used with literals

JRoybot (changes requested):

JRoybot (approved):

ferrom-jaia (approved):
Looks good!

PR #992, opened 2024-10-21
Task/add testing libraries to npm environment/jaia 1679

JRoybot:
This was a very minor update as most of what is needed for jest/react testing was in place.  Added user-events as a preferred method of triggering events in react.  This module interacts with the components as a user would as opposed to directly with the dom like fireEvent from testing-library/react.  

Note we can add additional modules as necessary when we add more testing capabilities, this should be adequate for things we have tried to test so far.


edsanville (dismissed):
Tested, builds and runs fine with this minor update

PR #993, opened 2024-10-25
task/remove-power-data-from-botstatus-message/JAIA-1728

ferrom-jaia:
## Description
Removing power settings from the bot status message as the customer does not need this data real time. This also reduced the size of the bot status message to 46 bytes.

## Testing
Launch a simulation and verified the bot performed a mission without any errors.


ferrom-jaia (commented):
Updates look good!

PR #995, opened 2024-10-25
Task/add simulator docker container to circleci/jaia 1731

tsaubergine:
# Overview

- Builds Docker image for simulations. Use existing `scripts/sim-docker/Dockerfile.in` to build image, but updates the Dockerfile to use the local branch of jaiabot rather than the release branch.
- Pushes Docker image to https://hub.docker.com/repository/docker/gobysoft/jaiabot-sim/general with a tag of `1.y-test`, `1.y-continuous`, `1.y-beta`, or the release version `1.14.0`. This means we keep a copy of each release but only the latest in each deb repo (beta, continuous, test) for each release version (1.y, etc.).
- Update README.md to mention how to download the image as the preferred method rather than building it yourself (which is still left as an option).
- Moved CMake version logic to `cmake/JaiaVersions.cmake` from `/CMakeLists.txt` so it can be called by the `cmake/ConfigureDockerfiles.cmake` script outside of the normal CMake configure step.

- ~To do - install .debs rather than rebuild jaiabot when running on CircleCI.~: This seems to overcomplicate things, and requires waiting for the .debs to build first. I left the Dockerfile as is (build from source).

## Testing performed

Built (using CircleCI) a 1.y-test image and ran the simulator. No good way to test the 1.y-continuous, etc. image until this PR is merged.

```
docker pull gobysoft/jaiabot-sim:1.y-test
docker tag gobysoft/jaiabot-sim:1.y-test jaiauser:jaia-sim-image
cd jaiabot/scripts/sim-docker
./launch-container.sh
```


ferrom-jaia (approved):
Looks great! Tested pulling the image and worked as expected.

PR #996, opened 2024-10-28
bug/bot-will-hold-on-ocean-floor-jaia-1733

edsanville:
Bug seems to be due to the Integral component of the PID controller.

*    PID target is initially 10.0 m

*    Motor goes max reverse (not incorrect)

*    Bot hits bottom at 5.48 m

*    Desired depth gets updated to 5.48 m

*    The P an D components are now ~0, but the I component has big magnitude from that period of descent where depth was far off

*    This Integral component has no mechanism for decreasing, as the bot never overshoots its target due to the seafloor being there

My solution was to clear the Integral component of the PID controller whenever the target_depth is updated.

Something else to consider:

* optionally, use a desired_depth slightly above measured seafloor?

This would prevent it from grinding against the seafloor if the bot can't quite reach the target_depth due to geometry/rounding errors in the pressure measurement.


edsanville:
Two findings:
* I needed to also reset the I term for `desired_setpoint` dives, in addition to `engineering_command` dives, since those are different code paths, doh!
* I removed the `THROTTLE_FOR_ZERO_NET_BUOYANCY`, because I don't think it was helping the PID controller, and it may result in grinding against the seafloor because the `ControlSurfaces/motor` was always at least -35 (or whatever `THROTTLE_FOR_ZERO_NET_BUOYANCY` was set to).

NOTE:  the simulator doesn't respond directly to `ControlSurfaces/motor`, it responds only to `DesiredSetpoints`, and therefore the PID controller may have some residual motor action from the `I` term, even though the desired depth has been reached.
(see `src/bin/simulator/simulator.cpp` line 301)

marshalllele (approved):
Looks good. Tested on the bench as well as on the water. Compared the dive profiles and motor values of two bots running release and then this branch to ensure that no unexpected behaviors arose.

PR #1001, opened 2024-11-01
task/eliminate-leading-zeros-in-task-parameter-inputs-JAIA-1751

michael-jaia:
### Before
https://github.com/user-attachments/assets/a5b8c1b7-c59b-4a69-9ad2-679c95635a7f

### After
https://github.com/user-attachments/assets/62d0d4c5-af2d-491f-bbc7-362ad6b7582b

### Testing
Entered a variety of values into all of the task parameter input sections (decimals, negative values, integers, alphabetical characters, and symbols). The default input handling manages the different values as expected. One possible improvement going forward would be to remove the red outline for values that do not match the expected precision defined in the dccl field and provide a more helpful message.


ferrom-jaia (approved):
Looks great! Nice solution. Tested in sim.

PR #1002, opened 2024-11-04
task/document-web-files-jaia-1750

edsanville:
* Start some documentation on the Jaia web code.  Gives an overview of folder structure, config files and what they do, as well as basic instructions to build and run the web apps.
* Also fixes the jest error (caused by not transforming some of the node_modules that are written using ECMAScript import syntax)


src/web/jest.config.js
```diff
@@ -5,198 +5,198 @@
...
-    "^.+\\.(js|jsx)$": "babel-jest",
-  },
+    // A map from regular expressions to paths to transformers
+    transform: {
+        "^.+\\.(ts|tsx)?$": "ts-jest",
+        "^.+\\.(js|jsx)$": "babel-jest",
+    },
 
-  // An array of regexp pattern strings that are matched against all source file paths, matched files will skip transformation
-  transformIgnorePatterns: ["<rootDir>/node_modules/(?!(ol|quick-lru|p-limit|yocto-queue)/)"],
```
JRoybot: Nice fix! after reading the docs related to this I am a bit confused why this change worked.  Doesn't the `?!` regex operator mean it should NOT ignore anything in node_modules that matches `ol|quick-lru|p-limit|yocto-queue` pattern?


Am, I reading it wrong?
https://jestjs.io/docs/configuration#transformignorepatterns-arraystring

src/web/README.md
```diff
@@ -1,5 +1,67 @@
...
+
+To run a specific test:
+
+```
+npm test <path to test file>
+```
+
+```
+npm test jcc/client/components/CommandControl/__tests__/CommandControl.test.tsx
+```
```
JRoybot, line 78: I would like to see an explanation of Babel's role in the web environment included here.  I am not sure I completely understand how it works in both jest and runtime environments and how the babel.config is used by webpack and jest.

JRoybot (changes requested):

src/web/jest.config.js
edsanville: You're not reading it wrong.  The problem is that the code error actually occurs in `node_modules/color-space`.  If you add that to the exclusion regex, then yet another module gets untransformed, etc.  By removing all regexes, it will transform all of `node_modules`.

src/web/README.md
edsanville: Good point, I'm learning this as I go myself.  I will add some bits about how babel fits in, and where and when it actually gets used in the pipeline.

JRoybot (dismissed):
Thanks for adding the Babel stuff!

ferrom-jaia (approved):
This is great documentation!

PR #1003, opened 2024-11-04
task/Make vpn.jaia.tech and packages.jaia.tech use similar SSH authentication with temporary keys as hubs and bots/jaia-1725

tsaubergine:
## Rationale

packages.jaia.tech and vpn.jaia.tech are important servers with significant security risks. Rather than use a long list of .ssh/authorized_keys, we should move to the same root / tmp Yubikey security model used by the hubs and bots.

## Updates

This PR adds `xxx.jaia.tech` as a valid `host` string for the `jaia` tool so you can run commands like:

```
jaia admin ssh add --user=ubuntu vpn.jaia.tech <some key> <some duration>
```

I already added /etc/ssh/sshd_config.d/jaia_sshd.conf (from jaiabot/rootfs/customization/includes.chroot/etc/ssh/sshd_config.d/jaia_sshd.conf) and /etc/jaiabot/ssh/root_authorized_keys (from jaiabot/rootfs/customization/includes.chroot/etc/jaiabot/ssh/root_authorized_keys) to both packages.jaia.tech and vpn.jaia.tech. I also took an AWS disk snapshot of both machine's hard drives in case we accidentally lock ourselves out during these changes.

## To do

~We need to clear the .ssh/authorized_keys file after verifying that `jaia ssh admin add` works correctly with the master Yubikey on packages.jaia.tech and vpn.jaia.tech.~: Done 1/13/25 by @tsaubergine


ferrom-jaia (approved):
Works as expected!

PR #1005, opened 2024-11-05
task/add-npm-test-to-circle-ci/JAIA-1734

JRoybot:
Copied required files for jest testing over to the build directory, added command to circleci to run all jest tests.  See circleci output from this PR to verify tests ran.
Note there is overlap between this and Ed's PR https://github.com/jaiarobotics/jaiabot/pull/1002

This should not be merged before that one.


tsaubergine:
CircleCI changes are fine with me; I can't comment on the rest.

tsaubergine (dismissed):

edsanville:
It looks good and works fine.

A related suggestion:  in `src/web/CMakeLists.txt`, maybe we should just copy the entire `src/web` directory (omitting `node_modules`) to the `build/intermediate/web` directory?  It would tighten up that file, and eliminate the need to keep adding more file copies in there.

JRoybot (commented):
This is a great idea, only concern is it will copy over some files that are not needed, most are small in insignificant, but if the web stuff has been built there will be a `dist ` directory that is not needed.  Also since `package.json` is now generated directly into the `intermediate_web_dir` above (line 12), we might want to exclude that from being copied over so a stale copy doesn't get in by accident.  Thoughts?

edsanville:
> This is a great idea, only concern is it will copy over some files that are not needed, most are small in insignificant, but if the web stuff has been built there will be a `dist ` directory that is not needed. Also since `package.json` is now generated directly into the `intermediate_web_dir` above (line 12), we might want to exclude that from being copied over so a stale copy doesn't get in by accident. Thoughts?

That's true.  Excluding `package.json` and `dist` is an easy enough fix, however fixing `jdv/client/dist` is more challenging due to the fact that I'm not using a recursive glob for this.  I will make a commit to omit those.

I view the tradeoff like this:
* Pros:  duplicate the src directory into the intermediate directory exactly, so we never see weird bugs related to config files or new directories not being copied over before the build.
* Cons:  copying over some unnecessary data (looks like about 1.12 MB on my system, 300k if we exclude `dist/`)

edsanville (approved):

ferrom-jaia (approved):
Looks good! Checked circleci to make sure the tests were runnning.

PR #1011, opened 2024-11-11
Bug/first boot fails on 1.14.0/Jaia-1758

tsaubergine:
Fixes failure in first boot on release 1.14.0:

Screenshot from 2024-11-11 16-24-26

- Set `DEBIAN_FRONTEND=noninteractive` so initial `apt install` of `jaiabot-embedded` doesn't fail
- Move authorized_keys files out of rootfs install as these are now in the Debian package
- Update the `tmp_authorized_keys` file after initial install of `jaiabot-embedded`
- Tested on VirtualBox image (focal).


ferrom-jaia (approved):
Looks good!

PR #1014, opened 2024-11-12
Task/MUI select test with callback example/jaia 1683/jaia 1686

JRoybot:
This Pull Request uses the MUI Select Component in the `TaskSettingsPanel` as an example of how to test these widgets JAIA-1683.  It utilizes a mock callback to change the props passed into the `TaskSettingsPanel` JAIA-1686.  It rerenders `TaskSettingsPanel` and validates the new props.  It also verifies the callback has been called the correct number of times and with the right Task Type.

`TaskSettngsPanel` was moved to it's own folder to follow the established folder structure from other tests.

The Select component was modified to use Native format as this greatly simplifies the resulting HTML and facilitates simpler testing.

The only method of getting the right element to trigger events that worked was using `data-testid` set on the `inputProps` to get the underlying input component and not the label or button used to decorate it.

Because the value of the Select Component is based on Props and requires a rerender to show updated values, a `mockOnChange` callback function was used to modify the Props with the new data created by the logic in the `TaskSettingsPanel`. The `jest-dom` does not automatically rerender components when their props are updated, that has to be done explicitly after the `mockOnChange `updates the Props.

A helper function was created to validate the contents of the new Task created by the panel. You can not put assertions in the mock functions as they only work when part of a test. Once the Props are updated by the `mockOnChange`, the helper function is used to validate the newly created task. This function was placed in the shared src/web/tests/helpers folder


src/web/tests/jest.setup.js
```diff
@@ -1,19 +1,19 @@
...
+
     disconnect() {}
 };
 
 // Silence output while running tests
-global.console.log = jest.fn()
-global.console.error = jest.fn()
-global.console.debug = jest.fn()
+global.console.log = jest.fn();
+global.console.error = jest.fn();
```
edsanville: I'm wondering if we should let errors show up.

src/web/tests/helpers/ValidateTask.tsx
```diff
@@ -0,0 +1,49 @@
+import { MissionTask, TaskType } from "../../jcc/client/components/shared/JAIAProtobuf";
+//Helper function to verify the task created by the Panel is valid
+export function validateTask(task?: MissionTask): void {
+    if (!task) {
+        return;
+    }
+    switch (task?.type) {
+        case TaskType.CONSTANT_HEADING:
+            //make sure all the parameters are defined
+            expect(task.constant_heading.constant_heading).toBeDefined();
```
edsanville: Perhaps an `expect(task.constant_heading).toBeDefined()` first?

src/web/tests/helpers/ValidateTask.tsx
```diff
@@ -0,0 +1,49 @@
...
+            expect(task.constant_heading.constant_heading_speed).toBeDefined();
+            expect(task.constant_heading.constant_heading_time).toBeDefined();
+            //none of these should be present
+            expect(task.dive).toBeUndefined();
+            expect(task.station_keep).toBeUndefined();
+            expect(task.surface_drift).toBeUndefined();
+            break;
+        case TaskType.DIVE:
+            if (task.dive.bottom_dive) {
```
edsanville: Same here, `expect(task.dive).toBeDefined()`?

JRoybot (commented):
thanks Ed

edsanville (dismissed):

src/web/tests/jest.setup.js
JRoybot: I let all log statements show up during debug.  Twomey set it this way so I restored it to how it was.  Either setting seems fine to me.  Easy enough to turn them back on if something breaks

edsanville (dismissed):

src/web/jcc/client/components/TaskSettingsPanel/__tests__/TaskSettingsPanel.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+    onChange: (task?: MissionTask) => mockOnChange(task),
+};
+
+//Mock of the onChange callback to update Props
+const mockOnChange = jest.fn().mockImplementation((task?: MissionTask) => {
+    mockProps.task = task;
+});
+
+describe("MUI Select Component Examples", () => {
+    test("Get by Test ID, Select all Options", async () => {
```
ferrom-jaia, line 20: What do you think about this test name: TaskSettingsPanel: Should update task type correctly for all options

src/web/jcc/client/components/TaskSettingsPanel/__tests__/TaskSettingsPanel.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+        // Change the Task Type to all options and verify the change occured
+        for (const value of Object.values(TaskType)) {
+            // Change the selection
+            await userEvent.selectOptions(selectElement, value);
+
+            // Rerender with updated props
+            rerender(<TaskSettingsPanel {...mockProps} />);
+
+            // Verify that the selected value is correct
+            await waitFor(() => {
```
ferrom-jaia, line 47: Should we include a timeout in case this gets stuck?

await waitFor(() => {
    expect(selectElement.value).toBe(value);
}, { timeout: 1000 });

src/web/jcc/client/components/TaskSettingsPanel/__tests__/TaskSettingsPanel.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+            if (value != TaskType.NONE) {
+                expect(mockOnChange).toHaveBeenCalledWith(
+                    expect.objectContaining({
+                        type: value,
+                    }),
+                );
+            }
+        }
+    });
+});
```
ferrom-jaia, line 68: Consider testing invalid input scenarios, e.g., if the task-select-input-id is missing or an unexpected value is passed to the select.

src/web/tests/helpers/ValidateTask.tsx
```diff
@@ -0,0 +1,58 @@
...
+            expect(task.dive).toBeUndefined();
+            expect(task.surface_drift).toBeUndefined();
+            expect(task.constant_heading).toBeUndefined();
+            break;
+        case TaskType.NONE:
+            expect(task.constant_heading).toBeUndefined();
+            expect(task.dive).toBeUndefined();
+            expect(task.surface_drift).toBeUndefined();
+            expect(task.station_keep).toBeUndefined();
+    }
```
ferrom-jaia: Should we add a default case?
default:
    throw new Error(`Unhandled TaskType: ${task.type}`);

src/web/tests/jest.setup.js
ferrom-jaia: I agree we should let errors show up

src/web/jcc/client/components/TaskSettingsPanel/__tests__/TaskSettingsPanel.test.tsx
ferrom-jaia: Ensure the validateTask function itself is tested. Create unit tests for edge cases like:
Null or undefined task.
A TaskType with incorrect or extra parameters.

src/web/tests/helpers/ValidateTask.tsx
```diff
@@ -0,0 +1,58 @@
+import { MissionTask, TaskType } from "../../jcc/client/components/shared/JAIAProtobuf";
+//Helper function to verify the task created by the Panel is valid
+export function validateTask(task?: MissionTask): void {
+    if (!task) {
+        return;
+    }
+    switch (task?.type) {
+        case TaskType.CONSTANT_HEADING:
+            //make sure all the parameters are defined
```
ferrom-jaia: Nitpicking: Should we capitalize and add a space with the comment? Not sure just want it to be consistent.

src/web/tests/helpers/ValidateTask.tsx
```diff
@@ -0,0 +1,58 @@
...
+            expect(task.constant_heading).toBeDefined();
+            expect(task.constant_heading.constant_heading).toBeDefined();
+            expect(task.constant_heading.constant_heading_speed).toBeDefined();
+            expect(task.constant_heading.constant_heading_time).toBeDefined();
+            //none of these should be present
```
ferrom-jaia: Nitpicking: Should we capitalize and add a space with the comment? Not sure just want it to be consistent.

src/web/tests/helpers/ValidateTask.tsx
```diff
@@ -0,0 +1,58 @@
...
+            expect(task.dive).toBeUndefined();
+            expect(task.station_keep).toBeUndefined();
+            expect(task.surface_drift).toBeUndefined();
+            break;
+        case TaskType.DIVE:
+            expect(task.dive).toBeDefined();
+            if (task.dive.bottom_dive) {
+                //Bottom Dive = true, expect no other parameters
```
ferrom-jaia: Nitpicking: Should we capitalize and add a space with the comment? Not sure just want it to be consistent.

src/web/tests/helpers/ValidateTask.tsx
```diff
@@ -0,0 +1,58 @@
...
+                expect(task.dive.max_depth).toBeUndefined();
+                expect(task.dive.depth_interval).toBeUndefined();
+                expect(task.dive.hold_time).toBeUndefined();
+            } else {
+                //Non Bottom Dive should have other parameters
```
ferrom-jaia: Nitpicking: Should we capitalize and add a space with the comment? Not sure just want it to be consistent.

src/web/tests/helpers/ValidateTask.tsx
```diff
@@ -0,0 +1,58 @@
...
+                expect(task.dive.max_depth).toBeDefined();
+                expect(task.dive.depth_interval).toBeDefined();
+                expect(task.dive.hold_time).toBeDefined();
+            }
+            expect(task.surface_drift).toBeDefined();
+            //none of these should be present
```
ferrom-jaia: Nitpicking: Should we capitalize and add a space with the comment? Not sure just want it to be consistent.

src/web/tests/helpers/ValidateTask.tsx
```diff
@@ -0,0 +1,58 @@
...
+            expect(task.station_keep).toBeUndefined();
+            expect(task.constant_heading).toBeUndefined();
+            break;
+        case TaskType.SURFACE_DRIFT:
+            expect(task.surface_drift).toBeDefined();
+            expect(task.surface_drift.drift_time).toBeDefined();
+            //none of these should be present
```
ferrom-jaia: Nitpicking: Should we capitalize and add a space with the comment? Not sure just want it to be consistent.

src/web/jcc/client/components/TaskSettingsPanel/__tests__/TaskSettingsPanel.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+            rerender(<TaskSettingsPanel {...mockProps} />);
+
+            // Verify that the selected value is correct
+            await waitFor(() => {
+                expect(selectElement.value).toBe(value);
+            });
+            // Validate the TaskSettingsPanel sent a valid Task to mockOnChange
+            validateTask(mockProps.task);
+
+            //Verify the mockOnChange function has been called the right number of times
```
ferrom-jaia: Nitpicking: Should we capitalize and add a space with the comment? Not sure just want it to be consistent.

src/web/jcc/client/components/TaskSettingsPanel/__tests__/TaskSettingsPanel.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { MissionTask, TaskType } from "../../shared/JAIAProtobuf";
+import { validateTask } from "../../../../../tests/helpers/ValidateTask";
+
+let mockProps: Props = {
+    isEditMode: true,
+    enableEcho: false,
+    onChange: (task?: MissionTask) => mockOnChange(task),
+};
+
+//Mock of the onChange callback to update Props
```
ferrom-jaia: Nitpicking: Should we capitalize and add a space with the comment? Not sure just want it to be consistent.

ferrom-jaia (changes requested):
Looks good! Just a couple of things to double check on. Great description!

src/web/jcc/client/components/TaskSettingsPanel/__tests__/TaskSettingsPanel.test.tsx
JRoybot: Sure

src/web/jcc/client/components/TaskSettingsPanel/__tests__/TaskSettingsPanel.test.tsx
JRoybot: I'm not sure, will look into it more.  I am pretty sure the basic waitFor has a built in timeout because I have seen tests timeout while debugging them.

src/web/jcc/client/components/TaskSettingsPanel/__tests__/TaskSettingsPanel.test.tsx
JRoybot: Good idea.

I thought this block of code in TaskSettingsPanel:onChangeTaskType was suspect but was reluctant to change it because not sure why it was done and if changing it would create any side-effect issues.

```
        // Change to NONE
        if (newTaskType == TaskType.NONE) {
            props.onChange(undefined);
            return;
        }
```
Do you know why this was done this way instead of returning a task with type set to None?  If this can/should be changed I will modify the test to validate the task when "NONE" is selected on the widget.

ferrom-jaia: Shoot, I do not know off the top of my head why this is the case. I think Ed wrote this so maybe he can have some input @edsanville

JRoybot: Ed already said it was so long ago he does not know why.

edsanville: It doesn't look like my code, but I could be wrong.  I'll investigate this now...

src/web/jcc/client/components/TaskSettingsPanel/__tests__/TaskSettingsPanel.test.tsx
ferrom-jaia: You are right. The default timeout should be 1 second

src/web/jcc/client/components/TaskSettingsPanel/__tests__/TaskSettingsPanel.test.tsx
edsanville: I apparently did write this code, crazy!  Looking at it, I think I did it that way because the file was re-factored from a previous version from before the NONE task type was added, and I was trying to retain the previous behavior exactly.

edsanville: I'd say we should do some testing, and get rid of it if possible.

JRoybot: I'll change it and try to see if anything breaks using the sim...

Also, currently validateTask simply returns if undefined task is passed in.  I think this is acceptable because Props might be passed in to a component without a task defined.

Let me know if you disagree

src/web/tests/helpers/ValidateTask.tsx
JRoybot: This is good idea, will flag it if someone adds a TaskType without updating the test helper.

src/web/jcc/client/components/TaskSettingsPanel/__tests__/TaskSettingsPanel.test.tsx
edsanville: It does have a built-in timeout of 1 second, from my testing.

src/web/jcc/client/components/TaskSettingsPanel/__tests__/TaskSettingsPanel.test.tsx
JRoybot: > Consider testing invalid input scenarios, e.g., if the task-select-input-id is missing or an unexpected value is passed to the select.

If the task-select-input-id is missing or incorrect the test will fail on line 27 (been there, done that)
I don't see a reason to test selecting an invalid value in the select widget since that is impossible for a user to do.. Let me know if you disagree.

JRoybot: > I'd say we should do some testing, and get rid of it if possible.

I got rid of it and exercised everything in the sim and did not see any errors or odd behavior.  Updated test logic to work with new functionality.

src/web/tests/jest.setup.js
edsanville: If we let errors show through, then we can use "errors" during debug, too.  Which is nice.

src/web/tests/helpers/ValidateTask/__tests__/ValidateTask.test.tsx
```diff
@@ -0,0 +1,383 @@
+import { validateTask } from "../ValidateTask";
+import { MissionTask, TaskType } from "../../../../shared/JAIAProtobuf";
+
+type testParams = {
+    description: string;
+    task: MissionTask;
+};
+
+// Test Cases of Valid Mission Tasks
+const validTaskTestCases: testParams[] = [
```
JRoybot: I moved the test case definitions in a `test_cases.json ` file .  It should make managing the test cases easier and make the code look cleaner.

michael-jaia (commented):
Hey Jeff, I just did a read-through of the PR. Here are the things I noted. Let me know what you think.

**1.** validateTask nicely encapsulates the expected structure of each task and helps with the readability of the TaskSettingsPanel tests.

**2.** I like the valid test cases for the valididateTask function to confirm the structure with real data. However, I'm not sold on the invalid test cases. I can get behind the ones with missing data, but the test cases that cross parameters seem to add unnecessary complexity. (I know this is something @ferrom-jaia suggested).

**3.** Since validateTask is closely related to the testing of the TaskSettingsPanel what do you think of them residing in the same folder?

**4.** The testCases.json file name might be a little generic, I think mock-tasks.json might be stronger (open to alternatives).

**5.** Possible file structure based on the above notes:
* TaskSettingsPanel
    * \_\_tests\_\_
        * TaskSettingsPanel.test.tsx
        * validate-task.test.ts
        * validate-task.ts
        * mock-tasks.json

ferrom-jaia:
> **2.** I like the valid test cases for the valididateTask function to confirm the structure with real data. However, I'm not sold on the invalid test cases. I can get behind the ones with missing data, but the test cases that cross parameters seem to add unnecessary complexity. (I know this is something @ferrom-jaia suggested).

Hi Twomey,

Thank you for reviewing the pull request!

The reason for including the invalid test cases is to help future developers better understand how to write comprehensive test cases. While I agree it may seem overly complex for a piece of code we're not actively using, I believe it serves as a useful example of the expected structure and thought process when creating tests.

michael-jaia:
Sounds good Jeff. You sold me on the invalid test cases and the location of the `validateTask` function in `src/web/tests/helpers`. I think it would be nice to change the file extensions of `ValidateTask.tsx` and `ValidateTask.test.tsx` to `.ts` and `.test.ts` since they don't involve any React.

ferrom-jaia (approved):

edsanville (commented):
Now that we let `console.error` events through to the console, this has appeared in the CommandControl test:
``` console.error
    Warning: An update to CommandControl inside a test was not wrapped in act(...).
    
    When testing, code that causes React state updates should be wrapped into act(...):
    
    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */
    
    This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
        at CommandControl (/home/ed/projects/jaia/jaiabot/src/web/jcc/client/components/CommandControl/CommandControl.tsx:301:9)

      1100 |
      1101 |     hubConnectionError(errMsg: String) {
    > 1102 |         this.setState({ disconnectionMessage: "Connection Dropped To HUB" });
           |              ^
      1103 |         console.error(errMsg);
      1104 |     }
      1105 |

      at call (node_modules/react-dom/cjs/react-dom.development.js:86:30)
      at printWarning (node_modules/react-dom/cjs/react-dom.development.js:60:7)
      at error (node_modules/react-dom/cjs/react-dom.development.js:27628:9)
      at warnIfUpdatesNotWrappedWithActDEV (node_modules/react-dom/cjs/react-dom.development.js:25547:5)
      at Object.scheduleUpdateOnFiber [as enqueueSetState] (node_modules/react-dom/cjs/react-dom.development.js:17925:7)
      at CommandControl.enqueueSetState [as setState] (node_modules/react/cjs/react.development.js:354:16)
      at CommandControl.hubConnectionError (jcc/client/components/CommandControl/CommandControl.tsx:1102:14)
      at jcc/client/components/CommandControl/CommandControl.tsx:1026:26
```

Is anyone else seeing this error?  The test passes regardless... but it looks like it's trying to connect to the GeoTIFF server, failing, and then trying to bring up the "Disconnected from the Hub" error message.

edsanville:
Since this isn't directly related to this PR, and it doesn't cause the test to fail, I'm going to approve the PR and we can deal with this in another PR.

edsanville (approved):
Looks good... we'll address the warning in another PR related to the CommandControl testing.

PR #1015, opened 2024-11-13
Task/MUI-button-test-example/JAIA-1684

JRoybot:
Examples of how to test MUI Button Widgets.  Was able to use labels, titles and test ids to manipulate the 2 buttons on the Rally Points Panel.


edsanville (dismissed):

src/web/jcc/client/components/RallyPointPanel/__tests__/RallyPointPanel.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+    {
+        description: "Test Delete Button using title",
+        getMethod: () => screen.getByTitle("Delete Rally Point Button"),
+        callBackMethod: mockDeleteRallyPoint,
+    },
+];
+
+describe("Rally Point Panel Button Tests", () => {
+    beforeEach(() => {
+        jest.clearAllMocks(); // Ensure a clean state for each test
```
ferrom-jaia: Nitpicking: Can we move the comment above the line.

src/web/jcc/client/components/RallyPointPanel/__tests__/RallyPointPanel.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+        // Get the button using the provided getMethod
+        const buttonElement = getMethod();
+        expect(buttonElement).toBeInTheDocument();
+
+        // Click the button
+        await userEvent.click(buttonElement);
+
+        // Verify that the callback was called
+        await waitFor(() => {
```
ferrom-jaia, line 81: Should we add a timeout:

await waitFor(() => {
              expect(selectElement.value).toBe(value);
}, { timeout: 1000 });

src/web/jcc/client/components/RallyPointPanel/__tests__/RallyPointPanel.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+        callBackMethod: mockDeleteRallyPoint,
+    },
+    {
+        description: "Test Delete Button using title",
+        getMethod: () => screen.getByTitle("Delete Rally Point Button"),
+        callBackMethod: mockDeleteRallyPoint,
+    },
+];
+
+describe("Rally Point Panel Button Tests", () => {
```
ferrom-jaia: What do you think about this test name: RallyPointPanel: Button Interaction Tests

src/web/jcc/client/components/RallyPointPanel/__tests__/RallyPointPanel.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+        // Get the button using the provided getMethod
+        const buttonElement = getMethod();
+        expect(buttonElement).toBeInTheDocument();
+
+        // Click the button
+        await userEvent.click(buttonElement);
+
+        // Verify that the callback was called
+        await waitFor(() => {
+            expect(callBackMethod).toHaveBeenCalledTimes(1);
```
ferrom-jaia, line 82: Should we verify the callBackMethod is what we expect?

ferrom-jaia: // Optional: Verify callback arguments
        // expect(callBackMethod).toHaveBeenCalledWith(expectedArgument);

src/web/jcc/client/components/RallyPointPanel/__tests__/RallyPointPanel.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+    },
+];
+
+describe("Rally Point Panel Button Tests", () => {
+    beforeEach(() => {
+        jest.clearAllMocks(); // Ensure a clean state for each test
+    });
+    test.each(testCases)("$description", async ({ getMethod, callBackMethod }) => {
+        // Render the Rally Point Panel with mock props
+        render(<RallyPointPanel {...mockProps} />);
```
ferrom-jaia, line 71: Should we Test for edge cases? For example where mockProps.selectedRallyFeature might be undefined or have invalid data.

ferrom-jaia (commented):
Looks good! I had a couple of comments.

src/web/jcc/client/components/RallyPointPanel/__tests__/RallyPointPanel.test.tsx
JRoybot: > Should we verify the callBackMethod is what we expect?

I accomplished that by using different mock functions for each button and specifying the method to be called in the test case definition.

JRoybot: > // Optional: Verify callback arguments // expect(callBackMethod).toHaveBeenCalledWith(expectedArgument);

I think that is a good idea but think it should be different task than just creating an example of the button test.  If you want to make it part of this I will.

src/web/jcc/client/components/RallyPointPanel/__tests__/RallyPointPanel.test.tsx
JRoybot: waitfor has built in tiimeout

src/web/jcc/client/components/RallyPointPanel/__tests__/RallyPointPanel.test.tsx
JRoybot: that would be a good unit test for the panel functionality, if you want to include that in the button example task let me know

src/web/jcc/client/components/RallyPointPanel/__tests__/RallyPointPanel.test.tsx
michael-jaia: I'm a little confused here. Would we be verifying that the test case has the correct callback function? And/or that it is supplied the correct argument? In the case of supplying the right argument, can we add the line `expect(callBackMethod).toHaveBeenCalledWith(mockSelectedRallyFeature);` to the test since all the callbacks expect a `OlFeature<Point>`?

src/web/jcc/client/components/RallyPointPanel/__tests__/RallyPointPanel.test.tsx
michael-jaia: I think these are some of the test cases we would want for `goToRallyPoint` and `deleteRallyPoint`, right? In other words, these edge case tests may not belong in RallyPointPanel.test.tsx since the failure would happen in CommandControl.tsx in `goToRallyPoint` and `deleteRallyPoint`.

michael-jaia (commented):
Hey Jeff, I think `RallyPointPanel.test.tsx` sets a clean structure for others to follow. I'm wondering if we should hold off on using the test-id hook in the example since it may lead to other developers overusing it when there are opportunities to use more RTL preferred hooks such as aria-labels.

src/web/jcc/client/components/RallyPointPanel/__tests__/RallyPointPanel.test.tsx
JRoybot: Not sure I fully understand the question.  If we want to see if RallyPointPanel handles bad props, then it would be a unit test for this component.  If we want to see if parents produce the right props it would be a unit test for the parent.  Again, this was intended as a simple example, not a unit test.

src/web/jcc/client/components/RallyPointPanel/__tests__/RallyPointPanel.test.tsx
JRoybot: Yes I'll make that update.,  I only tested that the right callback function got called but will check the argument as well.

michael-jaia:
> I don't have a strong opinion either way. I wanted to show all the ways I got it to work. I could either delete those examples or add a comment stating test-id was only show as example and is still considered a last resort.
> 
> Just let me know, which you prefer.

I like the idea of adding a comment to keep with the format of showing a variety of hooks while also warning about using test-id as a last resort.

src/web/jcc/client/components/RallyPointPanel/__tests__/RallyPointPanel.test.tsx
JRoybot: > I'm a little confused here. Would we be verifying that the test case has the correct callback function? And/or that it is supplied the correct argument? 

The test verifies the button calls the callback specified in the test case.
This review should verify the test case is correct.

edsanville (approved):
Code looks good to me.  The tests also ran and passed just fine.

ferrom-jaia (approved):
Looks goo.d. Tested by running npm test in the src/web directory.

PR #1016, opened 2024-11-14
bug/depth-contours-appear-under-a-custom-overlay-jaia-1730

edsanville:
The GeoTIFF custom layers default to zIndex=42, so increase the depth contour layer's zIndex to 43.
The custom layer's zIndex is loaded from the hub's GeoTIFF server, though... so, if that's configured with a higher zIndex, the contour map will be obscured again.
We probably need some rules about zIndex ranges, or add zIndex configurability to the map layers settings.


JRoybot (approved):

michael-jaia:
I agree Ed, we need a better system for managing z-indexes. As for this specific bug fix, I tested it with two GeoTiffs. The first test I used a single GeoTiff. The second test I used two GeoTiffs at the same time (just to confirm that the z-index does not increment when more than one GeoTiff is selected at a time).
contours-visible-with-geotiff
two-geotiffs-depth-contours-visible

michael-jaia (commented):
I agree Ed, we need a better system for managing z-indexes. As for this specific bug fix, I tested it with two GeoTiffs. The first test I used a single GeoTiff. The second test I used two GeoTiffs at the same time (just to confirm that the z-index does not increment when more than one GeoTiff is selected at a time).

contours-visible-with-geotiff
two-geotiffs-depth-contours-visible

michael-jaia (approved):

PR #1017, opened 2024-11-15
task/create-accordion-test-example-jaia-1685

edsanville:
Example to test that an Accordion component opens and closes, and calls the onChange function appropriately.


JRoybot (dismissed):
Looks good.  Made a few small changes.

- Moved mock callback function out of test and put in props.  I think this is a better pattern to use.  Once we start to build more tests I suspect we will find it useful to move some mocks out of test files into utility files that can be used in multiple tests.
- Added verification the Accordion Summary was visible before clicking it.

src/web/jcc/client/components/__tests__/Accordion.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+        //Summary should be visible
+        expect(summary).toBeVisible();
+        // Accordion starts closed, so details invisible
+        expect(details).not.toBeVisible();
+
+        // Click the summary
+        await userEvent.click(summary);
+
+        // Wait for onChange logic to be triggered, which opens the details
+        waitFor(() => {
```
ferrom-jaia: Should we add await before this? The waitFor calls are not awaited, which could lead to unexpected tests if the state changes asynchronously

src/web/jcc/client/components/__tests__/Accordion.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+            expect(details).toBeVisible();
+        });
+
+        // Click summary again
+        await userEvent.click(summary);
+
+        // Wait for onChange logic to be triggered again, which closes the details
+        waitFor(() => {
```
ferrom-jaia: Should we add await before this? The waitFor calls are not awaited, which could lead to unexpected tests if the state changes asynchronously

src/web/jcc/client/components/__tests__/Accordion.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <Accordion expanded={false} onChange={props.onChange}>
+            <AccordionSummary>
+                <Typography>Accordion Title Here</Typography>
+            </AccordionSummary>
+            <AccordionDetails>Accordion Details Here</AccordionDetails>
+        </Accordion>
+    );
+}
+
+describe("Accordion Tests", () => {
```
ferrom-jaia: Nitpicking: What about this for a test name: Accordion expands and collapses on summary click

ferrom-jaia (changes requested):
Looks good. Just a couple of comments. I think we need to move this test into something like: jcc/client/components/test-examples/__tests__

src/web/jcc/client/components/__tests__/Accordion.test.tsx
edsanville: You're right!  But, in this case it turns out I don't even need waitFor, (the await userEvent.click does the required waiting).

There was another problem with the code that caused me to resort, incorrectly, to `waitFor`.

I need to use the `isExpanded` state variable to control the opening and closing of the Accordion, (before, I had assumed it was internal state to the widget, but it is not).

By using `waitFor` without `await`, I made the test "pass" by not actually completing the `waitFor` callback!  So yeah, we must always `await` the `waitFor` calls.

src/web/jcc/client/components/__tests__/Accordion.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+        // Click the summary
+        await userEvent.click(summary);
+
+        // Wait for onChange logic to be triggered, which opens the details
+        expect(details).toBeVisible();
+
+        // Click summary again
+        await userEvent.click(summary);
+
+        // Wait for onChange logic to be triggered again, which closes the details
```
ferrom-jaia: Is this comment still true?

src/web/jcc/client/components/__tests__/Accordion.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+        //Summary should be visible
+        expect(summary).toBeVisible();
+        // Accordion starts closed, so details invisible
+        expect(details).not.toBeVisible();
+
+        // Click the summary
+        await userEvent.click(summary);
+
+        // Wait for onChange logic to be triggered, which opens the details
```
ferrom-jaia: Is this comment still true?

michael-jaia (commented):
Hi Ed, this test has a nice structure and reads well. I'm thinking we might want to create an `example` folder in `src/web/tests` so it is easy to find for people new to the repo. What do you think?

edsanville:
> Hi Ed, this test has a nice structure and reads well. I'm thinking we might want to create an `example` folder in `src/web/tests` so it is easy to find for people new to the repo. What do you think?

I'm down with that.  I'll move it over.

ferrom-jaia (approved):
Looks good. Tested by running npm test in the src/web directory.

PR #1019, opened 2024-11-19
Task/test app for repeatable missions/jaia 1759

tsaubergine:
## Mission Repeater tool

First see https://github.com/jaiarobotics/jaiabot/blob/task/test-app-for-repeatable-missions/jaia-1759/src/doc/markdown/page61_mission_repeater.md

Tested in sim, may need some updates to the script (in `jaiabot/config/templates/bot/jaiabot_mission_repeater.pb.cfg.in`) to run the exact test mission you want.

https://github.com/user-attachments/assets/fa1a4343-6746-4ef7-85f5-72ce238f4215

### jaiabot_mission_manager

- Update mission_manager to publish MissionReport on each state entry as well as on each loop(), so that any subscribers to this don't miss state changes that happen quickly. **Needed to allow `jaiabot_mission_repeater` to know all state transitions reliably**
- Subscribe to `jaiabot::groups::self_command` Commands from within the bot. **Needed to allow `jaiabot_mission_repeater` to PAUSE and RESUME the mission**

### jaiabot_pid_control

- Add new DesiredCourse type (SETPOINT_SUSPEND_PID) that suspends publication of LowControl in `jaiabot_pid_control` until the next DesiredCourse is received. **Needed so `jaiabot_mission_repeater` could stop publication of zero-valued LowControl while running its script (and publishing LowControl).**


ferrom-jaia (approved):
Looks good! 

Tested in sim.

Nick tested in the field as well with fleet 1.

PR #1035, opened 2024-12-13
feature/enter-full-screen-on-user-click

ferrom-jaia:
## Description

The default behavior on some tablets do not enter full screen mode when running the jcc. This will enter full screen mode when a user clicks on the jcc while using the tablet. 

## Testing

1. This needs to be tested on a tablet


michael-jaia (commented):
Tested on tablet and laptop. On the first tap on the tablet the JCC switched to full screen. I did not notice any issues in subsequent taps. On the first click on a laptop, I did not enter full screen mode (as expected). Subsequent clicks also did not result in full screen (as expected). No errors in the browser console too.

One thing we might be able to clean up is merging the function jccClicked with jccContainerClicked. It looks like we can add the functionality for full screen mode to jccContainerClicked.

michael-jaia:
Re-tested on Fleet 0 Hub and tablet. Initial tap led to full screen. Also powered the tablet off and on, and the next tap put the JCC back in full screen.

PR #1036, opened 2024-12-13
task/implement-new-web-dir-structure-jcc

michael-jaia:
### Summary
Moves files in the `jcc` directory to match the new web directory structure set out in https://github.com/jaiarobotics/jaiabot/pull/867 and documented [here](https://github.com/jaiarobotics/jaiabot/blob/1.y/src/doc/markdown/page111_developing_with_react_context.md). I recommend checking out the branch to see the folder structure in your IDE.

The goal is to provide a clear structure to the web directory so we can deliver features and bug fixes with more ease. To get the ball rolling, the PR:
* Separates React code from the code that users do not directly interact with such as the processing of task packets, survey lane calculations, and the creation of layers.
* Extracts the React code in the `jcc` folder one level up so eventually components and containers can be reused by different apps such as the JDV and JED.
* Sets a structure for how folders should be constructed and files should be named to make it easier for developers to add new code.

### Testing
Sent 3 Bots on survey mission consisting of bottom dives.

image


src/web/containers/MissionSpeedSettings/MissionSpeedSettings.tsx
```diff
@@ -1,12 +1,12 @@
 import React from "react";
```
JRoybot: Should this directory and the file names be changed to MissionSettingsPanel to match the name of the component and to be more consistent with the other Panel naming conventions?

src/web/containers/MissionSpeedSettings/MissionSpeedSettings.tsx
```diff
@@ -1,12 +1,12 @@
...
```
JRoybot: Maybe move these files into MissionSettings above since they are only used there?

src/web/containers/SetHubLocation/SetHubLocation.tsx
```diff
@@ -6,14 +6,14 @@ import { Map } from "ol";
 import PointerInteraction from "ol/interaction/Pointer";
```
JRoybot: Should this be part of the SettingsPanel directory?

src/web/tests/helpers/validate-task/validate-task.ts
JRoybot, line 1: Given the new structure we might want to reconsider putting these helper files in the TaskSettingsPanel/__tests__ directory

JRoybot: We may want to move mocks into folders they are mocking as well

JRoybot (commented):
Looks good, let me know if you want me to do any restructuring of the test files.

src/web/containers/MissionSpeedSettings/MissionSpeedSettings.tsx
JRoybot: This first comment was intended for the MissionSettings.tsx and the MissionSettings folder

michael-jaia: Yes, good call, I am going to change that now.

src/web/containers/MissionSpeedSettings/MissionSpeedSettings.tsx
michael-jaia: Moving MissionSpeedSettings folder into MissionControllerPanel to so the folder structure matches the layout on the interface.

src/web/containers/SetHubLocation/SetHubLocation.tsx
michael-jaia: Moving SetHubLocation into SettingsPanel folder.

JRoybot (dismissed):

src/web/tests/helpers/validate-task/validate-task.ts
ferrom-jaia: Yes, this makes sense to me.

JRoybot (approved):

ferrom-jaia (approved):
Looks good! Tested sim. Tested jest test. Reviewed directory structure.

PR #1037, opened 2024-12-13
task/add-hub-key-first-boot

ferrom-jaia:
## Description

We needed another way to setup the ssh keys on the hub and the bots. Now when you create an image for the hub you can save a ssh key pair next to the preseed.

This also allows the user to add the public key for the hub on the bots in the preseed as well.

This will enable us to use the JCU to complete other initial tasks.

## Testing

* This needs to be tested on a hub and a bot


tsaubergine:
Moved this code to https://github.com/jaiarobotics/jaiabot/blob/task/rework-first-boot-as-cloud-init/jaia-1806/rootfs/nocloud-init/install-hub-ssh-keys.sh in https://github.com/jaiarobotics/jaiabot/pull/1053

PR #1038, opened 2024-12-17
Bug/jcc wpt tap to move/jaia 1763

JRoybot:
Corrects issues where state of Move Waypoint Toggle got out of sync with goalBeingEdited.
Removed use of local state variable in GoalSettingsPanel to control the Tap To Move toggle, replaced with values from props
Added method to toggle edit mode in CommandControl
Added guard around logic to select waypoint to prevent reset if current waypoint is clicked


src/web/jcc/client/components/CommandControl/CommandControl.tsx
```diff
@@ -4190,6 +4204,7 @@ export default class CommandControl extends React.Component {
                         runList={this.getRunList()}
                         runNumber={goalBeingEdited?.runNumber}
                         enableEcho={this.state.enableEcho}
+                        moveWptMode={this.state.goalBeingEdited.moveWptMode}
```
JRoybot: Since every memeber of state.goalBeingEdited should we replace the individual props for each member with passing in the entire goalBeingEdited?

ferrom-jaia: Let's leave it for now.

ferrom-jaia (approved):
Looks good! Tested in sim.

PR #1039, opened 2024-12-19
bug/deploy-script-not-setting-systemd-config-files-correctly/JAIA-1788

ferrom-jaia:
## Description
Fix: Ensure correct PATH is used for systemd service generation

- Added an explicit export of ${HOME}/jaiabot/${build_dir}/bin to PATH before invoking systemd-local.sh.
   - This is added during a hub deployment, but was left out for a bot
- This ensures systemd.py resolves binaries from the current build directory rather than default or fallback paths (e.g., /usr/bin).
- Fixes path mismatches in generated systemd configurations, aligning with the intended deployment structure.

## Tests
- Deployed to flat bot
- Confirmed the services start successfully
- Confirm a jaiabot systemd config file is setup correctly


JRoybot (approved):

marshalllele (approved):
Looks good!

edsanville (approved):
Looks good to me!

PR #1040, opened 2024-12-20
bug/jdv-graph-table-map-data-not-aligned-when-hover/JAIA-1787

ferrom-jaia:
## Description

We noticed that when hovering over data values in jdv they were off by one value. The plot would show the correct value, but the table and the map data would not correspond correctly. The reason for this is we converted utime into a date to be readable by the user. In doing so we had to convert the time from microseconds to seconds. I added custom data to the plot to preserve the original utimes so that when we propagated the current hover time it would be in microseconds. 

## Video Testing
Behavior Before the fix:
[JDV-hover-testing-bug-before.zip](https://github.com/user-attachments/files/18212652/JDV-hover-testing-bug-before.zip)
After applying the fix:
[JDV-hover-testing-bug-after.zip](https://github.com/user-attachments/files/18212661/JDV-hover-testing-bug-after.zip)


marshalllele (approved):
Looks good! Verified in sim using >1Hz data series to ensure each individual value updates.

JRoybot (approved):

edsanville (approved):
Code looks good!

PR #1041, opened 2024-12-20
Task/combine toggle components in jcc/jaia 1784

JRoybot:
Removed unused WptToggle
Removed redundant EditModeToggle
Simplified toggleEditMode and cleaned up declarations using it
Replace use of EditModeToggle with JaiaToggle


ferrom-jaia (approved):
Looks good! Tested in sim. Found a issue with the goal settings not appearing if you clicked on the same wpt after closing the goal settings panel. Made a small update to fix.

PR #1042, opened 2024-12-21
bug/gps-data-rates-tvp-sky/JAIA-1552

ferrom-jaia:
## Description

We found that we we getting duplicate lat and long values. This pull request cleansup gps spi pty script and removes duplicates.

1. Moved the request for nmea messages that was in the while loop into the connect_spi function. This request only needs to be made once.
2. Ensure we are grabbing all the nmea messages by increasing the readbyes to 1024. This can lead to fragmented messages, which are dropped.
3. Change the read for data to 5 Hz from 100 Hz to reduce overhead.
4. Add a check to validate nmea sentences.

## Test Bot 2 Fleet 0

image


edsanville (approved):
Code looks good to me!

PR #1044, opened 2024-12-27
Feature/create driver for bio camera jaia 1760

edsanville:
In the `config/launch/camera-test` directory, you'll find a bot.launch.

This launch file will setup a local serial pipe between `/tmp/bot1_camera_0` and `/tmp/bot1_camera_1`.

The `jaiabot_driver_camera` binary will connect to `/tmp/bot1_camera_0`, and send `CameraCommand` and receive `CameraResponse` messages over that serial port.

You can use the liaison interface on the bot to send `CameraCommand` messages over the `jaiabot::camera` Group, which will be relayed over the serial line.

You can run the `src/python/camera/test/test_camera_driver.py` script to observe the CameraCommand messages:

```test_camera_driver.py --device /tmp/bot1_camera_1 --slave```

This will give you a command-line interface where you can type the protobuf text representations for `CameraResponse` messages, and see them come through to the `jaiabot_driver_camera`.

In production, the `src/python/camera/camera_driver.py` will be run on the zero and connected to the real serial port, like so:

```camera_driver.py --device {PATH_TO_SERIAL_PORT}```

You can provide the `--simulate` parameter, if you don't have a camera installed.


edsanville:
Comment:  we should probably rename the `lora` directory, as it now has nothing to do with LoRa radios.  I suggest `serial`.  I didn't do that because I didn't want to mission creep.

PR #1046, opened 2025-01-06
Bug/docker sim ci builds fail on tags/jaia 1805

tsaubergine:
Move checkout to clone line so docker cache of the `git clone` doesn't break the `git checkout <tag>`.


ferrom-jaia (approved):
Looks good!

PR #1048, opened 2025-01-07
task/define-typescript-classes-to-represent-data-model

michael-jaia:
### Overview
The Jaia Command & Control application is written in **TypeScript** using **OpenLayers** and **React**. The foundation of data management will now rest in instances of **TypeScript classes** that are modeled after the real-world operations of the Jaia system. These objects will supply data to OpenLayers based on vehicle movements and user interactions. They will also save data from a user’s interaction with the map. The **React Contexts** (`BotContext`,`HubContext`, and `MissionContext`) will store a reference to the objects that apply to them. For instance, the `BotContext` will contain a map of `Bot` class instances. The `BotContext` and `HubContext` will poll the Bot and Hub objects twice per second to retrieve the latest data. (One second is the frequency Bots and Hubs send statuses, we poll at twice the speed to reduce potential data age issues). With the latest data held in Context and updated every one second, the data will be accessible to child components without passing it through multiple levels using props, and the components will update when the new data comes through.

jcc-data-model drawio

This pull request contains the first pass at implementing the data model. The changes to React and OpenLayers will come in subsequent steps. 

### Testing
1. Unit tests that cover creating, editing, and deleting a single mission
2. Unit tests that cover creating and deleting multiple missions
3. In the `src/web/data/tests` directory there is an `index.ts` file that can be run from the command line (`node index.ts`) while the server and simulator is running to watch the data model ingest the Bot and Hub status messages.


src/web/data/task_packets/task-packets-new.ts
```diff
@@ -0,0 +1,60 @@
+import { DivePacket, DriftPacket, TaskType } from "../../utils/protobuf-types";
+
+export default class TaskPacket {
```
JRoybot, line 3: Will this replace the existing definition of TaskData in src/web/data/task_packets/task-packets.ts?  Is there still work needed to make that transition?

src/web/data/missions/__tests__/mission.test.ts
```diff
@@ -0,0 +1,51 @@
+import Mission from "../mission";
+import { waypointA, waypointB, waypointC, waypointD } from "../../tests/__mocks__/waypoint-mock";
+
+describe("operator adding and deleting single waypoints", () => {
+    let mission = new Mission();
+
+    test("adds first waypoint", () => {
```
JRoybot: I would combine these sequential tests into a single test.  There is no guarantee that jest will run the tests in this order.

src/web/data/missions/__tests__/mission.test.ts
```diff
@@ -0,0 +1,51 @@
+import Mission from "../mission";
+import { waypointA, waypointB, waypointC, waypointD } from "../../tests/__mocks__/waypoint-mock";
+
+describe("operator adding and deleting single waypoints", () => {
```
JRoybot: suggest adding tests to verify the setters and getters for Lat, Lon, Task, & CanMove

src/web/data/missions/__tests__/missions.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    missionA,
+    missionB,
+    missionC,
+    missionD,
+    missionE,
+    missionF,
+} from "../../tests/__mocks__/mission-mock";
+
+describe("operator adding and deleting single missions", () => {
+    test("adds first mission", () => {
```
JRoybot: Same comment as above, combine into single test since there is no guarantee these will run in order

src/web/data/missions/__tests__/missions.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    });
+
+    test("reset missions singleton", () => {
+        // Keep as the last "test" in the suite
+        missions.deleteAllMissions();
+    });
+});
+
+describe("operator adding and deleting multiple missions at once", () => {
+    const missionSet1 = [missionA, missionB, missionC, missionD];
```
JRoybot, line 58: same comment on sequential tests

src/web/data/sensors/imu.ts
```diff
@@ -0,0 +1,40 @@
+export default class IMU {
+    private heading: number;
+    private pitch: number;
+    private roll: number;
```
JRoybot, line 4: does the imu output yaw?  If so may want to include it now in case it is ever needed

src/web/data/sensors/pressure.ts
```diff
@@ -0,0 +1,13 @@
+export default class Pressure {
+    private depth: number;
+
```
JRoybot, line 3: Does pressure always get converted to depth before use?  Is salinity also available?

src/web/data/task_packets/task-packets-new.ts
```diff
@@ -0,0 +1,60 @@
...
+    private botID: number;
+    private startTime: number;
+    private endTime: number;
+    private taskType: TaskType;
+    private divePacket: DivePacket;
+    private driftPacket: DriftPacket;
+
+    constructor() {}
+
+    getBotID() {
```
JRoybot: Since a TasikPacket is part of the Bot Data set, is botID needed here?  Are we risking this ID getting out of sync with the both the TaksPacket is associated with from a data structure point of view?

src/web/data/waypoints/waypoint.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export default class Waypoint {
+    private lat: number;
+    private lon: number;
+    private task: Task;
+    private canMove: boolean;
+
+    constructor() {}
+
+    getLat() {
```
JRoybot: Is there ever a use case where a lat or lon is updated alone?  If not I suggest combining to getPosition, setPosition(lat: number, lon: number).  Or go one step further and use a position type that included the lat and lon

src/web/package-lock.json
```diff
@@ -53,7 +53,7 @@
                 "@types/jquery": "^3.5.30",
                 "@types/jqueryui": "^1.12.23",
                 "@types/lodash.clonedeep": "^4.5.9",
-                "@types/node": "^22.6.0",
+                "@types/node": "^22.10.2",
```
JRoybot, line 56: should we update src/web/package.json.in to point to the newer version of node?  We are currently specifying ^22.6.0

src/web/tsconfig.json
```diff
@@ -13,7 +13,7 @@
         "target": "ES2021",
         "esModuleInterop": true,
         "lib": ["ES2021", "dom"],
-        "types": ["@testing-library/jest-dom", "@types/plotly.js"],
+        "types": ["@testing-library/jest-dom", "@types/plotly.js", "node"],
```
JRoybot, line 16: curious why this was needed...

src/web/utils/protobuf-types.ts
```diff
@@ -576,6 +576,7 @@ export enum MissionState {
     IN_MISSION__UNDERWAY__MOVEMENT__REMOTE_CONTROL__SETPOINT = "IN_MISSION__UNDERWAY__MOVEMENT__REMOTE_CONTROL__SETPOINT",
     IN_MISSION__UNDERWAY__MOVEMENT__REMOTE_CONTROL__STATION_KEEP = "IN_MISSION__UNDERWAY__MOVEMENT__REMOTE_CONTROL__STATION_KEEP",
     IN_MISSION__UNDERWAY__MOVEMENT__REMOTE_CONTROL__SURFACE_DRIFT = "IN_MISSION__UNDERWAY__MOVEMENT__REMOTE_CONTROL__SURFACE_DRIFT",
+    IN_MISSION__UNDERWAY__MOVEMENT__TRAIL = "IN_MISSION__UNDERWAY__MOVEMENT__TRAIL",
```
JRoybot, line 579: were these additions merged from another task?

JRoybot (commented):
Conducted a static review, will build branch and test before final approval

src/web/data/waypoints/waypoint.ts
```diff
@@ -0,0 +1,42 @@
+import Task from "../tasks/task";
+
+export default class Waypoint {
+    private lat: number;
+    private lon: number;
+    private task: Task;
+    private canMove: boolean;
```
JRoybot: If this is to control the tap to move function we may want to call this tapToMove or moveOnMap to avoid confusion canEdit toggle associated with missions

src/web/data/task_packets/task-packets-new.ts
michael-jaia: TaskPacket is just a class that matches the message format of what a Bot sends to a Hub after a task. We will need to evaluate the TaskData class when we examine the management of task packets on web side.

src/web/data/sensors/imu.ts
michael-jaia: Would yaw be the same as heading?

src/web/data/sensors/pressure.ts
michael-jaia: The sensor outputs it in millibars and then we convert that to meters on the Bot in the driver in `src/bin/drivers/blue_robotics_pressure_sensor`. On the web side we only see the depth because that is what is sent in the BotStatus message in `src/lib/messages/jaia_dccl.proto`

src/web/data/task_packets/task-packets-new.ts
michael-jaia: Yes, good call, it there does arise a need for it, we can add it back.

src/web/data/waypoints/waypoint.ts
michael-jaia: Switching to `position` of the existing type `GeographicCoordinate`.

src/web/data/sensors/imu.ts
JRoybot: Yes sorry

src/web/data/task_packets/task-packets-new.ts
JRoybot: OK, I was just curious about the file name "-new"

src/web/tsconfig.json
michael-jaia: I needed it to make TypeScript compile with the imports in `src/web/data/tests/index.ts`

src/web/utils/protobuf-types.ts
michael-jaia: They were added because they had been updated in `src/web/shared/JAIAProtobuf.ts` but not in `src/web/utils/protobuf-types`. We want to move away from the shared folder which is why we have `src/web/utils/protobuf-types`; the move will happen towards the end of the web refactor. I needed to add them because TypeScript caught the mismatch.

src/web/data/waypoints/waypoint.ts
michael-jaia: I'm hesitant to call it `tapToMove` because I think after the refactor we may want to switch to a mode where we can drag it around the map. What do you think of `canMoveOnMap`?

JRoybot: Got it.  canMoveOnMap works for me :-)

src/web/data/missions/__tests__/mission.test.ts
michael-jaia: I wasn't sure about this one when writing the tests. At the time, I did some searching to see what the norm is around testing setters and getters, and it seemed to me like people found it unnecessary to add unit tests for them. On past projects, would you include tests for the base getters and setters?

michael-jaia:
> have you decided to abandon the concept of a run? Currently a mission has a set of runs assigned to bots but the user can switch the assignment of a run from one bot to another. Are we going to get rid of that capability or will we need to keep track of those assignments somewhere else?

Yes the distinction between runs and missions always seemed blurry in code and with users of the system. By using the terminology mission and mission sets, I think we can achieve the same functionality with less confusion. Here are some examples that come to mind:
* "A Bot can perform a mission"
* "The Bots are running their missions"
* "We loaded in a mission set to send the Bots on a survey"

michael-jaia:
> We will need to scrub the user interface for instances of Run and the R1, R2 labels etc

Yes, this will happen through out the refactor.

michael-jaia:
> General comment on the update rate of the Hub and Bot Status, since those are being updated at 1 hz and this code is running asynchronously from the Hub and Bot code we could see data ages of close to a second. It is typical to sample periodic data from an asynchronous source at least twice as fast as it is produced to reduce potential data age issues.

That is a good point, I will make that change in https://github.com/jaiarobotics/jaiabot/pull/1049

src/web/data/missions/__tests__/mission.test.ts
JRoybot: I agree it is pretty nit-picky.  Remember I have worked mostly on defense programs with strict testing requirements.  So yes we tested them all but it is debatable how much value they provided.  It is unlikely someone could screw up a setter or getter and it doesn't get caught by peer review but it happens.

JRoybot (approved):

src/web/data/missions/__tests__/mission.test.ts
michael-jaia: Ok let's hold off on the getter and setter unit testing for now. If we have time at the end, we can add it, but I think it is best if we plow ahead the bigger ticket items first.

PR #1049, opened 2025-01-07
task/use-react-context-to-handle-bot-status-data

michael-jaia:
### Overview
The Jaia Command and Control application will use React Context to mange the state of the application. Right now we mange state in a top level component, `CommandControl`, but this has led to an unwieldy file and confusion tracing the path of props. Going forward, we will utilize a `BotContext`, `HubContext`, and `MissionContext`. On a high level, a child component “using” `ContextA` will dispatch an event which will be handled by `ReducerA`. `ReducerA` will consist of switch-case statements to update the state of `ContextA`. When `ContextA` is updated, its consumers (child components) will be re-rendered with the most up to date data. This structure will remove layers of prop passing and relieve the `CommandControl` component of its state management duties for the entire application. Now with Context, we can manage state with tools designed for the job and divide the data into groups that reflect the structure of the physical Jaia system.

`BotContext` will store a reference to the map of `Bot` objects that are part of the data model. Every one second, the `BotContext` will dispatch an event that saves the latest data from the Bots in a new copy of the `BotContext` state. This will be returned by the reducer causing a re-render of the child components with the latest data. We need to make a copy of the `BotContext` state to signal to React that data changed. Without making the copy, React will not re-render because it will see the `bots` property in state is pointing to the same location in memory. By making a copy of state in the reducer (`let mutableState = { ...state };`), React detects a change in state since the previous state and the new state do not share the same location. We do this process twice per second reduce potential data age issues with the Bots and Hubs producing status messages every 1 second..

`HubContext` follows the same structure as `BotContext` to store a map of Hub objects that are part of the data model.

### Built On
1. https://github.com/jaiarobotics/jaiabot/pull/1048

### Testing
1. Added `console.log` statements to `handleBotsPolled` function to view the data coming from the Bots in simulation.
2. Opened the Hub Details panel to make sure the data updated and the links opened to the expected IP addresses.


JRoybot (dismissed):
Reviewed code, built branch and tested in Sim

JRoybot (dismissed):

michael-jaia:
### Testing (*after resolving merge conflicts*)
* Added console.log statements to handleBotsPolled function to view the data coming from the Bots in simulation.
* Opened the Hub Details panel to make sure the data updated and the links opened to the expected IP addresses.

JRoybot (approved):
Looks great

PR #1050, opened 2025-01-07
task/use-react-context-to-store-mission-data

michael-jaia:
### Overview
`MissionContext` will store a reference to the map of `Mission` objects that are part of the data model. In addition, it will be the bridge between interactions with OpenLayers and the data model. When a user interacts with the map while mission planning, the click events will be captured by the `on` method of our singleton instance of the OpenLayers' `Map` class. Since the Map instance will be accessible inside React Components, the `on` method can be called so it can dispatch events to the `MissionContext`. With the event handled in `MissionContext` it can call functions that 1) update the OpenLayers `Map` instance (e.g. calling the `addFeature` method) and 2) update the data model. With the data model updated, we can bring the `MissionContext` state up to date by setting a copy equal to `missions.getMissions()` which will return a map of the `Mission` instances in the data model and re-render child components. This will put OpenLayers, React, and the data model in a synchronized state.

This pull requests establishes the base structure of `MissionContext`. It does not bridge OpenLayers and the data model yet.

### Built On
1. https://github.com/jaiarobotics/jaiabot/pull/1048
2. https://github.com/jaiarobotics/jaiabot/pull/1049


src/web/context/Mission/MissionContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+interface MissionContextType {
+    missions: Map<number, Mission>;
+}
+
+interface Action {
+    type: string;
+    missions?: Map<number, Mission>;
+}
+
+interface MissionContextProviderProps {
```
JRoybot, line 17: Why are we referring to these as Props?  I think it may create confusion with the typical Props passed down through the React call change.  Looks like this is to keep track of React Nodes that consume the MissionContext.  Perhaps something like MissionContextClients ??

src/web/context/Mission/MissionContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { MissionActions } from "./mission-action";
+import { missions } from "../../data/missions/missions";
+import Mission from "../../data/missions/mission";
+
+interface MissionContextType {
+    missions: Map<number, Mission>;
+}
+
+interface Action {
+    type: string;
```
JRoybot: Do you envision using strings for all the actions or is this just a place holder for TBD actions?

JRoybot: I suspect we will use something that includes an enumerated set of actions and perhaps a set of optional data fields.  Is that in line with your plans?

src/web/context/Mission/MissionContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @param {MissionContextType} state Holds the most recent reference to state
+ * @param {Action} action Contains data associated with a state update
+ * @returns {MissionContextType} The updated state object
+ */
+function missionReducer(state: MissionContextType, action: Action) {
+    let mutableState = { ...state };
+    switch (action.type) {
+        default:
+            return state;
```
JRoybot, line 35: Just a stub, for now...

src/web/context/Mission/mission-action.ts
```diff
@@ -0,0 +1 @@
+export enum MissionActions {}
```
JRoybot, line 1: Obviously a place holder... it is imported in MissionContext.tsx but not used.  Will this replace the Action defined on line 12 above?

src/web/jcc/App.tsx
```diff
@@ -13,7 +14,9 @@ export default function App() {
             <GlobalContextProvider>
                 <HubContextProvider>
                     <BotContextProvider>
-                        <CommandControlWrapper />
+                        <MissionContextProvider>
```
JRoybot, line 17: Will this pass in the "children" once they are ready to use the reducer?

src/web/context/Mission/MissionContext.tsx
michael-jaia: Yes that should be `type: MissionActions`

src/web/context/Mission/mission-action.ts
michael-jaia: Yes

src/web/jcc/App.tsx
michael-jaia: `<CommandControlWrapper />` will be passed as `children` to `MissionContextProvider`. Here is a better explanation: https://react.dev/learn/passing-props-to-a-component#passing-jsx-as-children

src/web/context/Mission/MissionContext.tsx
michael-jaia: They really are Props because `MissionContextProvider` is a function component that accepts Props, specifically it is looking for the `children` property in the Props object (`props.children`) but we use destructuring (`{ children }`) so we do not need to access `children` with `props.children` instead we can just use `children`. Here is an example from the docs: https://react.dev/learn/passing-props-to-a-component#passing-jsx-as-children (I also posted this link in one of the comments below).

JRoybot (approved):

PR #1051, opened 2025-01-09
Debconf queries with reasonable defaults should be medium priority

tsaubergine:
I'm getting prompted to set these debconf values on apt upgrade of jaiabot-embedded, where I'm guessing since defaults are set in jaiabot-embedded.templates you'd rather these defaults be used automatically.

To ensure that existing systems don't require the user to manually set the debconf value (either via the command line or the command-line UI - neither of which are supported by JCU), the priority needs to be medium or lower, not high:


ferrom-jaia (approved):
Looks good.

PR #1057, opened 2025-01-21
task/implement-design-for-open-layers-structure/JAIA-1824

michael-jaia:
## Overview
The core components of the OpenLayers map live in `src/web/openlayers`. They are separated into folders and imported into the `src/web/openlayers/maps/map.ts` file that instantiates a singleton instance of the OpenLayers `Map` class. We place the map components into separate folders to avoid circular dependencies that can occur from constructing the map and needing to access its data in different areas of the code.

```
// File Structure
openlayers/
├── controls
├── features
├── layers
├── maps
└── views
```

### Map
The singleton instance of the OpenLayers `Map` class is instantiated with base configurations in `src/web/openlayers/map.ts`. We then create a React Component `Map` which returns `<div id="map></div>"`. When the component first renders, we set the `target` of the map to `"map"`, so OpenLayers will render the map inside of the `div` element. It is important to wrap the location of the OpenLayers map inside this React Component becasue it allows us to dispatch actions through the React tree when users interact with the map.

### Layers
Layers will be created in `src/web/openlayers/layers`. The directory is divided into tile and vector layers. Vector layers inherit from the `JaiaVectorLayer` base class to maintain a consistent structure among the vector layers.

Also inside `src/web/openlayers/layers` is `jaia-layers.ts`. This file exports a singleton of type `JaiaLayers`. This class defines methods for adding and controlling the visibility of OpenLayers layers used in the application.

#### Creating a New Vector Layer
1. Create a new file in `src/web/openlayers/layers/vector` with the naming convention `{layer-name}-layer.ts`.
2. Define a class `LayerName` that extends `JaiaVectorLayer`.
3. Add the layer name to the enum `LayerTitles` in `src/web/types/openlayer-types.ts`.
4. Pass the layer title in the `super()` call such as `super(LayerTitles.BOT_LAYER);`
5. Override the base layer functions as desired.
6. Export a singleton instance of the class in the form `export const layerName = new LayerName();`

### Sources
Sources will be created in the instantiation of OpenLayers `Layers`. Source is the level that contains Features (icons we want on the map).

### Features
Features will be created within the `src/web/openlayers/features` folder. A feature file such as `bot-feature.ts` will contain functions to create the OpenLayers features which can be added to a layer. Expect these files to include functions such as `generateFeature` and `generateStyle`.

#### Adding and Deleting Features
* *Click Events*
    * The event will be first captured in the `map.on("click", callback)` statement in our `Map` React Component.
    * The `callback` will be a function containing a switch-case statement to determine what secondary click handlers need to be called based on the location of the click (*ex:* is this a click on the map or an existing waypoint, etc) and the state of the application (*ex:* is edit mode enabled).
    * The secondary click handlers will be the location where the data model is updated, OpenLayers is updated, and an action is dispatched to the one of the Contexts. 
    * To add or remove the Feature, we would import the associated layer singleton, get the appropiate layer and source, and call the methods defined in the class that update the layer.

* *Data Poll From Server*
    * When new data comes from the server to be rendered on the map such as task packets, it will **1)** be added to the data model and **2)** be added to the correct layers. We do not need to utilzie Context here because the new data such as the task packets will be added to `Bot` instances that get updated in the `BotContext` every 0.5 seconds. There will be a call to update the applicable layers when new data arrives in `index.tsx`.

### Automated Tests
* `src/web/openlayers/layers/__tests __` includes tests for:
    * Adding features to the map
    * Removing features from the map
    * Toggling layer visibility
* `src/web/components/Map/__tests__` includes tests for:
    * Clicking on Bot and Hub features on the map

## JIRA
Story Link: https://jaia-innovation.atlassian.net/browse/JAIA-1812


src/web/components/Map/Map.tsx
```diff
@@ -0,0 +1,87 @@
...
+    const globalDispatch = useContext(GlobalDispatchContext);
+
+    useEffect(() => {
+        map.setTarget("map");
+        map.on("click", (event: MapBrowserEvent<UIEvent>) => {
+            handleMapClick(event);
+        });
+    });
+
+    const handleMapClick = (event: MapBrowserEvent<UIEvent>) => {
```
JRoybot, line 27: Will this replace `clickEvent(evt: MapBrowserEvent<UIEvent>)` in `CommandControl` ?  
Will we add handling of other feature types here as well?

src/web/components/Map/__tests__/Map.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+};
+
+const hubStatusMock1: PortalHubStatus = {
+    hub_id: 1,
+    portalStatusAge: 1,
+};
+
+test("Select and deselect Bot and Hub icons on map", () => {
+    // Resolve click to being on a Feature of type MapFeatureTypes.BOT
+    mapModule.map.forEachFeatureAtPixel = jest.fn().mockReturnValue(botFeatureMock);
```
JRoybot, line 43: Nice way of testing this

src/web/context/Global/GlobalContext.tsx
```diff
@@ -207,18 +208,16 @@ function handleClickedBotTab(mutableState: GlobalContextType) {
  * @returns {GlobalContextType} Updated mutable state object
  */
 function handleClickedHubMapIcon(mutableState: GlobalContextType, hubID: number) {
```
JRoybot: This should probably wait until the BotDetails branch is merged as it will have a lot of updates to this file that will need to be merged manually

src/web/data/bots/bot.ts
```diff
@@ -14,6 +20,7 @@ export default class Bot {
     private missionStatus: MissionStatus;
     private botSensors: BotSensors;
     private taskPackets: TaskPacket[];
+    private location: GeographicCoordinate;
```
JRoybot, line 22: will mission go away when the mission manager is merged?

src/web/data/bots/bots.ts
```diff
@@ -4,9 +4,11 @@ import Bot from "./bot";
 
 class Bots {
     private bots: Map<number, Bot>;
+    private selectedBotID: number;
```
JRoybot: why is this needed?  Shouldn't we use `SelectedPodElement` from `GlobalContext` ?

src/web/data/hubs/hubs.ts
```diff
@@ -3,9 +3,11 @@ import Hub from "./hub";
 
 class Hubs {
     private hubs: Map<number, Hub>;
+    private selectedHubID: number;
```
JRoybot: Shouldn't we use SelectedPodElement from GlobalContext ?

src/web/jcc/App.tsx
```diff
@@ -7,15 +7,20 @@ import { HubContextProvider } from "../context/Hub/HubContext";
...
 import { CommandControlWrapper } from "../containers/CommandControl/CommandControl";
+import Map from "../components/Map/Map";
+
+// Style
+import "./App.less";
 
 export default function App() {
     return (
-        <div>
+        <div id="app">
```
JRoybot: why not name it "jcc" ?

src/web/jcc/App.tsx
```diff
@@ -7,15 +7,20 @@ import { HubContextProvider } from "../context/Hub/HubContext";
...
             <GlobalContextProvider>
                 <HubContextProvider>
                     <BotContextProvider>
                         <MissionContextProvider>
-                            <CommandControlWrapper />
+                            {/* <CommandControlWrapper /> */}
```
JRoybot: Why is this commented out?

src/web/openlayers/features/bot-feature.ts
```diff
@@ -0,0 +1,74 @@
...
+// Style
+const botIcon = require("../../style/icons/bot.svg");
+
+const TEXT_OFFSET_RADIUS = 11;
+
+export function generateBotFeature(botID: number) {
+    const bot = bots.getBot(botID);
+
+    if (!bot) {
+        return new Feature();
```
JRoybot, line 29: what does the calling function do with an empty Feature?

src/web/openlayers/features/bot-feature.ts
```diff
@@ -0,0 +1,74 @@
...
+            }),
+            rotateWithView: true,
+            offsetX: -TEXT_OFFSET_RADIUS * angleToXY(heading).x,
+            offsetY: -TEXT_OFFSET_RADIUS * angleToXY(heading).y,
+        }),
+    });
+}
+
+function getBotIconColor(bot: Bot) {
+    if (bots.getSelectedBotID() === bot.getBotID()) {
```
JRoybot: Shouldn't we use SelectedPodElement from GlobalContext ?

src/web/openlayers/features/hub-feature.ts
```diff
@@ -0,0 +1,67 @@
...
+import Hub from "../../data/hubs/hub";
+import { hubs } from "../../data/hubs/hubs";
+import { view } from "../views/view";
+import { MapIconColors } from "../../utils/style";
+import { MapFeatureTypes } from "../../types/openlayers-types";
+
+// Style
+const hubIcon = require("../../style/icons/hub.svg");
+
+const TEXT_OFFSET_RADIUS = 11;
```
JRoybot, line 20: Do we have a common place we can put common constants like this?  Used in bot-feature as well

src/web/openlayers/features/hub-feature.ts
```diff
@@ -0,0 +1,67 @@
...
+                color: "black",
+            }),
+            offsetX: 0,
+            offsetY: TEXT_OFFSET_RADIUS,
+        }),
+    });
+}
+
+function getHubIconColor(hub: Hub) {
+    if (hubs.getSelectedHubID() === hub.getHubID()) {
```
JRoybot: Shouldn't we use SelectedPodElement from GlobalContext ?

JRoybot (commented):
Nice Job on this.  Left some comments, mostly questions.

src/web/components/Map/Map.tsx
michael-jaia: Yes

src/web/data/bots/bot.ts
michael-jaia: Yes

src/web/jcc/App.tsx
michael-jaia: That is a testing artifact. I was testing with clean slate. I will revert this.

src/web/jcc/App.tsx
michael-jaia: Good call, I will switch this

src/web/openlayers/features/hub-feature.ts
michael-jaia: We do not, adding this to the list of tickets that need to be made.

src/web/openlayers/features/bot-feature.ts
michael-jaia: I thought it could be good to return an empty feature to keep the return type consistent. The calling function "renders" the feature, but it will not appear on the map since it does not have a location.

michael-jaia:
In response to the use of `selectedNode` from `GlobalContext`, we cannot access that object outside of a React component. This may mean that we want to create a `Global` class with a singleton `global` in the data model that bridges the gap between some of those UI states that we need outside of React. Its properties should be a subset of the `GlobalContextType` that is updated in the `GlobalContext.tsx` file.

JRoybot (dismissed):

michael-jaia:
## Problem
It seems logical for the `GlobalContext` to know the `selectedNode` because the `selectedNode` impacts what panels are rendered and what styles components receive. The catch is we need OpenLayers to know about the `selectedNode` to correctly style Features. This means we cannot soley store the `selectedNode` data in `GlobalContext` because it cannot be accessed by OpenLayers.

## Solution
1. Create a `JaiaGlobal` class in the data model
2. Export a singleton instance of the `JaiaGlobal` class named `jaiaGlobal`
3. The `JaiaGlobal` class should have a property `selectedNode` which is of type `{type: NodeType, id: number}`
4. When a user clicks a Bot icon on the map, the following calls would be made:
    ```
    // Update data model
    jaiaGlobal.setSelectedNode({ type: NodeType.BOT, id: botID });
    
    // Update OpenLayers
    botLayer.updateFeatures();
    hubLayer.updateFeatures();
    
    // Update Context
    globalDispatch({ type: GlobalActions.CLICKED_NODE });
    ```
    **First**, we update the data model because this is the backbone of our application. It is our truth, and it is used by our OpenLayers and React code. Inside setSelectedNode, the select/deselect decision will be made based on the current selection.
    **Second**, we are free to update OpenLayers since the data model has already been updated.
    **Third**, we dispatch an event to the `GlobalContext`. We do not need to pass the node type or node ID since `GlobalContext` will get that data from the data model (the `JaiaGlobal` singleton).

JRoybot (approved):
Updates all look good

PR #1058, opened 2025-01-22
task/combine-contexts-for-simplicity/JAIA-1831

michael-jaia:
## Overview
After working within the data model and Context structure, it became apparent that there is a better way to manage the data in Context. By combing `BotContext`, `HubContext`, and `MissionContext` into a singular `JaiaSystemContext`, we reduce two polls into one (since we no longer need to poll from `BotContext` and `HubContext`), and we make synchronization between the data model and Context possible with one call. Now, when a change occurs to the data model, `JaiaSystemContext` can be synchronized by dispatching the action SYNC_REQUESTED. This will prevent multiple re-renders from occurring if for example both the `BotContext` and the `MissionContext` needed to be synced together.

## Testing
Ensured the Hub details panel displayed the Hub status data as expected since this is the only Component utilizing either the `BotContext`, `HubContext`, or `MissionContext`.


JRoybot (approved):
Nice Job on this, didn't see anything out of place.  Would like to see this merged so I can update the BotDetails branch with the changes from here.

michael-jaia:
> Can you provide an example of dispatching SYNC_REQUESTED and provide any guidance on best use

Here is an example from the Missions panel that I am working on. Deleting a mission is an interaction that can affect a Bot and a mission. Instead of updating the `bots` and `missions` property in `JaiaSystemContext` separately, a dispatch of type `JaiaSystemActions.SYNC_REQUESTED` gets the latest data from the data model for Bots, Hubs, and missions and saves it into `JaiaSystemContext`. by calling the high level getters: `bots.getBots()`, `hubs.getHubs()`, and `missions.getMissions()`.

Best practice for handling is updates is to make changes in the following order:
`Data Model --> OpenLayers --> React Context`

The action `SYNC_REQUESTED` I think encompasses the idea of bringing Context in sync with the data model while maintaining the structure of past tense actions used in the React documentation for reducers.

```    
    const handleDeleteAllMissionsClick = () => {
    // Update data model
    missions.deleteAllMissions();
    missionsManager.clear();

    // Update OpenLayers

    // Update JaiaSystemContext
    jaiaSystemDispatch({ type: JaiaSystemActions.SYNC_REQUESTED });
    };
```

PR #1062, opened 2025-01-27
task/refactor-bot-details/JAIA-1807 

JRoybot:
Refactored the BotDetails Panel to use Global Context and the Data Model and to clean things up as much as possible.

- Renamed file BotDetails.tsx
- Renamed component BotDetails
- Moved Bot commands to src/web/types/commands.ts
- Split functions that send commands into an issue/send model (similar to HubDetails) to isolate API logic from React code
- Moved non-react code to bot-details.ts
- Created BotDetails test skeleton, no functioning test yet
- Replaced all references to the bot and hub data from props with data from the Global Context
- Replaced as many references to mission data from props as possible
   - Still need to implement missions fully in the data model
   - Some mission related functions like Toggle Edit Mode and Run Mission not working until that is done
- Added repeat_index to MissionStatus in data model
- Added missing TrailParameters to src/web/utils/protobuf-types.ts
   - replaced use of src/web/shared/JAIAProtobuf with src/web/utils/protobuf-types.ts in opened files
- Changed DownloadQueue to a queue of Bot IDs to simplify use
- Used GlobalActions to handle closing of panel and toggling accordion states
- Simplified GlobalActions and GlobalContext handling of selecting the hub or bots (now referred to as nodes)
- Nested support functions into BotDetails and moved non-React code into bot-details.ts
- Moved as much inline logic for onClick and onChange properties into handler functions as possible at this time
- Modified jestconfig to allow non test files in __tests__ without dummy tests
   - TaskSettingsPanel contains dummy test in file that can be removed

Known Follow up Work

- Once Mission Management has been added to GlobalContext, remove mission and run data from Props and refactor logic
-  TakeControl functions and logic still needs rework, should use GlobalContext
- props.deleteSingleMission to be removed once mission management is incorporated into data model
- RC Mode needs to be reworked and incorporated into the data model
- Download Queue should be reworked globally
- Download Processing should be reworked globally
- Logic involved with disabling buttons and button controls needs rework, mostly left alone
-


michael-jaia (changes requested):
Jeff, the function comments are a good start. Can you get them to follow the structure here: https://github.com/jaiarobotics/jaiabot/blob/1.y/src/doc/markdown/page110_coding_standards.md

src/web/containers/BotDetails/bot-details.ts
```diff
@@ -0,0 +1,303 @@
...
+ *
+ * @returns {string}
+ */
+export function getDistToHub(bot: Bot, hub: Hub) {
+    const botGPS: GPS = bot.getBotSensors().getGPS();
+    const hubGPS: GPS = hub.getHubSensors().getGPS();
+
+    let distToHub = "N/A";
+    if (botGPS && hubGPS) {
+        const botloc = turf.point([botGPS.getLat(), botGPS.getLon()]);
```
michael-jaia: NIT: `botLocation` and `hubLocation`

src/web/containers/BotDetails/bot-details.ts
```diff
@@ -0,0 +1,303 @@
...
+/**
+ * Provides message for clicking on map
+ *
+ * @returns {string}
+ */
+// TODO Edit mode toggle and related items not functional
+// until more functionality is add to the mission management
+export function getClickOnMapString(mission: Mission) {
+    let editString = "";
+    if (mission?.getCanEdit()) {
```
michael-jaia: We should be able to remove the `?`

src/web/containers/BotDetails/bot-details.ts
```diff
@@ -0,0 +1,303 @@
...
+    }
+    return distToHub;
+}
+
+/**
+ * Provides message for clicking on map
+ *
+ * @returns {string}
+ */
+// TODO Edit mode toggle and related items not functional
```
michael-jaia: For uniformity with the function comments, lets pop the TODO into the @notes section

src/web/containers/BotDetails/bot-details.ts
```diff
@@ -0,0 +1,303 @@
...
+// until more functionality is add to the mission management
+export function getClickOnMapString(mission: Mission) {
```
michael-jaia: Maybe `getHelperText`? I think `getClickOnMapString` could be a little confusing.

src/web/containers/BotDetails/bot-details.ts
```diff
@@ -0,0 +1,303 @@
...
+    }
+    return editString;
+}
+
+/**
+ * Provides off load percentage
+ *
+ * @returns {string}
+ */
+export function getBotOffloadPctString(botID: number, hub: Hub) {
```
michael-jaia: I think we should call this `getBotOffloadPercent`

src/web/containers/BotDetails/bot-details.ts
```diff
@@ -0,0 +1,303 @@
...
+    return botOffloadPercentage;
+}
+
+/**
+ * Provides repeat status message
+ *
+ * @returns {string}
+ */
+export function getRepeatNumberString(mission: Mission, missionStatus: MissionStatus) {
+    var repeatNumberString = "N/A";
```
michael-jaia: Switch to `let`

src/web/containers/BotDetails/bot-details.ts
```diff
@@ -0,0 +1,303 @@
...
+    }
+    return botOffloadPercentage;
+}
+
+/**
+ * Provides repeat status message
+ *
+ * @returns {string}
+ */
+export function getRepeatNumberString(mission: Mission, missionStatus: MissionStatus) {
```
michael-jaia: Similar here, I'm thinking `getRepeat` or `getMissionRepeat` might be crisper. Or maybe `formatRepeatProgress`

src/web/containers/BotDetails/bot-details.ts
```diff
@@ -0,0 +1,303 @@
...
+    var repeatNumberString = "N/A";
+    if (missionStatus?.repeat_index != null) {
```
michael-jaia: I am finding this logic a little confusing. If `missionStatus.repeat_index` is not null, would `mission.getRepeats()` ever be null?

src/web/containers/BotDetails/bot-details.ts
```diff
@@ -0,0 +1,303 @@
...
+    return botLogging;
+}
+
+/**
+ * Provides Active wapoint name and distance
+ *
+ * @returns {string , string}
+ */
+
+export function getActiveWptStrings(missionStatus: MissionStatus) {
```
michael-jaia, line 141: Should this be broken down into two different functions?

src/web/containers/BotDetails/bot-details.ts
```diff
@@ -0,0 +1,303 @@
...
+        if (mission.runIdInEditMode === run.id) {
+            mission.runIdInEditMode = "";
+        }
+        return run.command;
+    } else {
+        return null;
+    }
+}
+
+interface DisableInfo {
```
michael-jaia: Let's move the interface to the top of the file

src/web/containers/BotDetails/bot-details.ts
```diff
@@ -0,0 +1,303 @@
...
+// Utility Imports
+import * as turf from "@turf/turf";
+
+/**
+ * Provides class name and status age
+ *
+ * @returns { string , string }
+ */
+
+export function getStatusAge(bot: Bot) {
```
michael-jaia: I think for this one we should just return the class name and get the status age from `bot.getStatusAge()` in other parts of the code

michael-jaia (changes requested):

src/web/containers/BotDetails/bot-details.ts
JRoybot: I think we need it.  The details panel can be  displayed when a bot has no mission.

src/web/containers/BotDetails/bot-details.ts
JRoybot: It shouldn't be but since they are managed separately I figured it would be a good check just in case something was broken.  I will simplify it

src/web/containers/HubDetails/HubDetails.tsx
```diff
@@ -1,7 +1,12 @@
 // React -- Jaia
-import React, { useContext, useEffect, useState } from "react";
-import { GlobalContext, GlobalDispatchContext } from "../../context/Global/GlobalContext";
+import { useContext, useEffect, useState } from "react";
 import { GlobalActions } from "../../context/Global/GlobalActions";
+import {
+    GlobalContext,
+    GlobalDispatchContext,
+    HubAccordionNames,
+    PodElement,
```
michael-jaia: Should this be `NodeType`?

src/web/context/Global/GlobalContext.tsx
```diff
@@ -9,8 +9,9 @@ export interface GlobalContextType {
     clientID: string;
     controllingClientID: string;
     selectedPodElement: SelectedPodElement;
-    showHubDetails: boolean;
+    shownDetails: PodElement;
```
michael-jaia: What do you think about `visibleDetails`?

src/web/context/Global/GlobalContext.tsx
```diff
@@ -9,8 +9,9 @@ export interface GlobalContextType {
...
     hubAccordionStates: HubAccordionStates;
+    botAccordionStates: BotAccordionStates;
     isRCMode: boolean;
 }
```
michael-jaia, line 17: Can `selectedPodElement` be removed?

src/web/context/Global/GlobalContext.tsx
```diff
@@ -19,40 +20,84 @@ export interface SelectedPodElement {
...
+    imu: false,
+    sensor: false,
+};
+
 export interface GlobalAction {
     type: GlobalActions;
     clientID?: string;
-    hubID?: number;
-    hubAccordionName?: string;
+    elementType?: PodElement;
```
michael-jaia: Should these be switched to `nodeType` and `nodeID`?

src/web/context/Global/GlobalContext.tsx
```diff
@@ -19,40 +20,84 @@ export interface SelectedPodElement {
...
+    elementID?: number;
+    hubAccordionName?: HubAccordionNames;
+    botAccordionName?: BotAccordionNames;
 }
 
 interface GlobalContextProviderProps {
     children: ReactNode;
 }
 
+// TODO: NodeType
```
michael-jaia: Switch to `NodeType`

src/web/context/Global/GlobalContext.tsx
```diff
@@ -19,40 +20,84 @@ export interface SelectedPodElement {
...
-    commands: false,
-    links: false,
-};
-
 export const globalDefaultContext: GlobalContextType = {
     clientID: "",
     controllingClientID: "",
-    selectedPodElement: null,
-    showHubDetails: false,
+    selectedPodElement: { type: PodElement.NONE, id: -1 },
```
michael-jaia: Switch to `selectedNode`

src/web/context/Global/GlobalContext.tsx
```diff
@@ -138,132 +177,100 @@ function handleExitedRCMode(mutableState: GlobalContextType) {
...
-        mutableState.selectedPodElement.type === PodElement.HUB;
-
-    if (isHubSelected) {
-        mutableState.showHubDetails = false;
+function handleClickedNode(mutableState: GlobalContextType, type: PodElement, id: number) {
+    if (isNaN(id)) throw new Error("Invalid hub or bot id");
+
+    // Clicked currently selected element
+    if (mutableState.selectedPodElement.type == type && mutableState.selectedPodElement.id == id) {
+        // Close deselect and close details
```
michael-jaia: Maybe switch to "Close details and deselect node"

src/web/context/Global/GlobalContext.tsx
```diff
@@ -138,132 +177,100 @@ function handleExitedRCMode(mutableState: GlobalContextType) {
...
-function handleClickedHubTab(mutableState: GlobalContextType, hubID: number) {
-    if (isNaN(hubID)) throw new Error("Invalid hubID");
-
-    const isHubSelected =
-        mutableState.selectedPodElement !== null &&
-        mutableState.selectedPodElement.type === PodElement.HUB;
-
-    if (isHubSelected) {
-        mutableState.showHubDetails = false;
+function handleClickedNode(mutableState: GlobalContextType, type: PodElement, id: number) {
```
michael-jaia: Fantastic function, night and day from the previous implementation

src/web/context/Global/GlobalContext.tsx
```diff
@@ -138,132 +177,100 @@ function handleExitedRCMode(mutableState: GlobalContextType) {
...
-    if (isHubSelected) {
-        mutableState.selectedPodElement = null;
-        mutableState.showHubDetails = false;
-    } else {
-        mutableState.selectedPodElement = { type: PodElement.HUB, id: hubID };
-        mutableState.showHubDetails = true;
-    }
-    return mutableState;
-}
+function handleClickedHubAccordion(mutableState: GlobalContextType, accordionName: string) {
```
michael-jaia: I think we can have `accordionName` of type `HubAccordionNames`

src/web/context/Global/GlobalContext.tsx
```diff
@@ -138,132 +177,100 @@ function handleExitedRCMode(mutableState: GlobalContextType) {
...
 /**
- * Opens and closes the HubDetails accordion tabs
+ * Opens and closes the BotDetails accordion tabs
  *
  * @param {GlobalContextType} mutableState State object ref for making modifications
  * @param {string} accordionName Which accordion to open or close
  * @returns {GlobalContextType} Updated mutable state object
  */
-function handleClickedHubAccordion(mutableState: GlobalContextType, accordionName: string) {
+function handleClickedBotAccordion(mutableState: GlobalContextType, accordionName: string) {
```
michael-jaia: Same here, I think we can have `accordionName` of type `BotAccordionNames`

src/web/jest.config.js
```diff
@@ -171,6 +171,8 @@ const config = {
     // The regexp pattern or array of patterns that Jest uses to detect test files
     // testRegex: [],
 
+    // Only treat files in __tests__ directories with test in the file name as tests
+    testRegex: "__tests__/.*test\\.[jt]sx?$",
```
michael-jaia, line 175: Nice add!

src/web/types/jaia-system-types.ts
```diff
@@ -4,4 +4,5 @@ export interface MissionStatus {
     missionState?: MissionState;
     activeGoal?: number;
     distanceToActiveGoal?: number;
+    repeat_index?: number;
```
michael-jaia: Let's switch this to `repeatIndex` to fit the camelCase convention

src/web/utils/command.ts
```diff
@@ -93,3 +57,34 @@ export async function sendHubCommand(hubID: number, hubCommand: CommandInfo) {
         jaiaAPI.postCommandForHub(command);
     }
 }
+
+// TODO Look into consolidating the logic below
```
michael-jaia: Remove todo comment, I made a note to add a ticket

michael-jaia (changes requested):
Jeff, I just went through the second half of the PR. Great changes. The comments will include a couple of things I noticed along the way.

src/web/context/Global/GlobalContext.tsx
JRoybot: No we need that to get the bot ID

src/web/containers/BotDetails/BotDetails.tsx
```diff
@@ -0,0 +1,1135 @@
...
+
+    useEffect(() => {
+        addDropdownListener("accordionContainer", "botDetailsAccordionContainer", 30);
+    }, []);
+
+    // Make sure everything we need from Context is valid
+    // Otherwise do not render panel
+    if (
+        jaiaSystemContext === null ||
+        globalContext.visibleDetails != NodeType.BOT ||
```
michael-jaia: Do we need both `globalContext` checks?

src/web/containers/BotDetails/BotDetails.tsx
```diff
@@ -0,0 +1,1135 @@
...
+    );
+
+    useEffect(() => {
+        addDropdownListener("accordionContainer", "botDetailsAccordionContainer", 30);
+    }, []);
+
+    // Make sure everything we need from Context is valid
+    // Otherwise do not render panel
+    if (
+        jaiaSystemContext === null ||
```
michael-jaia, line 124: If we check for `jaiaSystemContext` being `null`, should we check for `globalContext` being `null`?

src/web/containers/BotDetails/BotDetails.tsx
```diff
@@ -0,0 +1,1135 @@
...
+                create: () => "none",
+            },
+        }),
+    );
+
+    useEffect(() => {
+        addDropdownListener("accordionContainer", "botDetailsAccordionContainer", 30);
+    }, []);
+
+    // Make sure everything we need from Context is valid
```
michael-jaia: What do you think of not having these comments here? I feel like the code tells the story in this case.

src/web/containers/BotDetails/BotDetails.tsx
```diff
@@ -0,0 +1,1135 @@
...
+    if (
+        jaiaSystemContext === null ||
+        globalContext.visibleDetails != NodeType.BOT ||
+        globalContext.selectedNode.type != NodeType.BOT
+    ) {
+        return <div></div>;
+    }
+
+    // Pull Data from the Context
+    const DEFAULT_HUB_ID = 1;
```
michael-jaia, line 131: This will be a good case for the ticket on creating a file for reusable const variables

src/web/containers/BotDetails/BotDetails.tsx
```diff
@@ -0,0 +1,1135 @@
...
+        }
+        return dataOffloadButton;
+    }
+
+    /**
+     * Provides Health Row
+     *
+     * @returns {React.Fragment} Health row HTML
+     */
+    function healthRow(bot: Bot, allInfo: boolean) {
```
michael-jaia, line 227: I would like to see this as its own component, but I think that will be better suited for a different ticket.

src/web/containers/BotDetails/__tests__/BotDetailsMockProps.tsx
```diff
@@ -0,0 +1,115 @@
...
+import {
+    HealthState,
+    BotType,
+    MissionState,
+    CommandType,
+    MissionStart,
+    MovementType,
+} from "../../../utils/protobuf-types";
+
+let mission: MissionInterface = {
```
michael-jaia: Since this interface is going to change, I don't think it makes sense to keep this mock and the test skeleton in the code

src/web/context/Global/GlobalContext.tsx
michael-jaia: What about changing `SelectedPodElement` to `SelectedNodeElement`?

src/web/containers/BotDetails/bot-details.ts
michael-jaia: What do you think of an implementation like this:
```
export function getWaypontHelperText(mission: Mission) {
    if (!mission || mission.getCanEdit()) {
        return "Click on the map to create waypoints"
    }
    return "Click edit toggle to create waypoint";
}
```

src/web/containers/BotDetails/BotDetails.tsx
```diff
@@ -0,0 +1,1135 @@
...
+    if (!bot) {
+        return <div></div>;
+    }
+
+    // TODO some mission related code was changed to use the mission
+    // from globalContext but the management of missions in globablContext
+    // is not complete.  getMissionID will return -1 untill it is done
+    // known bugs affected by this include
+    // Edit Mode Toggle, Play Button etc
+    let mission: Mission;
```
michael-jaia: I'm thinking we may not need this if statement because mission will be undefined if it does not exist regardless of this check

src/web/containers/BotDetails/BotDetails.tsx
JRoybot: I think it is a good idea because a bot has to be selected AND the visibleDetails should be Bot.   In theory one should never be set without the other but because that is external there is no garuantee

src/web/context/Global/GlobalContext.tsx
JRoybot: went with SelectedNode

src/web/containers/BotDetails/BotDetails.tsx
michael-jaia: If we get into a state where the `visibleDetails` equals `NodeType.BOT` but the `selectedNode.type` does not equal `NodeType.BOT` then we have a fundamental bug on our hands which should not make its way into the code with testing and peer reviews. I think the extra check can be perceived as not trusting what we know, and I do not want that to be sprinkled throughout the codebase.

src/web/containers/BotDetails/BotDetails.tsx
```diff
@@ -0,0 +1,1132 @@
...
+    const missionID = missionsManager.getMissionID(botID);
+    mission = missions.getMission(missionID);
+
+    const missionStatus: MissionStatus = bot.getMissionStatus();
```
michael-jaia, line 149: We need to be a little tighter with defining our variables at the top. I think we should be consistent with how we are getting the sensor data (even if it means we need a few more lines of code). GPS, IMU, Temperature, and Pressure should all be accessed in a similar structure. Same goes for missionState, sometimes we use the `missionState` variable and sometimes we use `missionStatus?.missionState`. I would like to see the whole chain, so we know where there is the possibility of a value being undefined.

src/web/containers/BotDetails/BotDetails.tsx
```diff
@@ -0,0 +1,1132 @@
...
+    }
+
+    // TODO some mission related code was changed to use the mission
+    // from globalContext but the management of missions in globablContext
+    // is not complete.  getMissionID will return -1 untill it is done
+    // known bugs affected by this include
+    // Edit Mode Toggle, Play Button etc
+    let mission: Mission;
+    const missionID = missionsManager.getMissionID(botID);
+    mission = missions.getMission(missionID);
```
michael-jaia: I think we should use `mission = jaiaSystemContext.missions.get(missionID)` and move this up with the Hub and Bot definitions for consistency

src/web/containers/BotDetails/BotDetails.tsx
```diff
@@ -0,0 +1,1132 @@
...
+        });
+    }
+
+    async function runRCMode(bot: Bot) {
+        if (!bot) {
+            warning("No bots selected");
+            return null;
+        }
+
+        const botLat = bot.getBotSensors()?.getGPS()?.getLat();
```
michael-jaia: Related to the comment up top, I think it is key, that we pick a way to access the data and be consistent with it.

src/web/containers/BotDetails/BotDetails.tsx
```diff
@@ -0,0 +1,1132 @@
...
+                                        className="nestedAccordionContainer"
+                                    >
+                                        <AccordionSummary
+                                            expandIcon={<ExpandMoreIcon />}
+                                            aria-controls="panel1a-content"
+                                            id="panel1a-header"
+                                        >
+                                            <Typography>GPS</Typography>
+                                        </AccordionSummary>
+                                        <AccordionDetails>
```
michael-jaia, line 1016: We do not need `?` for accessing the sensor data since the classes will have the accessor methods regardless of whether the data is present. We would need to `?` for the actions we take after accessing the data.

src/web/containers/BotDetails/BotDetails.tsx
JRoybot: I'll make the change but I was always taught to never trust another piece of code ;-)

src/web/containers/BotDetails/BotDetails.tsx
JRoybot: This is in one of the functions I did not touch because I think it will need to be rewritten later.  The reason for the inconsistency is because bot in that context is the inner function parameter not the bot declared in BotDetails When I started those functions were not nested in BotDetails so everything needed to be passed in.  Now that they are nested all the data is available as part of BotDetails
Do you want me to restructure all of those function parameter profiles for the functions we intend to revisit?

src/web/containers/BotDetails/BotDetails.tsx
JRoybot: Removed most of them, please check to make sure I did what you expected

src/web/containers/BotDetails/BotDetails.tsx
JRoybot: Made changes I thought were appropriate.  Please double check it is what you expected

michael-jaia (approved):

michael-jaia:
Jeff, great work sifting through `Details.tsx` and splitting the bulk of it into `BotDetails.tsx` and `bot-details.ts`. Integrating the component with our `GlobalContext`  and `JaiaSystemContext` is a big step forward.

PR #1065, opened 2025-01-29
Remove mindisk option as this leads to out of disk space on EC2

tsaubergine:
Increase disk image to 8GB overlay (same as real bots/hubs).

Removed `--mindisk`.

Increased Cloudhub default to 32 GB


ferrom-jaia (approved):
Looks good!

PR #1067, opened 2025-02-03
task/refactor-mission-controller-panel/JAIA-1830

michael-jaia:
## Overview
This pull request includes the initial work for refactoring the `MissionControllerComponent` component and its children. It establishes a base structure that can be extended to the complete feature set. I want to pause at this point to prevent to the pull request from becoming too large for a reviewer to digest.

The refactor creates the component `MissionsPanel` which will eventually replace `MissionControllerPanel`.

In the initial work:
* `RunAssignMenu` was refactored to `MissionAssignMenu`
* `RunList` was refactored to `MissionsList`
* `MissionsList` includes what was the `RunItem`
* `RunPanel` will no longer be needed

## Future Work
Tickets will be generated in JIRA to complete the refactor of all features in the missions panel and mission accordions.

## Testing
Tests written in `src/web/containers/MissionsPanel/__tests__` exercise the base functionality of the missions panel and mission accordions.


src/web/components/MissionAssignMenu/MissionAssignMenu.tsx
```diff
@@ -0,0 +1,94 @@
+// React
+import React, { useContext } from "react";
```
JRoybot: React does not appear to be referenced anywhere, probably can be deleted

src/web/components/MissionAssignMenu/MissionAssignMenu.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+// MUI
+import FormControl from "@mui/material/FormControl";
+import InputLabel from "@mui/material/InputLabel";
+import MenuItem from "@mui/material/MenuItem";
+import Select, { SelectChangeEvent } from "@mui/material/Select";
+
+// Jaia
+import Bot from "../../data/bots/bot";
```
JRoybot: Bot not referenced

src/web/components/MissionAssignMenu/MissionAssignMenu.tsx
```diff
@@ -0,0 +1,94 @@
...
+    return (
+        <FormControl sx={{ minWidth: 120 }}>
+            <InputLabel>Bot</InputLabel>
+            <Select
+                label="Assign"
+                onChange={(evt: SelectChangeEvent) => handleMenuSelection(evt)}
+                value={getMenuValue().toString()}
+            >
+                <MenuItem value={UNASSIGNED_ID}>Unassigned</MenuItem>
+                {Array.from(jaiaSystemContext.bots.values()).map((bot) =>
```
JRoybot, line 82: Should this array be sorted in case the bots are not in numeric order?

src/web/containers/MissionControllerPanel/RunPanel/RunList/RunItem/RunItem.tsx
```diff
@@ -151,11 +150,6 @@ export default class RunItem extends React.Component<RunItemProps, RunItemState>
                     </AccordionSummary>
                     <AccordionDetails>
                         <span className="runItemInfo">
-                            <RunAssignMenu
```
JRoybot, line 154: Will this panel eventually go away or should it be renamed so we are not using the term "Run"

src/web/containers/MissionsPanel/MissionsList/MissionsList.tsx
```diff
@@ -0,0 +1,165 @@
+// React
+import React, { useContext } from "react";
```
JRoybot: React not referenced

src/web/containers/MissionsPanel/MissionsList/MissionsList.tsx
```diff
@@ -0,0 +1,165 @@
...
+     * Triggered when the operator clicks the duplicate mission button
+     *
+     * @param {number} missionID ID of the mission to be duplicated
+     * @returns {void}
+     *
+     * @notes
+     * To be implemented in a separate ticket
+     */
+    const handleDuplicateMissionClick = (missionID: number) => {
+        globalDispatch({ type: GlobalActions.DESELECT_POD_ELEMENT });
```
JRoybot: GlobalActions is behind what is on developer-friendly-jcc  Should use `GlobalActions.CLICKED_NODE `

src/web/containers/MissionsPanel/MissionsList/MissionsList.tsx
```diff
@@ -0,0 +1,165 @@
...
+                            </AccordionDetails>
+                        </Accordion>
+                    </ThemeProvider>
+                );
+            })}
+        </div>
+    );
+}
+
+function MissionAccordionTitle(props: MissionAccordionTitleProps) {
```
JRoybot, line 144: Does this need to use a props interface?  Why not
`function MissionAccordionTitle(missionID: number)`

src/web/containers/MissionsPanel/MissionsPanel.tsx
```diff
@@ -0,0 +1,142 @@
+// React
+import React, { useContext } from "react";
```
JRoybot: React not referenced

src/web/containers/MissionsPanel/__tests__/MissionsPanel.test.tsx
```diff
@@ -0,0 +1,171 @@
+import { render, screen } from "@testing-library/react";
```
JRoybot, line 1: Very nice job testing this!

src/web/context/Global/GlobalContext.tsx
```diff
@@ -268,6 +284,45 @@ function handleClickedHubAccordion(mutableState: GlobalContextType, accordionNam
...
+    return mutableState;
+}
+
+/**
+ * Deselects the currently selected pod element
+ *
+ * @param {GlobalContextType} mutableState State object ref for making modifications
+ * @returns {GlobalContextType} Updated mutable state object
+ */
+function handleDeselectPodElement(mutableState: GlobalContextType) {
```
JRoybot: merge in changes from developer-friendly-jcc

src/web/data/missions_manager/missions-manager.ts
```diff
@@ -52,6 +57,52 @@ class MissionsManager {
...
+    /**
+     * Finds the lowest Bot ID that is not assigned to a mission
+     *
+     * @returns {number} Bot ID or (-1) if all Bots are already assigned to missions or outside of comms range
+     */
+    getNextAvailableBotID() {
+        for (let [botID, bot] of bots.getBots()) {
+            if (
+                this.getMissionID(bot.getBotID()) === this.UNASSIGNED_ID &&
+                convertMicrosecondsToSeconds(bot.getStatusAge()) < this.STATUS_AGE_MAX_SECONDS
```
JRoybot, line 84: Would it be better to check the health of the bot instead of just the status age?

src/web/data/missions_manager/missions-manager.ts
```diff
@@ -1,8 +1,13 @@
+import { bots } from "../bots/bots";
+import { missions } from "../missions/missions";
+import { convertMicrosecondsToSeconds } from "../../shared/Utilities";
+
 class MissionsManager {
     private botsToMissions: Map<number, number>;
     private missionsToBots: Map<number, number>;
 
     readonly UNASSIGNED_ID = -1;
```
JRoybot, line 9: This is declared in a bunch of individual files maybe define it in src/web/data/missions/missions.ts to enforce consistency

JRoybot (changes requested):
Looks good.  Left a few comments

src/web/containers/MissionControllerPanel/RunPanel/RunList/RunItem/RunItem.tsx
michael-jaia: Yes, this panel will eventually go away

src/web/containers/MissionsPanel/MissionsList/MissionsList.tsx
michael-jaia: We cannot just pass a number because React is expecting an object. We can spread like: `{ missionID }: MissionAccordionTitleProps` but we still need the interface

src/web/data/missions_manager/missions-manager.ts
michael-jaia: I think we want to use status age because it will take into account a "healthy" or "unhealthy" Bot that is outside of comms range.

src/web/components/MissionAssignMenu/MissionAssignMenu.tsx
michael-jaia: Should be good once with the sorted map now!

src/web/data/missions_manager/missions-manager.ts
michael-jaia: Yes, we have a ticket here: https://jaia-innovation.atlassian.net/browse/JAIA-1844?atlOrigin=eyJpIjoiYTg0OTMzMmI4NThiNDcyYWI0ZWMwOTQzMGFiOTY5ZWYiLCJwIjoiaiJ9

src/web/data/jaia_global/jaia-global.ts
```diff
@@ -23,7 +23,7 @@ class JaiaGlobal {
     }
 
     deselectNode() {
```
JRoybot, line 25: Does this function need to be public.  Any caller of it would know the selected node and could just call setSelectedNode.  Then the calling functions would not need to check if it was an already selected node

src/web/containers/MissionsPanel/MissionsList/MissionsList.tsx
```diff
@@ -86,7 +87,12 @@ export default function MissionsList() {
      * To be implemented in a separate ticket
      */
     const handleDuplicateMissionClick = (missionID: number) => {
-        globalDispatch({ type: GlobalActions.DESELECT_POD_ELEMENT });
+        // Update data model
+        jaiaGlobal.deselectNode();
```
JRoybot: Could we update `jaiaGlobal` in the `handleClickedNode` of `src/web/context/Global/GlobalContext.tsx` so all the code that deals with clicking nodes does not need to make both calls?

JRoybot (changes requested):

src/web/containers/MissionsPanel/MissionsList/MissionsList.tsx
michael-jaia: We can, I hesitate to do this because it breaks the pattern of 1) updating the data model, 2) updating OpenLayers, and 3) updating React. I don't like making two calls for the same purpose, but I also think consistency is important so others can follow easier. What do you think about this?

src/web/data/jaia_global/jaia-global.ts
michael-jaia: I'm a little confused, are you saying we don't really need deselectNode public because we can make a call like: `jaiaGlobal.setSelectedNode(jaiaGlobal.getSelectedNode())`?

JRoybot: I see now, the only use of it is when a new mission is created so you are just deselecting whatever is selected .  Retract original comment.

src/web/containers/MissionsPanel/MissionsList/MissionsList.tsx
JRoybot: Agree it is a tricky paradigm.  I think needing to make 2 calls to 2 different modules to achieve a single action is never a good idea and is going to be a constant source of confusion and errors.  Ideally we would not need to track SelectedNode in 2 places but are only doing it for OL.  Ideally, only the OL code would need to import jaiaGlobal but the pattern we have now requires it any place a node can be selected.

Reviewing the way handleClickedNode works now that jaiaGlobal was created is pretty wonky.  SelectedNode is part of GlobalContext but we are relying on the state of jaiaGlobal.getSelectedNode() being different than what GlobalContext thinks it is.  And that ONLY works because the developer knew he had to call jaiaGlobal.setSelectedNode first.  It would be one thing if we had to call 2 functions if they both worked independently but we have a model that one relies on the other being called first to work properly.

I think handleClickedNode should be changed back to be more like it was and accept a SelectedNode as an action parameter.  It should check if it is different than what is in GlobalContext, change what is necessary in GlobalContext and THEN update jaiaGlobal to be in sync with it.

If you want to discuss more we can huddle.

michael-jaia:
Migrated changes to data model to React. I'm wondering if you think there is a little too much redundancy between the handler functions in `JaiaSystemContext.tsx` when modifying `mutableState`. For instance the multiple calls that go like:
```
mutableState.bots = bots.getBots();
mutableState.missions = missions.getMissions();
```

src/web/containers/MissionsPanel/MissionsPanel.tsx
```diff
@@ -55,20 +51,15 @@ export default function MissionsPanel() {
...
      */
     const handleDeleteAllMissionsClick = () => {
-        // Update data model
-        missions.deleteAllMissions();
-        missionsManager.clear();
-
-        // Update OpenLayers
-
-        // Update Context
+        // Clear accordion states
```
JRoybot: confusing comment

src/web/containers/MissionsPanel/MissionsPanel.tsx
```diff
@@ -55,20 +51,15 @@ export default function MissionsPanel() {
...
         globalDispatch({ type: GlobalActions.CLICKED_DELETE_ALL_MISSIONS });
```
JRoybot: Should we change `GlobalActions.CLICKED_DELETE_ALL_MISSIONS` to `GlobalActions.DELETE_ALL_MISSIONS` for consistency?

src/web/containers/MissionsPanel/MissionsPanel.tsx
```diff
@@ -55,20 +51,15 @@ export default function MissionsPanel() {
...
-        jaiaSystemDispatch({ type: JaiaSystemActions.SYNC_REQUESTED });
+
+        jaiaSystemDispatch({ type: JaiaSystemActions.DELETE_ALL_MISSIONS });
```
JRoybot, line 59: When reading this it is not clear why both events need to be dispatched, maybe can be cleared up with comments.

src/web/containers/MissionsPanel/MissionsPanel.tsx
```diff
@@ -84,16 +75,12 @@ export default function MissionsPanel() {
...
      *
      * @returns {void}
      */
     const handleAutoAssignClick = () => {
-        // Update data model
-        missionsManager.autoAssign();
-
-        // Update JaiaSystemContext
-        jaiaSystemDispatch({ type: JaiaSystemActions.SYNC_REQUESTED });
+        jaiaSystemDispatch({ type: JaiaSystemActions.AUTO_ASSIGN_BOTS_TO_MISSIONS });
```
JRoybot: What do you think of just `JaiaSystemActions.AUTO_ASSIGN_MISSIONS`

src/web/context/JaiaSystem/JaiaSystemContext.tsx
```diff
@@ -42,10 +44,20 @@ export const JaiaSystemDispatchContext = createContext(null);
 function jaiaSystemReducer(state: JaiaSystemContextType, action: Action) {
     let mutableState = { ...state };
     switch (action.type) {
+        case JaiaSystemActions.INIT:
+            return handleInit(mutableState);
         case JaiaSystemActions.DATA_MODEL_POLLED:
             return handleDataModelPolled(mutableState, action.bots, action.hubs);
```
JRoybot: Since `JaiaSystemActions.DATA_MODEL_POLLED` is only dispatched from within this file why do we need to pass the bots and hubs maps through the action interface?  Couldn't `handleDataModelPolled` just do `bots.getBots()` & `hubs: hubs.getHubs()` instead of `pollDataModel`

src/web/context/JaiaSystem/JaiaSystemContext.tsx
```diff
@@ -42,10 +44,20 @@ export const JaiaSystemDispatchContext = createContext(null);
 function jaiaSystemReducer(state: JaiaSystemContextType, action: Action) {
     let mutableState = { ...state };
     switch (action.type) {
+        case JaiaSystemActions.INIT:
```
JRoybot, line 44: I'm a little confused why we need INIT and DATA_MODEL_POLLED.  Might be easier to explain to me outside of the review.

JRoybot (changes requested):
Might make sense to have a huddle when you check my comments.  There has been a lot going on in this review and want to make sure we both understand each others comments.

src/web/containers/MissionsPanel/MissionsPanel.tsx
michael-jaia: Changed to RESET_MISSION_ACCORDIONS

src/web/containers/MissionsPanel/MissionsPanel.tsx
michael-jaia: Is it clearer with the renaming of the action?

src/web/containers/MissionsPanel/MissionsPanel.tsx
```diff
@@ -51,15 +51,13 @@ export default function MissionsPanel() {
...
+     * Dispatches the actions to clear all missions when an operator clicks the delete all missions button
      *
      * @returns {void}
      */
     const handleDeleteAllMissionsClick = () => {
-        // Clear accordion states
-        globalDispatch({ type: GlobalActions.CLICKED_DELETE_ALL_MISSIONS });
-
         jaiaSystemDispatch({ type: JaiaSystemActions.DELETE_ALL_MISSIONS });
+        globalDispatch({ type: GlobalActions.RESET_MISSION_ACCORDIONS });
```
JRoybot, line 60: Reads 1000% better!

src/web/context/JaiaSystem/JaiaSystemContext.tsx
```diff
@@ -46,51 +43,46 @@ function jaiaSystemReducer(state: JaiaSystemContextType, action: Action) {
...
 
 /**
- * Puts Context in sync with the data model
+ * Saves the latest data from incoming Bot and Hub status messages to state
  *
  * @param {JaiaSystemContextType} mutableState State object ref for making modifications
  * @returns {JaiaSystemContextType} Updated mutable state object
  */
-function handleInit(mutableState: JaiaSystemContextType) {
+function handleDataModelPolled(mutableState: JaiaSystemContextType) {
```
JRoybot: How bout handlePollDataModel to match the new Action name?

JRoybot (dismissed):
Looks really good.  Left one nitpick, up to you if you change or not.

PR #1071, opened 2025-02-06
task/refactor-node-list-panel/JAIA-1842

JRoybot:
Refactored NodeListPanel to use Context instead of Props and general clean up of code.

- Removed all Props
- Used Context for all data
- Converted faultLevel from function to Map
- Removed BotTab and HubTab functions and rendered all in place
- Created new helper functions to support new structure
- Moved styles from CommandControl to NodeListPanel

**Testing**

Tested by exercising code in browser and checking all styles and functionality.
Unit test verifies Nodes are sorted correctly, have correct style, and change style when clicked

**Notes**  
- Because the panel does not use callbacks to tell CommandControl a hub or bot selection does not affect mission editing.  To create waypoints one needs to use the Mission Panel for now.  NodeListPanel is dispatching GlobalAction to select node.  This functionality has been verified but is not used everywhere yet.
- Considered creating an abstract class for nodes and combining the bots and hubs maps in jaiaSystemContext but decided to leave as is for now.  Was not clear the benefits outweighed the impact of the change.


src/web/containers/NodeListPanel/NodeListPanel.tsx
```diff
@@ -4,136 +4,91 @@ import {
     GlobalDispatchContext,
```
michael-jaia: What do you think of calling this component `NodeList`?

src/web/containers/NodeListPanel/NodeListPanel.tsx
```diff
@@ -4,136 +4,91 @@ import {
...
-        const handleClick = () => {
-            props.didClickHub(hub.hub_id);
-            globalDispatch({
-                type: GlobalActions.CLICKED_NODE,
-                nodeType: NodeType.HUB,
-                nodeID: hub.hub_id,
-            });
-        };
+    // NodeListPanel
+    const jaiaSystemContext = useContext(JaiaSystemContext);
```
michael-jaia: I think it would be nice to move lines 61-67 to the top of the component function for consistency with other components and we will have the values defined before they are used.

src/web/containers/NodeListPanel/NodeListPanel.tsx
```diff
@@ -4,136 +4,91 @@ import {
...
+    };
+
+    /**
+     *
+     * @param {NodeType} nodeType indicates it it is a Bot or Hub
+     * @param {number} nodeID ID of hub or bot
+     * @param {HealthState} healthState Health of hub or bot
+     * @returns {string} className
+     */
+    function getClassName(nodeType: NodeType, nodeID: number, healthState: HealthState) {
```
michael-jaia: In this function lets use `const` for the variable types and `===` to be strict about the equality

src/web/containers/NodeListPanel/NodeListPanel.tsx
```diff
@@ -4,136 +4,91 @@ import {
...
+     */
+    const handleClick = (nodeType: NodeType, nodeID: number) => {
+        globalDispatch({
+            type: GlobalActions.CLICKED_NODE,
+            nodeType: nodeType,
+            nodeID: nodeID,
+        });
+    };
+
+    /**
```
michael-jaia: Add a quick description about what the function does on the first line of the docstring

michael-jaia (changes requested):
Jeff nice work, the component has a good flow to it, and it is easy to follow what is happening. I left a few comments on things we might be able to improve before merging.

src/web/containers/NodeListPanel/NodeListPanel.tsx
JRoybot: No strong preference but feel NodeListPanel is more consistent with the ohter containers, I would expect NodeList to be a component more than a container

src/web/containers/NodeListPanel/NodeListPanel.tsx
JRoybot: Are you asking for a docstring added to NodeListPanel or modify the one on handleClick?

src/web/containers/NodeListPanel/NodeListPanel.tsx
JRoybot: OK, will update

src/web/containers/NodeListPanel/NodeListPanel.tsx
michael-jaia: Ok that is something I am starting to notice that the line between components and containers is getting blurry as we do the clean up. NodeListPanel is not really a panel in the sense that it opens a window to display addition information so I think we should drop panel from the name. I'm ok with leaving it in containers folder or putting it in the components folder. I think towards the end, we will get clearer on the distinction and do some folder shifting.

src/web/containers/NodeListPanel/NodeListPanel.tsx
michael-jaia: Just adding a description as part of the docstring format to the getClassName function

michael-jaia: Just adding a description as part of the docstring format to the getClassName function

michael-jaia (approved):

PR #1074, opened 2025-02-12
task/remove-run-info-panel/JAIA-1855

JRoybot:
- Removed RunInfoPanel and all references to it
- Removed logic to handle clicking the Flag icons
- Changed Flag icons to display "M" instead of "R"

Notes: 

- `runList: MissionInterface;` of the `CommandControl` `state` is still used in much of the code and still needs to be factored out
- `src/web/jdv/client/src/shared/MissionFeatures.ts` will need further refactoring once runList is removed from all data models


src/web/shared/Styles.ts
```diff
@@ -604,11 +605,10 @@ export function getFlagStyle(
 ) {
     let icon = createFlagIcon(goal.task?.type, isSelected, Number(runNumber), canEdit);
     const isTask = goal.task?.type && goal.task.type !== "NONE";
-
     return new Style({
         image: icon,
         text: new Text({
-            text: `R${runNumber}`,
+            text: `M${runNumber}`,
```
JRoybot, line 611: This was the only change made, rest was formatted on commit

michael-jaia (approved):

PR #1079, opened 2025-02-18
task/show-waypoints-on-missions-layer/JAIA-1862

michael-jaia:
## Overview
This pull request creates waypoints when a mission is in edit mode and the operator clicks on the map as well as when the a Bot is selected without a mission.

## Details
When a user clicks on the map with a mission in edit mode, the flow of informaiton will be as follows:
1. The click will trigger `handleMapClick` in `components/Map.tsx`
2. This will lead to the function call `handleAddWaypointClick(coordinate)`
3. Here the `ADD_WAYPOINT` action will be dispatched with the lat/lon of the click
4. In `context/JaiaSystem/JaiaSystemContext.tsx`, `handleAddWaypoint(mutableState, location)` will be called.
    * If the `missionIDInEditMode` property attached to the `missions` singleton does not equal (-1), then we call `mission.addWaypoint(location)`
5. `mission.addWaypoint(location)` lives in `data/missions/mission.ts`, and it pushes a `Waypoint` to the end of the `waypoints` array. After it calls `missionLayer.addWaypoint(missionID, waypointNum)`.
6. `missionLayer.addWaypoint` in `openlayers/layers/vector/mission-layer.ts` makes the calls to place the waypoint icon (and connecting lines if necessary) to the map.

## Notes
1. Switched waypoint property `position` to `location` to match with `Bot` and `Hub` properties
2. The `Missions` class now has a property `missionIDInEditMode` to replace the `canEdit` property on the `Mission` class. This allows us to know what mission is in edit mode without having to iterate through the map of missions.
3. The `Mission` class now handles the `moveableWaypoint`. This replaces the `canMoveOnMap` property on the `Waypoint` class to simplify state management.

## Tests
Automated tests include:
* Clicking on the map twice with a mission in edit mode to show waypoint and line segment creation
* Clicking on the map with no mission in edit mode to show no waypoints appear

## JIRA
* Story: https://jaia-innovation.atlassian.net/browse/JAIA-1812
* Combines subtasks: **1862**, **1873**, and **1874**


src/web/utils/constants.ts
```diff
@@ -1,4 +1,5 @@
 export const UNASSIGNED_ID = -1;
+export const NO_MISSION_ID = -1;
```
JRoybot: Is there a reason UNASSIGNED_ID can't be used?  It is already used for bots and hubs, could be used for missions as well.

src/web/types/openlayers-types.ts
```diff
@@ -9,4 +9,6 @@ export enum MapFeatureTypes {
     NONE = "NONE",
     BOT = "BOT",
     HUB = "HUB",
+    WAYPOINT = "WAYPOINT",
```
JRoybot, line 12: Don't see anything for the Mission (was Run) flags on the map.  Why are they handled differently?

src/web/components/Map/Map.tsx
```diff
@@ -36,8 +48,16 @@ export default function Map() {
                     return;
             }
         }
+
+        handleAddWaypointClick(event.coordinate);
```
JRoybot, line 52: It looks like this only works for adding waypoints, not selecting them.  If a user were to selects an existing waypoint will there be a different handler in the switch above?

src/web/containers/BotDetails/bot-details.ts
```diff
@@ -70,7 +71,7 @@ export function getDistanceToHub(botGPS: GPS, hubGPS: GPS) {
  * until mission management refactor is complete
  */
 export function getWaypontHelperText(mission: Mission) {
-    if (!mission || mission.getCanEdit()) {
+    if (!mission || missions.getMissionIDInEditMode() === mission.getMissionID()) {
```
JRoybot, line 75: Is this temporary?  I thought the description said enabling editing missions from the Details panel was to be removed.

src/web/containers/BotDetails/bot-details.ts
```diff
@@ -285,20 +286,23 @@ export function disablePlayButton(
 
     return disableInfo;
 }
+
 /**
  * @notes Refactor comming with new mission management implementation
  */
 export function toggleEditMode(mission: Mission) {
```
JRoybot, line 294: same comment, temporary?

src/web/data/missions/mission.ts
```diff
@@ -1,10 +1,12 @@
+import { GeographicCoordinate } from "../../utils/protobuf-types";
+import { missionLayer } from "../../openlayers/layers/vector/mission-layer";
 import Waypoint from "../waypoints/waypoint";
 
 export default class Mission {
     private missionID: number;
     private waypoints: Waypoint[];
     private repeats: number;
-    private canEdit: boolean;
+    private movableWaypoint: Waypoint;
```
JRoybot: could we just use an index to the array instead of the object reference?

src/web/data/missions/missions.ts
```diff
@@ -38,10 +41,19 @@ class Missions {
         return this.getMissions().get(missionNum);
     }
 
+    getMissionIDInEditMode() {
```
JRoybot, line 44: Will we ever need a getMission(missionID : number) function?  Seems like that could be useful.

src/web/openlayers/features/waypoint-feature.ts
```diff
@@ -0,0 +1,174 @@
...
+ * @param {number} missionID Used to determine color of waypoint
+ * @returns {Feature} Waypoint icon to display on map
+ */
+export function generateWaypointFeature(
+    location: GeographicCoordinate,
+    waypointNum: number,
+    missionID: number,
+) {
+    if (!location) {
+        return new Feature();
```
JRoybot, line 35: Are we forcing all users of this function to handle the return of an empty Feature?

src/web/openlayers/features/waypoint-feature.ts
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    const coordinate: Coordinate = [location.lon, location.lat];
+    const feature = new Feature({
+        geometry: new Point(fromLonLat(coordinate, view.getProjection())),
```
JRoybot, line 40: Is there a "set" function for adding the Point to the feature?  I think it would be easier to understand if that was done the same way "type" , "ID" and style are assigned.  Could create empty feature above line 34 and then line 35 could just return that.

src/web/openlayers/features/waypoint-feature.ts
```diff
@@ -0,0 +1,174 @@
...
+export function generateWaypointLineFeature(
+    startLocation: GeographicCoordinate,
+    endLocation: GeographicCoordinate,
+    missionID: number,
+) {
+    if (!startLocation || !endLocation) {
+        return new Feature();
+    }
+
+    const startCoordinate = fromLonLat(
```
JRoybot, line 95: it's so weird OL uses "LonLat" convention.  I've been dealing with geo-spacial software for 40+ years and have never come across anything that didn't use "LatLon" convention.

JRoybot (changes requested):
Nice Job on this.  I left a few nit picky comments

src/web/utils/constants.ts
michael-jaia: I wasn't sure about this one...it felt a little weird to use UNASSIGNED_ID for missionIDInEditMode because it is not dealing with Bot and mission assignments in my head. I also didn't love the idea of having to repeat something that seems similar too. What do you think?

src/web/types/openlayers-types.ts
michael-jaia: We will add the run flags in a separate ticket

src/web/components/Map/Map.tsx
michael-jaia: Exactly there will be a switch for type WAYPOINT

src/web/containers/BotDetails/bot-details.ts
michael-jaia: Going to keep since I will be adding the click Bot to create a mission functionality back

src/web/data/missions/mission.ts
michael-jaia: Yes, I can make this change

src/web/data/missions/missions.ts
michael-jaia: Yeah I think when someone needs that they can add it, or we can create a ticket to add it

src/web/openlayers/features/waypoint-feature.ts
michael-jaia: I think the empty feature serves as a good "null" value. Would you rather see us explicitly return `null`?

src/web/openlayers/features/waypoint-feature.ts
michael-jaia: I will look into that

src/web/openlayers/features/waypoint-feature.ts
michael-jaia: Yeah I agree, I was confused when I first started working with OL.

src/web/openlayers/features/waypoint-feature.ts
JRoybot: I defer to you on this, just curious what is best in the React paradigm, is there example of how to handle the empty feature?

src/web/openlayers/features/waypoint-feature.ts
michael-jaia: Yes we can do that

```
  const feature = new Feature();
  feature.setGeometry(new Point(fromLonLat(coordinate, view.getProjection()))),
```

I will make a separate ticket to make the changes for all of the features.

src/web/openlayers/features/waypoint-feature.ts
michael-jaia: The empty feature is handled by OpenLayers in source.addFeature()

PR #1053, opened 2025-01-09
Rework Fleet Config / First Boot: JAIA-1806, JAIA-1791, JAIA-1773, JAIA-1776, JAIA-1772; XBee improvements: JAIA-1775, JAIA-1782; Ansible job for major upgrade: JAIA-1724

tsaubergine:
## Overview

This PR removes our custom first-boot.sh script and preseed in favor of a `cloud-init` based first boot configuration setup (https://jaia-innovation.atlassian.net/browse/JAIA-1806). Fleet config is now done as a step **before** creating and booting a fleet and creates a Protobuf TextFormat fleet configuration file that encapsulates the relevant parts of the fleet setup (https://jaia-innovation.atlassian.net/browse/JAIA-1791).

All the new fleet config tools are part of new `jaia admin fleet` actions.

If you're recreating an existing fleet, you'll need to first manually remove the VPN on vpn.jaia.tech, e.g. for fleet5:

```
jaia ssh vpn.jaia.tech
sudo systemctl stop wg-quick@wg_fleet5.conf
sudo systemctl disable wg-quick@wg_fleet5.conf
sudo rm /etc/wireguard/wg_fleet5.conf
```

For example, the process of creating a new fleet 5:

```
# Answer the whiptail questions
jaia admin fleet create private_jaia/fleet_config/fleet5.cfg

# Temporarily authorize the new bots/hubs to connect to vpn.jaia.tech and add themselves to a new VPN
jaia admin fleet vpn_authorize private_jaia/fleet_config/fleet5.cfg

# insert flashed SD card for Hub 1
jaia admin fleet generate private_jaia/fleet_config/fleet5.cfg hub 1
# unmount & remove Hub 1 card, insert flashed SD card for Hub 2
jaia admin fleet generate private_jaia/fleet_config/fleet5.cfg hub 2
# unmount & remove Hub 2 card, insert flashed SD card for Bot 1
jaia admin fleet generate private_jaia/fleet_config/fleet5.cfg bot 1
# unmount & remove Bot 1 card, insert flashed SD card for Bot 2
jaia admin fleet generate private_jaia/fleet_config/fleet5.cfg bot 2

# Deauthorize the temporary key (optional, the key is only authorized for 1 day in the `jaia admin fleet vpn_authorize` step above)
jaia admin fleet vpn_authorize private_jaia/fleet_config/fleet5.cfg --rm
```

## Rationale

[...]

PR #1091, opened 2025-02-26
Task/implement comms over wifi/jaia 254

tsaubergine:
Implements wifi comms in addition to or instead of XBee comms on runtime fleets.

## Major Changes

1. Changed `jaia_comms_mode` environmental variable to be a single value (as before: e.g., 'wifi', 'xbee') OR a comma separated list of modes, e.g., 'wifi,xbee'. This will allow us to grow to support more comms modes in the future and choose the appropriate set for a given fleet. 
2. Updated debconf on jaiabot-embedded to support multiselect field to choose modes (which get set in /etc/jaiabot/runtime.env via systemd.py). Defaults to 'xbee' to preserve existing behavior on package update unless new settings are given (via `dpkg-reconfigure jaiabot-embedded` or new fleet config).
3. Changed subnet index for wifi to '1' to remove conflict with XBee (which remains as '0'). 
4. Forked goby UDP_DRIVER as Jaia "libjaiabot_wifi" driver to allow us to write in support for multi-hub 
5. Implemented multi-hub in Jaia "libjaiabot_wifi" in an analogous fashion to the Xbee driver (swap IP addresses rather than Xbee serial numbers).
6. Added Link enumeration in protobuf and assign as a field to DCCL messages upon receipt to record/display which link the message came over. This is implemented via the set_link_data callback of the goby::middleware::Subscriber and determining the Link from the source ID address (new `jaiabot::comms::set_link_type` function).
7. New ansible playbook `change-comms-links.yml` to allow JCU to change the comms links in use.

##  Notes:
1. Until we switch to using the Ansible inventory (https://jaia-innovation.atlassian.net/browse/JAIA-1776 / https://github.com/jaiarobotics/jaiabot/pull/1053) to determine the managed bots list, this PR will require an XBee configuration (even when using Wifi only) for hub manager to support Hubs which have an ID that isn't 1 (the default hub). This is because the subscriptions for hub manager are determined off the Xbee configuration list (until JAIA-1776 is resolved).
2. The hub_manager is now configured to subscribe to bots on all the configured links (all the links defined in `jaia_comms_mode`)

## Testing

1. Tested in local simulation (`jaiabot/config/launch/simulation`) swapping hub=1 and hub=2 (WIFI only).
2. Tested using flat fleet 0 (for hub=1 only) for both XBee and WIFI simultaneously.
3. Updated unit test for comms conversions (modem id <-> bot id, etc. in comms.h).
4. Tested using VirtualBox fleet with hub=1 and hub=2 (WIFI only). Hub=2 has to use scan for bots as it doesn't have an Xbee configuration (see Note 1 above).

## Minor Changes

Three changes to remove port conflicts on localhost simulations with multiple hubs:
1. Liaison in simulation uses ports 30010 + hub_id and 30100 + bot_id (runtime uses 30000 as before). 
2. Use UDP port 31000 for Wifi packets at runtime, and port 31000 + hub_id (for hubs) / port 31000 + 31 (max hubs) + node_id (for bots) for simulation.
3. Use UDP port 32000 + hub id (for hubs) and 32100 + node_id (for bots) for GPSD at simulation.

Also:
1. Fixed `source config/launch/simulation/preseed.goby` for zsh users.
2. hub_manager now keeps track of bot_ids not bot_modem_ids which simplifies the logic now that we have a one-to-many mapping of bot_ids to modem_ids. This allows for cleaner code than trying to preserve the previous approach.
3. Updated comms.h functions to support one-to-many bot/hub IDs to modem_ids.
4. Updated check in mission manager for duplicate message rejection to reject older messages as well as same aged messages, in case one of our comms links is slower and provides older messages, these are rejected.


tsaubergine:
I merged 1.y updates and retested in VirtualBox and using fleet0's bot1/hub1

src/bin/mission_manager/app.cpp
```diff
@@ -792,7 +801,7 @@ void jaiabot::apps::MissionManager::handle_command(const protobuf::Command& comm
 
     // Make sure the command is not a repeat
     // If it is, then we should not handle the command and exit
-    if (prev_command_time_ == command.time())
+    if (prev_command_time_ >= command.time())
```
ferrom-jaia: This probably needs to be added to the comment, but this will occasionally cause issues. 

Comment from you:
"DCCL time codec only sends 24 hours (by default, this can be increased) worth of time to save space, and then realigns the time on the receive side. As you can't always count on causality (due to minor time mismatches), it uses the closest time, which means you get +/-12 hours before it's wrong."

**Example of scenario:**
The customer hub had a date: March 8th
The customer bot had a date: April 14th

* Hub - time does not update with ntp sync before sending command to bot
* Bot - records the prev_command_time
* Hub - time syncs to the correct time during the mission
* Hub - sends command to bot
* Bot - checks the command time it no longer accepts the new commands

tsaubergine: Hmm. I think you have the scenario backwards (I think this would only trigger when the hub had a future date that gets reset to the "past") but the inverse would cause the problem you're talking about.

The bigger issue I'm concerned about is:

1. Command 1 is sent via acomms and wifi
2. Command 1 is received via wifi.
3. Command 2 is sent via acomms and wifi
4. Command 2 is received via wifi.
5. Command 1 is received via acomms.

If we do == then step 5 would accept command 1. If we do >= then command 1 is rejected after command 2 is received (via any comms pathway).

Can we have a warning for bad NTP sync? Do we not already?

tsaubergine: Alternatively we can add a counter to the hub command that isn't the time stamp and use this for duplicate rejection (instead of the timestamp logic). 16 bits or so should be plenty large to not even worry about overflow during any sane mission.

ferrom-jaia: The scenario I described occurred on a customer system. It was something to do with the encoding of the time on the hub and then the decoding of the time on the bot. 

I like the counter, the only issue is if the hub is turned off during a mission (battery dies or other issue) and then the bot would discard the commands. 

We might want to keep a history of timestamps that we previously accepted.

The bad ntp sync might be the way to go so the hub cannot send commands until the hub gets a good one or the user inputs a time.

tsaubergine: I think a circular buffer (of some reasonable size, e.g. 100) of the timestamps of previously accepted commands would be more than sufficient to solve this problem. If the timestamp is in the buffer we reject the command as already seen.

tsaubergine: Separately I think we should definitely warn on invalid NTP sync as this can invalidate data.

ferrom-jaia: Yes, that sounds great! 

And I agree. I will make sure we have a ticket.

ferrom-jaia: Added ticket: https://jaia-innovation.atlassian.net/browse/JAIA-1900

tsaubergine: Added buffer of last 100 timestamps: https://github.com/jaiarobotics/jaiabot/pull/1091/commits/7fd10b792d85566bfd9c00b4b0bb47b705052eff

ferrom-jaia (approved):
Looks great! Tested in the field with fleet 3. Tested with xbee radio, wifi, and xbee and wifi.

PR #1092, opened 2025-02-27
task/make-single-context/JAIA-1898

michael-jaia:
### Summary
After thinking more about multiple contexts versus a single context, the team decided to move forward with a single context structure. In [1058](https://github.com/jaiarobotics/jaiabot/pull/1058), we combined `BotContext`, `HubContext`, and `MissionContext` into `JaiaSystemContext` to avoid calls to different contexts to achieve a single objective. In hindsight, this should have a been an indication to create a single context for the entire application, however, I was off the mindset that separating UI logic (such as opening and closing panels) from system operations (such as adding a waypoint) would make navigating the code easier. 

After further development, we realized we were in a similar situation making multiple calls to achieve a single objective. Furthermore, edge cases appeared which made it difficult to decide whether a piece of data should live in `GlobalContext` versus `JaiaSystemContext`. Now, we will use a single context `JaiaContext`. Towards the end of the project, we can look into moving functions into separate files, but for now I think the best path forward is to continue building out the application which will give us a better sense of what should be grouped together.

### Misc
The Python changes are needed to build the web app. Once the long term fix is merged into `1.y` we can merge those changes into `developer-friendly-jcc` and remove the temporary changes.


JRoybot (approved):
Great Job on this.  Everything looks good and is simpler now!

PR #1094, opened 2025-03-01
task/configure-visible-panel-structure-to-integrate-new-code/JAIA-1876

michael-jaia:
### Summary
Establishes a minimal JCC that renders the refactored code. This will provide us a clean environment for continuing the development of the developer friendly JCC.

minimal-jcc


src/web/components/SideButtonList/SideButtonList.tsx
```diff
@@ -0,0 +1,60 @@
...
+        return <div></div>;
+    }
+
+    /**
+     * Dispatches action to open/close panel associated with button
+     *
+     * @param {PanelNames} panelName Name of panel associated with button
+     * @returns {void}
+     */
+    const handlePanelButtonClick = (panelName: PanelNames) => {
```
JRoybot, line 27: Already commented, might want this to be more generic `handleButtonClick` or ``handleSideButtonClick` and use an enum of buttons if we think we may have or end up with a button that is not associated with a Panel

src/web/data/missions/missions.ts
```diff
@@ -46,7 +46,6 @@ class Missions {
         mission.setMissionID(missionID);
         this.setMissionIDInEditMode(missionID);
         this.setNextMissionID(this.getNextMissionID() + 1);
-        return missionID;
```
JRoybot: Please leave this in I used the return value when I implemented opening the accordion in handleAddMission as part of my PR

src/web/style/openlayers/colors.ts
```diff
@@ -1,5 +1,5 @@
 export enum OpenLayersColors {
-    DEFAULT = "whtie",
+    DEFAULT = "white",
```
JRoybot, line 2: :-)

JRoybot (changes requested):
This looks good, just put the return value back in addMission

src/web/components/SideButtonList/SideButtonList.tsx
michael-jaia: Yes, I think I want to leave it like this for now. Not all buttons need to render a panel, for example, if a panel does not need to be rendered such as when clicking on the rally button, we would add a function handleRallyButtonClick that dispatches the appropriate action. handlePanelButtonClick should just be a reusable function for the buttons that do render a panel.

JRoybot (approved):

PR #1095, opened 2025-03-02
task/clean-up-command-sending-code/JAIA-1843

michael-jaia:
### Summary
* Now that the confirmation for sending commands is handled by the buttons, we just need to make a call to utilize the `jaiaAPI` object to post the command.
* Each command button has a function `onDialogClose` to prepare a `Command` (`src/web/utils/protobuf-types`)  and send it via a call to `sendBotCommand` or `sendHubCommand` in `src/web/utils/commands.ts` which uses the `jaiaAPI`.

### Notes
* Given the shake up to to the command sending structure, it seemed like the appropriate time to remove the old command code from `BotDetails.tsx`. This is also a necessary step to prepare the component to be visible in the minimal JCC app.
* This branch is built off of `task/configure-visible-panel-structure-to-integrate-new-code` so we want to merge https://github.com/jaiarobotics/jaiabot/pull/1094 **prior to reviewing this pull request**.


src/web/containers/BotDetails/BotDetails.tsx
```diff
@@ -69,41 +55,10 @@ import ExpandMoreIcon from "@mui/icons-material/ExpandMore";
...
-export function BotDetails(props: BotDetailsProps) {
-    // TODO We will replace these old objects from Props with ones from context
-    const missionFromProps = props.mission;
-
-    const takeControl = props.takeControl;
-    const deleteSingleMission = props.deleteSingleMission;
-    takeControlFunction = props.takeControl;
-    // End Old code
 
+export function BotDetails() {
```
JRoybot, line 61: YaY!  Looks so much better

src/web/containers/CommandControl/CommandControl.tsx
```diff
@@ -4227,17 +4227,6 @@ export default class CommandControl extends React.Component {
                     <NodeList />
                 </div>
```
JRoybot, line 4229: Why not render the new BotDetails container?

src/web/context/Jaia/JaiaContext.tsx
```diff
@@ -31,8 +31,6 @@ export interface JaiaContextType {
     hubs: Map<number, Hub>;
     missions: Map<number, Mission>;
 
-    clientID: string;
```
JRoybot, line 34: Will we need to add something new but similar in the future?

src/web/data/missions/missions.ts
```diff
@@ -46,7 +46,6 @@ class Missions {
         mission.setMissionID(missionID);
         this.setMissionIDInEditMode(missionID);
         this.setNextMissionID(this.getNextMissionID() + 1);
-        return missionID;
     }
```
JRoybot: leave the return in

JRoybot (changes requested):
Nice clean up job.  Always satisfying to throw out the trash :-)  Just need to put the return back in addMission.  A few other questions for clarification purposes.

src/web/containers/CommandControl/CommandControl.tsx
michael-jaia: CommandControl.tsx has too much going on with regards to how its rendering components. App.tsx will replace CommandControl.tsx without all of the baggage.

src/web/context/Jaia/JaiaContext.tsx
michael-jaia: Yes

src/web/data/missions/missions.ts
michael-jaia: Now that we are using the structure:
```
const newMission = new Mission();
const newMissionID = newMission.getMissionID();
```
Do we need the return?

JRoybot: yes to set the accordion

JRoybot (approved):

PR #1098, opened 2025-03-03
feature/add-configurable-pressure-sensor-readings/JAIA-1861

ferrom-jaia:
## Description

We needed a way to set the data rate for the pressure reading. To do this we added a configuration to the pressure sensor python driver. We also changed the pressure sensor driver to a push in terms of sending data. The c++ side does not need to send a query for information anymore. 

## Test
* Tested on flat fleet
* Ensured the data rate is what was set


ferrom-jaia (commented):
Tested with the flat fleet.

PR #1102, opened 2025-03-05
Workaround graphviz bug

tsaubergine:
This bug in Graphviz https://gitlab.com/graphviz/graphviz/-/issues/1408 is now hitting us (it's fairly random).

This only effects 'spline=ortho' which is the best looking way to render the Goby pub/sub graphs, but not if it doesn't work.

I've changed this to 'spline=line' which is a bit harder to read but will work until we are using a patched Graphviz (~2.y~ 3.y?)


ferrom-jaia (approved):
Nice find!

PR #1104, opened 2025-03-15
Feature/add logic to start camera in state machine jaia 1914

edsanville:
Configuration options were added to the jaiabot_mission_manager:
A `start_camera_command` of type `CameraCommand`, which can be configured to start recording video or images.
A `stop_camera_command` of type `CameraCommand`, which can be configured to stop recording video or images.

When the mission manager enters the `DivePrep` state, it sends the `start_camera_command`, and when it enters the `UnpoweredAscent` state, it sends the `stop_camera_command` to the driver.


src/python/camera/camera_driver.py
```diff
@@ -0,0 +1,117 @@
...
+        self.last_image_capture = 0.0
+        self.videoprocess = None
+
+        self.output_dir = datetime.datetime.now().strftime('%Y-%b-%d')
+        os.makedirs(self.output_dir, exist_ok=True)
+
+    def do_command(self, command: CameraCommand):
+        log.info(command)
+
+        if command.type == CameraCommand.CameraCommandType.START_IMAGES:
```
JRoybot, line 72: Would a switch be more efficient and easier to read?

JRoybot (dismissed):
Nice Job on this Ed, gave it a thorough read.  Didn't really see anything out of place.

src/python/camera/camera_driver.py
edsanville: Python doesn't have `switch`, although newer versions have `match` from version 3.10.  It might be a good idea to switch to `match` when we migrate from python 3.8.

JRoybot: or use if ... elif .. elif ... else

edsanville: Good point on the elif!

JRoybot (dismissed):
All looks good to me.  Nice job guys!

ferrom-jaia:
Closing as we merged into development branch for bio

PR #1115, opened 2025-03-20
bug/1.y/Remove known hosts on major upgrade 

tsaubergine:
Avoids issue where the hub can't reconnect on SSH after it is upgraded but some of the bots weren't,  after these  bots are upgraded.


ferrom-jaia (approved):
Looks great!

PR #1116, opened 2025-03-20
Add check to ensure new fleet config ssh key matches.

tsaubergine:
If the hub doesn't have the same ssh key on major upgrade, it won't be able to connect to any un-upgraded bots.

This new check ensures that the hub keys is the same in the fleet configuration, or else refuses to upgrade (the solution here would be to fix the fleet.cfg to match the existing hub key).

Moved version checking tasks to their own file at config/ansible/major_upgrade/tasks/check-versions.yml

Remove "run_once: true" to ensure hub fails if any bot fails, so we don't go to the reboot-check splash screen when we shouldn't.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -188,6 +188,7 @@ filter-template-debian-builds: &filter-template-debian-builds
     branches:
       only:
         - "1.y"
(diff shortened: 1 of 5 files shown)


tsaubergine:
Superceded by https://github.com/jaiarobotics/jaiabot/pull/1123

PR #1118, opened 2025-03-20
task/1.y/Increase number of fleets to 250

tsaubergine:
Our max is 255 fleets (IPv4 octet). I made it 250 for a round number.

Also, added option to specify an existing hub private key when creating a fleet config (used when upgrading an existing fleet).


ferrom-jaia (approved):
Looks great!

PR #1122, opened 2025-03-21
Task/2.y/refactor major upgrade

tsaubergine:
Exact cherry-pick of https://github.com/jaiarobotics/jaiabot/pull/1123 with the exception of two new additions:
- Remove hub .ssh/known_hosts so that bots that are upgraded by a 2.y hub can still be connected to after upgrade.
- `chmod 0600 {{ hub_staging_dir }}/jaiabot/init/{{ hub_privkey }}` which is required by SSH to avoid an error.

This also omits the reuse of the old hub ssh key that is now in https://github.com/jaiarobotics/jaiabot/pull/1123 with the assumption that the 2.y-3.y upgrade will reuse the same Yubikeys.

This needs to be the same as 1.y so that 2.y hubs can upgrade 1.y bots.

Should be merged with https://github.com/jaiarobotics/jaiabot/pull/1123


ferrom-jaia (approved):
Looks good! Tested with fleet7.

PR #1123, opened 2025-03-21
Task/1.y/refactor major upgrade

tsaubergine:
Minor refactor of functionality of major_upgrade ansible playbook based on testing this week.

Split playbook into two playbooks:
- hub staging: just the hub
  - All tasks that used to be "delegate_to: localhost"
  - By move this to a hub-only playbook, we avoid chances for race conditions when previously all bots would delegate the same tasks to the same hub
  - Add working check for SSH key
     - New option to `jaia-generate-first-boot.py` (aka `jaia admin fleet generate`): `--hub-ssh-keys-only` which only outputs the hub keys from the fleet configuration. This is used by this ansible task.
- actual upgrade: all nodes (all bots and hub)

I also added an output variable in JCU `major_upgrade_error` that gives more useful error feedback to the user for common issues.

I tested this on a Virtualfleet with 10 bots and 3 hubs, in various permutations of upgrading bots, then upgrading more bots, then switching hubs and upgrading more bots.

Updated `config/ansible/major_upgrade/tasks/hub-check-ssh-key.yml` to persist old private keys on the hub to "reuse" and then restore them to .ssh/oldkeys after upgrade, and add them to .ssh/config

Should be merged with https://github.com/jaiarobotics/jaiabot/pull/1122


ferrom-jaia (approved):
Looks good! Tested with fleet7.

PR #1125, opened 2025-03-31
Bug/2.y/docker simulator metadata mac/jaia 1929

JRoybot:
The Jaiabot Docker Simulator was not running properly on newer Mac computers.  Users reported that bots would all go unhealthy after a few seconds due to missing metadata.  Investigation found the root cause was the simulation was using rosetta2 to run in emulation mode and was running very slowly.  

To support the ARM64 architecture without having to deal with differences between the MacOS and Linux environments, the ARM64 debian packages already being built by circleci for the Raspberry Pi were used to create the docker image rather than building everything from source.  

This provided several benefits
- Create images built to run on both ARM and AMD architectures
- Do not have to support development in MacOS
- Automatically uploads to Dockerhub with new releases of jaiabot software
- Smaller images that are faster to transfer 
- Faster startup of JCC in running containers
- Does not require users to install the jaibot git repo to use the simulator

Summary of changes

- Modified launch scripts to use soversioned libraries
- Modified launch scripts to run liason directly without the `goby_liaison_jaiabot` script
- Removed all references to the `goby_liaison_jaiabot` script
- Rewrote Dockerfile.in to install debian packages instead of building from source
- Updated sim-docker support scripts to work with the new image format
- Updated circleci config to create both architecture specific images and push to Dockerhub
    - Need to remove "bug/2.y/docker-simulator-metadata-mac/JAIA-1929" from `filter-template-debian-builds` & `filter-template-docker-sim-builds` after merge
- Updated README file 

Testing
-Pushes to github triggered circleci builds each step of the way
-Ran images created by circleci on both an M1 Mac and an Ubuntu 22.04 system and verified simulation ran properly
-Did complete build and ran things locally on Ubuntu 24.04 VM
-Ran support scripts to  generated images locally and run them on Ubuntu 22.04


tsaubergine:
Perhaps we should move the README.md to a named page in src/doc/markdown, so it shows up in https://docs.jaia.tech/?

scripts/sim-docker/entrypoint.sh
```diff
@@ -1,11 +1,11 @@
...
 export JAIA_REST_API_PRIVATE_KEY=simulation
-cd /jaiabot/build/amd64/share/jaiabot/web/server/
-source /jaiabot/build/web_dev/python/venv/bin/activate
+cd /usr/share/jaiabot/web/server
+source /usr/share/jaiabot/python/venv/bin/activate
 ./app.py &
 cd /jaiabot/build/amd64/share/jaiabot/web/rest_api/
-./app.py -e 1:localhost:40000 &
-cd /jaiabot/config/launch/simulation/
+./app.py localhost &
```
tsaubergine: Should read something like:

```
export JAIA_REST_API_PRIVATE_KEY=simulation
source /usr/share/jaiabot/python/venv/bin/activate
// ...
cd /usr/share/jaiabot/web/rest_api
./app.py &
```

tsaubergine: Specifically the path is wrong (`/usr/share/jaiabot/web/rest_api` not `/jaiabot/build/amd64/share/jaiabot/web/rest_api/`) and there's no `localhost` argument

tsaubergine: To test just open http://localhost:9092/jaia/v1/status/all - you'll get all the bot/hub statuses in JSON

JRoybot: > export JAIA_REST_API_PRIVATE_KEY=simulation
> source /usr/share/jaiabot/python/venv/bin/activate
> cd /usr/share/jaiabot/web/rest_api
> ./app.py &
> ```

I tried these commands at the command line of a running container before committing any changes.
I got the following response 

```
(venv) root@18eaae2f07a0:/usr/share/jaiabot/web/rest_api# ./app.py &
[1] 6925
(venv) root@18eaae2f07a0:/usr/share/jaiabot/web/rest_api# WARNING:root:The configuration file /etc/jaiabot/rest_api.pb.cfg was not found. Using command line parameters only.
WARNING:root:no ip specified, using HUB 1 at localhost:40000
Starting endpoint for Jaia streaming protocol for Hub: 1
```
http://localhost:9092/jaia/v1/status/all 
returned
"Forbidden
You don't have the permission to access the requested resource. It is either read-protected or not readable by the server."

Not sure what else is needed

tsaubergine:
We have documentation for a reason: See
https://docs.jaia.tech/md_page12_rest_api.html

http://localhost:9092/jaia/v1/status/all?api_key=simulation

ferrom-jaia (approved):
Looks great! Tested the image by pulling it down from dockerhub. Good job @JRoybot and thank you @tsaubergine for the guidance for this task!

PR #1126, opened 2025-03-31
task/cleanup-crc-check-code-sub-sensor-app

michael-jaia:
### Summary
Removes byte to hex conversions that arose from debugging and initial development. Now we rely on the `std::string` that comes from accessing the protobuf type `bytes` in C++ when passing data to the CRC functions. This PR also removes the unused CRC function in the sensor app.

### Testing
Successfully sent `SensorRequest` messages and received `SensorData` messages between the Pi and STM32.

image


ferrom-jaia (approved):
Looks good!

PR #1129, opened 2025-04-07
feature/deploy-to-stm32-without-debugger

michael-jaia:
### Overview
To deploy code to the STM32 MCU without using the ST-Link debugger, copy the directory `stm32-deploy` onto a vehicle. Prior to running `deploy.sh`, you will need to move a `JAIA_BIO-PAYLOAD.elf` file from the STM32CubeIDE inside this directory.

### Testing
In the top terminal we see the `jaiabot_sensors` application send the `ENTER_BOOTLOADER_MODE` command to the STM32. We know that command is handled properly because in the bottom terminal we reach 100% for the the write and verify. When the final reset of the STM32 occurs, I see the lights on the payload board flash as expected.

stm32-deploy-test


ferrom-jaia (approved):
Looks good. Saw Twomey testing this.

PR #1130, opened 2025-04-07
Update HelpWindow.tsx

Kait211:
I am making changes to the help window in the sim.

I Just fixed the issue of pushing with terminal. I was correct in my theory of needing an ssh key, I now know how to push on terminal and wrote notes on the process so am not stuck on this again.


ferrom-jaia:
Hey Kaitlyn,

I am closing this for now. The JCC Refactor is almost complete. We will add this back in after it is done. Thank you for putting the effort in!

PR #1131, opened 2025-04-08
feature/2.y/docker-sim-jdv/JAIA-1951

JRoybot:
Add support for JDV in the JAIA Docker Simulation

Summary of changes

- Added dev branch to cirlceci config filters
  - remove before merging
- Updated sim image
  - installed rsync to support JDV
  - added ENTRYPOINT so it isn't needed at run time
 - Created docker-compose file to manage all of the options needed to launch the sim container
 - eliminated obsolete sim_env_vars.txt file
 - Updated documentation

Testing

- circleci updated images with every push
- ran simulation on linux and mac to verify JCC, JDV and REST API
  - ran missions
  - downloaded data from hub
  - viewed data in JDV
  - confirmed data folder is visible on host machine


edsanville (dismissed):
Code looks good, and the jdv image runs on both my Ubuntu 22.04 machine and my macOS 15.3.1 with M1 processor.

src/doc/markdown/page121_using_simulator_docker.md
```diff
@@ -41,24 +45,63 @@ As of 2.y release the Jaiabot Docker Simulation images are generated for both AM
 
 ## Run the Simulator
 
-### Set the simulation run parameters
+### Using docker compose
+Docker compose provides a platform-independent way to configure and run docker containers.  The easiest way to launch the JAIA Simulator in docker is to use a `docker-compose.yml` file.  The user can download a pre-defined file from `scripts/sim-docker/docker-compose.yml` at https://github.com/jaiarobotics/jaiabot
```
ferrom-jaia: The link should point to the docker compose yaml

src/doc/markdown/page121_using_simulator_docker.md
```diff
@@ -41,24 +45,63 @@ As of 2.y release the Jaiabot Docker Simulation images are generated for both AM
...
-JAIA_SIM_BOTS=3
-JAIA_SIM_WARP=2
-JAIA_SIM_FLEET=20
-```
-The following command will launch the Jaiabot Simulator in a Docker container.  This can be run from any directory but make sure `sim_env_vars.txt` is in that directory.
+From the same directory as the `docker-compose.yml` file the following command will shut down the simulator, stop the container and remove it.
+
+`docker compose down`
+
+### Using docker run
```
ferrom-jaia, line 178: Is there any reason to keep instructions for running it like this? I am not sure if this will make it confusing for some.

src/doc/markdown/page121_using_simulator_docker.md
```diff
@@ -68,32 +111,34 @@ Explanation of command.
...
 
-## Accessing JCC
+## Using JCC
 
 At this point the simulation is up and running in the docker container and the user simply needs to open a browser and open the following URL
 
 http://localhost:40001/
 
-## Stop the Simulator
+## Using JDV
```
ferrom-jaia: We need a section for accessing the rest api (just give an example address or point to the documentation: http://52.36.157.57/md_page12_rest_api.html)

ferrom-jaia (changes requested):
Looks good! Just a couple of things to review.

scripts/sim-docker/launch-container.sh
```diff
@@ -1,3 +1,15 @@
...
+
+if [[ -z "$JAIA_SIM_WARP" ]]; then
+    export JAIA_SIM_WARP=3
+fi
+
+if [[ -z "$JAIA_SIM_FLEET" ]]; then
+    export JAIA_SIM_FLEET=30
+fi
+
+docker run --rm --name jaia-sim-container -d -i -t -p 40001:40001 -p 9092:9092 -p 40011:40011 --env JAIA_SIM_BOTS=$JAIA_SIM_BOTS --env JAIA_SIM_WARP=$JAIA_SIM_WARP --env JAIA_SIM_FLEET=$JAIA_SIM_FLEET -v ./jdv_data:/var/log/jaiabot/bot_offload jaiauser:jaia-sim-image /bin/bash -li "/entrypoint.sh"
```
JRoybot: I will update this, the entrypoint is no longer needed

src/doc/markdown/page121_using_simulator_docker.md
JRoybot: Is the comment because the file is named ".yml" instead of ".yaml" or is it because you want a direct link to the file in github?

If the latter, I did not do that for 2 reasons.  The user might have a clone of the repo so wouldn't need the URL, just the path.  I the file only exists on this dev branch now so there is no existing permalink to point to.

src/doc/markdown/page121_using_simulator_docker.md
JRoybot: Will move to an "Advanced" appendix

src/doc/markdown/page121_using_simulator_docker.md
JRoybot: Will do

src/doc/markdown/page121_using_simulator_docker.md
ferrom-jaia: Yes, I was thinking the latter we could just use what the path will be: https://github.com/jaiarobotics/jaiabot/tree/2.y/scripts/sim-docker/docker-compose.yml

Some of our customers have never used github before so I want to make it as easy as possible.

ferrom-jaia (approved):
Looks great! Tested on my machine (windows and linux env)

PR #1132, opened 2025-04-08
task/add-udev-rules-for-bio

michael-jaia:
## Summary
Distinguish between Arduino and BIO USB devices by evaluating the vendor and product IDs. The Arduino device lives at `/dev/arduino` and the BIO payload device lives at `/dev/bio-payload`. 

## Testing
* Applied the udev rules to to Fleet 0 Bot 2. 
* Ran `sudo udevadm control --reload-rules && sudo udevadm trigger`
* Connected and disconnected Arduino and BIO Payload USBs while viewing the list of devices

**Testing Matrix**

image

## Considerations
Will the Arduinos of older JaiaBots have the same vendor and product ID? A 2023 vehicle did have the same identification. The vendor and product ID comes from the USB to serial UART chip which on the Arduino Nano is the FT232R from Future Technology Devices International (https://ftdichip.com/wp-content/uploads/2020/08/DS_FT232R.pdf). As long as this chip does not change, we will not have a discovery problem. The vendor and product ID for this chip can be found on page 30 of the linked datasheet.

P23069-L23046

### DIFF
--- a/config/gen/bot.py
+++ b/config/gen/bot.py
@@ -50,7 +50,7 @@ if "jaia_arduino_type" in os.environ:
 if jaia_arduino_type == "spi":
     jaia_arduino_dev_location="/dev/ttyAMA1"
 elif jaia_arduino_type == 'usb':
-    jaia_arduino_dev_location="/dev/ttyUSB0"
+    jaia_arduino_dev_location="/dev/arduino"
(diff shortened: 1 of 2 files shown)


ferrom-jaia (approved):
Looks good!

PR #1137, opened 2025-04-10
task/publish-ph-do-temperature-readings

michael-jaia:
### Summary
The message definitions for pH, dissolved oxygen, and pressure/temperature already included a temperature field. We just needed to add the temperature data to the messages being published.


ferrom-jaia (approved):
Static review looks good!

PR #1138, opened 2025-04-10
task/hide-mcu-raw-data-from-goby-logs

michael-jaia:
### Summary
Hides raw MCU data from Goby logs to prevent confusion that occurs when visualizing the oneof field in JDV.


ferrom-jaia (approved):
Looks good!

PR #1139, opened 2025-04-10
task/update-ec-driver-to-reflect-optional-fields

michael-jaia:
## Summary
The AtlasScientificOEMEC message fields changed from required to optional. We reflect this change in the driver by checking what fields the incoming message contains.


ferrom-jaia (approved):
Looks good!

PR #1141, opened 2025-04-10
task/reset-stm32-when-starting-jaiabot-sensors-service

michael-jaia:
### Summary
When the `jaiabot` services or `jaiabot_sensor` service restarts, we want to reset the BIO payload board to clear any stale data that could cause problems transmitting sensor data. This pull request adds a reset script that toggles the USB bus 1 on/off. This script is called with the `ExecStartPre` hook of the `jaiabot_sensors` systemd service configuration.

Based on my Internet investigation, a USB connected device is assigned to a USB bus based on its speed characteristics. This is important for the reset script because it allows us to hardcode the location of the USB in `/sys/bus/usb/devices/`.

### Testing
* Moved `systemd.py`, `reset-bio-payload-board.sh`, and `goby-app.service.in` to Fleet 0 Bot 2.
* Ran `sudo dpkg-reconfigure jaiabot-embedded`
* Ran `sudo systemctl restart jaiabot` --> BIO board resets
* Ran `sudo systemctl restart jaiabot_sensors` --> BIO board resets

### Helpful Resources
* https://stackoverflow.com/questions/52501585/how-to-set-bus-number-of-usb-device-in-linux
* https://www.linux.org/threads/linux-usb-port-guide-speeds-and-power.51470/


ferrom-jaia (approved):
Looks good. /usr/bin is hardcoded because that is the default location our scripts go regardless of apt installing them or deploying.

PR #1143, opened 2025-04-11
task/add-voltage-to-ph-do-and-fluorometer-messages

marshalllele:
This branch changes the name of the fluorometer group from "concentration" to "fluorometer" to better reflect the actual sensor. It also adds logging for the pH and DO temperature raw voltages, as well as the fluorometer concentration's raw voltage. These voltage values come from the STM32's 12 bit ADC.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -188,6 +188,7 @@ filter-template-debian-builds: &filter-template-debian-builds
     branches:
       only:
         - "1.y"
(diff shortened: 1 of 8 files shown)


marshalllele:
[Untitled.mov.zip](https://github.com/user-attachments/files/19713753/Untitled.mov.zip)

Tested on flat bio with matching payload board code deployed

michael-jaia:
Code Review Testing Results:
image

ferrom-jaia (approved):

PR #1144, opened 2025-04-14
task/show-errors-for-failed-bio-sensor-initializations

michael-jaia:
### Overview
When a sensor on the BIO payload board does not initialize successfully, a warning will appear in the JCC to alert the operator.

### Implementation
1. A failed initialization will result in the FAILED value living in the `Sensors` array on the STM32 for the particular sensor.

2. When the metadata is transmitted for each sensor, if it is in a `FAILED` state, then the `Metadata` message will contain the field `init_failed` which will be set to true.

3. The `jaiabot_sensors` application will process the incoming metadata. If it comes across a message that has the `init_failed` field set to true, it will insert the sensor into the `failed_initializtions` set.

4. The overridden `health` function will add a warning to the health thread for the sensor that failed. For example, `jaiabot::protobuf::WARNING__INIT_FAILED__TURNER__C_FLUOR.`

5. This addition will be picked up by the JCC code and display in the Bot details panel.

### Testing
**Bar30 Failed Initialization**
bar30-failed-initialization

**All Sensors Failed Initializations**
all-sensor-failed-initializations

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -188,7 +188,6 @@ filter-template-debian-builds: &filter-template-debian-builds
     branches:
       only:
         - "1.y"
-        - "task/show-errors-for-failed-bio-sensor-initializations"
 
 # which branches to run the normal source build
 filter-template-normal-builds: &filter-template-normal-builds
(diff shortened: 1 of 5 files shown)


michael-jaia:
### Additional Testing
**Baseline**
baseline


**Disconnected Bar30**
disconnect-bar30


**Connected Bar30**
connect-bar30


**Triggered Atlas DO Init Failure**
trigger-atlas-init-fail

PR #1145, opened 2025-04-14
feature/add-error-handling-for-missing-sensor-data-mcu

ferrom-jaia:
## Description

This will add warnings for missing ec, ph, conductivity, and fluorometer data. It will add a error for missing bar30 data. If we detect missing data a sensor request message will be sent on a configurable rate (default 20 seconds). 

## Testing
1. Power on BIO payload board. All sensors are reporting.
2. Disconnect BIO payload board. Sensor values freeze in Liaison.
3. After 20+ seconds of no new sensor data, warnings and errors appear in Liaison. (The time can vary because of the `jaiabot_sensors` application update frequency).
4. Reconnect BIO payload board. Sensor values start to update as new cfg messages are received on the payload board. The warnings and errors are removed in Liaison.

### DIFF
--- a/src/bin/sensors/config.proto
+++ b/src/bin/sensors/config.proto
@@ -32,30 +32,35 @@ message AtlasOEMECThreadConfig
 {
     optional int32 sample_rate = 1 [default = 10];
     optional int32 report_timeout_seconds = 2 [default = 20];
+    optional int32 trigger_cfg_resend_timeout_seconds = 3 [default = 20];
 }
 
 message BlueRoboticsBar30ThreadConfig
 {
     optional int32 sample_rate = 1 [default = 10];
     optional int32 report_timeout_seconds = 2 [default = 20];
+    optional int32 trigger_cfg_resend_timeout_seconds = 3 [default = 20];
 }
 
 message AtlasOEMPHThreadConfig
 {
     optional int32 sample_rate = 1 [default = 10];
     optional int32 report_timeout_seconds = 2 [default = 20];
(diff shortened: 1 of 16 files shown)


src/bin/sensors/config.proto
```diff
@@ -28,10 +28,49 @@ import "goby/middleware/protobuf/serial_config.proto";
 
 package jaiabot.config;
 
+message AtlasOEMECThreadConfig
+{
+    optional int32 sample_rate = 1 [default = 10];
+    optional int32 report_timeout_seconds = 2 [default = 20];
+    optional int32 trigger_cfg_resend_timeout_seconds = 3 [default = 20];
```
michael-jaia: Do you like better resend_cfg_timeout_seconds?

src/bin/sensors/config.proto
```diff
@@ -28,10 +28,49 @@ import "goby/middleware/protobuf/serial_config.proto";
...
 message Sensors
 {
     optional goby.middleware.protobuf.AppConfig app = 1;
     optional goby.zeromq.protobuf.InterProcessPortalConfig interprocess = 2;
     optional goby.middleware.protobuf.SerialConfig mcu_serial = 3;
-    optional int32 sample_rate = 4;
+    optional AtlasOEMECThreadConfig ec = 10;
+    optional BlueRoboticsBar30ThreadConfig bar30 = 20;
+    optional AtlasOEMPHThreadConfig ph = 30;
+    optional AtlasOEMDOThreadConfig dissolved_oxygen = 40;
```
michael-jaia, line 74: Should we use `do` here?

PR #1148, opened 2025-04-16
bug/missing-data-vision-system-config

ferrom-jaia:
## Description

Add missing config for the jaiabot data vision app.


marshalllele (approved):
Tested on F45H01, everything looks good!

PR #1149, opened 2025-04-17
task/add-config-option-for-camera-jaia-1919

edsanville:
I added a `CameraType` enum for the `jaiabot_mission_manager` config. This can be `NO_CAMERA` or `AFT`.

The `jaiabot_mission_manager` only sends camera commands if `cfg().camera_is_present()` is `true`, and the start/stop `CameraCommand` has also been defined.


JRoybot (dismissed):
All code changes looks good to me.  Did not build or run changes, unfamiliar with the base branch efforts

JRoybot (approved):

PR #1155, opened 2025-04-20
feature/add-temperature-compensate-salinity-jaia-1852

edsanville:
Translated the calibrated salinity function into C++ and used it to calculate the `salinity` field in `BotStatus`.  The `raw_salinity` field holds the salinity as reported by the sensor chip itself.


src/lib/utils/calibrated_salinity.h
```diff
@@ -0,0 +1,68 @@
+#include <math.h>
+
+double calculate_calibrated_salinity(const double measured_conductivity, const double temperature,
```
ferrom-jaia: I think we need unit test for this.

src/lib/utils/calibrated_salinity.h
```diff
@@ -0,0 +1,68 @@
...
+    const double dS = (t - 15) *
+                      (b0 + (b1 * sqrt(Rt)) + (b2 * Rt) + (b3 * pow(Rt, 1.5)) + (b4 * Rt * Rt) +
+                       (b5 * pow(Rt, 2.5))) /
+                      (1 + k * (t - 15));
+
+    const double S = a0 + (a1 * sqrt(Rt)) + (a2 * Rt) + (a3 * pow(Rt, 1.5)) + (a4 * Rt * Rt) +
+                     (a5 * pow(Rt, 2.5)) + dS;
+
+    return S;
+}
```
marshalllele: The calculate_calibrated_salinity() function aligns with what I've found in research as well - looks good!

src/lib/utils/calibrated_salinity.h
marshalllele: After some conversations with Jason and looking into the values we're actually getting from the Atlas Scientific chip, I think that calculate_derived_salinity() would be a better name for this function.

marshalllele (changes requested):

src/lib/utils/calibrated_salinity.h
edsanville: Added a Boost unit test to the `src/tests/utils` directory.

ferrom-jaia (approved):
Looks great! Tested against this calculator: https://www.physocean.icm.csic.es/Utilities/calculators/salcon-en.html

PR #1158, opened 2025-04-24
task/multiple-crc-files-in-lib-jaia-1968

edsanville:
* Moved the fletcher16 checksum function to `src/crc/fletcher16.h`.  This checksum is used in the Arduino comms.
* Moved the crc32 checksum to `src/crc/crc32.h`.  This checksum is used in the Pi Zero comms for the camera.
* Deleted the unused crc-16 code, which was from an failed iteration of the Arduino comms.  The Arduino doesn't have enough memory for the crc-16 tables.
* Renamed the mis-named SerialThreadCRC32 class to SerialThreadFletcher16, because it uses Fletcher-16 checksums.
* Moved the SerialThreadCRC32 and SerialThreadFletcher16 classes to `src/lib/serial`.


ferrom-jaia (approved):
Looks good!

PR #1160, opened 2025-04-30
Fix 2.y AWS upgrades to use correct destination repo (release, beta, etc.)

tsaubergine:
Previously rootfs/cloud/aws/packer/packer-create-ami-by-upgrade.sh was only creating output AWS AMIs that had the `test` repo enabled.  This meant even when I created a "continuous" repo Cloudhub, I got one with the `test` Debian repo enabled.

This was an oversight, which is fixed here.

Tested that this still build correctly on `test`. Obviously I can't test other repos until it is merged.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -178,6 +178,7 @@ filter-template-debian-builds: &filter-template-debian-builds
     branches:
       only:
         - "2.y"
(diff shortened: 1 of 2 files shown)


ferrom-jaia (approved):
Nice! Looks good!

PR #1161, opened 2025-05-01
feature/2.y/iridium/jaia-1972/jaia-2021

tsaubergine:
## Changes

- Added configuration for SBD Iridium support using the existing Goby drivers. Tested in simulation (requires https://github.com/tsaubergine/virtual_iridium/).
- Added framework for launching "specialty" AWS servers like iridium.jaia.tech (create_specialty_server.sh). I will expand to cover (and replace) vpn.jaia.tech, packages.jaia.tech in the future so that we have a reproducible way to make these servers. I will also use the framework to launch the new cloud.jaia.tech login server (see https://github.com/jaiarobotics/jaiabot/pull/1170).
- Changed Cloudhub to be one of two types:
  - Secondary: Existing Cloudhub types where JCC connects directly to a real hub and most of the hub code isn't run on the Cloudhub machine
  - Primary: Much closer to a real hub, just not with firmware code. This is selected if any comms modes exists for Cloudhub after removing the local ones (xbee, wifi). At the moment this would be only if "iridium" is in use, but will gracefully expand later to support other remote comms links such as LTE/cellular.
- Changed Cloudhub's create_vpc.sh's configuration file (vpc.conf) to use a fleet configuration file (FLEET_CONFIG). This allows Primary Cloudhubs to correctly provision hub SSH keys. The SSH_PUBKEYS variable in vpc.conf is removed as this information is now read from the fleet configuration file.
- Updated the fleet configuration file to include a mapping of IMEI to bot IDs for Iridium. This gets installed on the hub as `/etc/jaiabot/iridium.json`, which is then read by the Python configuration generation code. This also includes credentials for connecting to RockBLOCK (if in use).
- We can use either the RockBLOCK (https://jaia-innovation.atlassian.net/browse/JAIA-2021) or Iridium DirectIP (MetOcean) service for a given fleet, but not mix them within a fleet.
 
## Testing performed

- Tested in simulation
- Tested with bot 1 fleet 6 and cloudhub fleet 6
  - Still need more data on performance of Iridium modem (observed frequent low CIEV values)
- Tested at Jaia in-water in July
- Tested during Silent Swarm

## Other

- Updated to [Goby 3.3.0](https://github.com/GobySoft/goby3/releases/tag/3.3.0) (for IPv6 support on Iridium RockBLOCK shore driver). **This version of Goby is currently in test/2.y *staging* on packages.jaia.tech, so the mirror will need to be updated to use this version on continuous when (just before) this PR is merged.**


config/gen/common/comms.py
```diff
@@ -64,9 +62,58 @@ def wifi_hub_remotes(this_node_id, fleet_index):
...
 
 def xbee_mac_slots(node_id):
-    slots = 'slot { src: ' + str(xbee_modem_id(node_id)) + ' slot_seconds: 0.1 }\n'
+    slots = 'slot { src: ' + str(modem_id("xbee", node_id)) + ' slot_seconds: 0.1 }\n'
+    return slots
+
+def iridium_mac_slots(node_id):
+    # SBD is rate 0 in the Goby driver
+    sbd_rate=0
+    slots = 'slot { src: ' + str(modem_id("iridium", node_id)) + ' slot_seconds: 15 rate: ' + str(sbd_rate) + ' [goby.acomms.iridium.protobuf.transmission]: { if_no_data_do_mailbox_check: false } }\n'
```
ferrom-jaia, line 75: What was the thought behind making the slot_seconds 15 seconds?

config/gen/common/comms.py
```diff
@@ -64,9 +62,58 @@ def wifi_hub_remotes(this_node_id, fleet_index):
...
+    return slots
+
+def iridium_shore_mac_slots(node_id):
+    sbd_rate=0
+    # no reason to really rate limit commands substantially
+    slots = 'slot { src: ' + str(modem_id("iridium", node_id)) + ' slot_seconds: 5 rate: ' + str(sbd_rate) + ' }\n'
```
ferrom-jaia, line 81: Can you explain what the rate value does?

config/gen/hub.py
```diff
@@ -163,11 +213,14 @@
...
     this_hub='hub'+ str(hub_index) +'-fleet' + str(fleet_index)
     inventory='/etc/jaiabot/inventory.yml'
-    if hub_index == cloudhub_index:
+    if is_cloudhub:
         vfleet_playbooks=config.template_substitute(templates_dir+'/hub/_liaison_prelaunch_vfleet_playbooks.pb.cfg.in')
     else:
         vfleet_playbooks=''
-        
+
+    limit=''
```
ferrom-jaia, line 221: What does this limit variable provide?

ferrom-jaia (commented):
Looks great! I had a couple of questions that I have listed in the review.

config/gen/common/comms.py
tsaubergine: Probably the realistic minimum between MO SBD packets (maybe could do 10 but that would be pushing it).

tsaubergine: We can see if we can push it lower but if we're doing 60s between the highest rate data right now there's no need.

config/gen/common/comms.py
tsaubergine: Every driver in Goby can support multiple data rates (integers) that typically correspond to some look up table (based on the driver) of the effective throughput (lower is typically more error resistant but lower throughput using more robust codes, and higher is more error prone but higher throughput). The Iridium protocol (at the time of writing this driver) supported SBD (message based protocol) and RUDICS (stream based protocol), so I assigned SBD rate 0 and RUDICS rate 1: https://github.com/GobySoft/goby3/blob/3.0/src/acomms/modemdriver/iridium_driver_common.h#L41-L45

tsaubergine: The modems we are using only support SBD so we just put the rate: 0 in the MAC cycle to tell the driver to use SBD

config/gen/hub.py
tsaubergine: limit gets passed to Ansible in Goby Liaison (JCU), and limits the playbook to be only the hosts in the limit string (pattern).

The default  (https://github.com/jaiarobotics/jaiabot/blob/2.y/src/lib/liaison/prelaunch/liaison_upgrade.cpp#L429) is to set limit to include the current Hub and all Bots (so that, e.g., hub 1 doesn't try to configure hub 2 which is never reachable on our current Wifi network). This means that we don't get unreachable errors for other hubs that just would confuse people.

Primary Cloudhubs can conceivable reach all Hubs and all Bots at once (if they're powered on) via the CloudHub VPN, so the limit is set to "all" in this case.

ferrom-jaia (approved):
Thank you Toby for the responses!

PR #1163, opened 2025-05-02
feature/2.y/jaia-simulator-starter-stopper-gui

ferrom-jaia:
## Description

This adds a gui to automate the manual tasks that the user would have to do to start/stop the jaia simulator.


ferrom-jaia:
Merging as a baseline.

PR #1165, opened 2025-05-02
feature/report-specific-conductivity

marshalllele:
## Description
Adds reporting of specific conductivity in addition to the currently reported raw conductivity.

For conductivity values below 8000 µS/cm, we use a standard linear temperature compensation model:
```
EC25 = ECt / (1 + 0.0191 * (T - 25))
```
Where:

* EC25 – Specific conductivity at 25 °C
* ECt – Measured raw conductivity
* T – Measured water temperature in °C

This linear model is commonly used for KCl-based calibration standards (e.g., 147, 706, and 1413 µS/cm), and performs well in that range.

However, for higher conductivity values, this model did not meet our target accuracy of <2% error in validation tests. To address this, we implemented a viscosity-based compensation model with an auto-tuned exponent b. This method uses empirical data to more accurately correct for temperature effects at high ion concentrations and provides significantly better performance across 10,000–100,000 µS/cm.

References:
 * https://www.aqion.de/site/112
 * https://pubs.usgs.gov/wsp/2311/report.pdf
 * https://link.springer.com/article/10.1023/B:EMAS.0000031719.83065.68 (Hayashi study, 2004)
 * https://www.hamiltoncompany.com/process-analytics/conductivity-knowledge/temperature-influence-on-conductivity-standards

## Testing
To run unit tests for specific conductivity:
```
./build.sh --enable-testing
ctest --test-dir build/arm64/ --output-on-failure --verbose
```

## Note
You must also merge the following pull request to enable this feature in the firmware:
https://github.com/jaiarobotics/stm32/pull/26


src/bin/sensors/drivers/atlas_scientific__oem_ec.h
```diff
@@ -50,6 +52,7 @@ class AtlasScientificOEMECDriver
     int32_t sample_rate_{10};
     int32_t report_timeout_{20};
     int32_t resend_cfg_timeout_{20};
+    jaiabot::protobuf::PressureTemperatureData last_pressure_temperature_data_;
```
edsanville: Looks like this ivar never gets used.

src/lib/messages/sensor/atlas_scientific__oem_ec.proto
```diff
@@ -11,6 +11,7 @@ message AtlasScientificOEMEC
     };
 
     optional double conductivity = 1;
+    optional double specific_conductivity = 2;
```
edsanville: It looks like this field doesn't get used (yet).  Should we do the `specific_conductivity` and `salinity_calculated` in the fusion app (for example), so those derived calculations are only done in one place?

src/lib/utils/specific_conductivity.h
```diff
@@ -0,0 +1,82 @@
+/**
+ * @brief Viscosity ratio: µ_t / µ_25 using Hayashi's equation (log base 10)
+ *
```
edsanville, line 20: Maybe we should include a link to the original source for this equation.

edsanville (dismissed):
Code looks good to me.

src/lib/messages/sensor/atlas_scientific__oem_ec.proto
ferrom-jaia: I see what you mean (Using the DRY method). It really comes down to the report speed. The Hydros are eventually going to start using the stm32 board which would deprecate the python drivers for the pressure and conductivity sensors.

scripts/util_helpers/test_specific_conductivity.py
```diff
@@ -0,0 +1,81 @@
...
+
+model = LinearRegression()
+model.fit(X_poly, y)
+
+# Show the results
+print("Intercept:")
+print(model.intercept_)
+print("\nCoefficients:")
+for name, coef in zip(poly.get_feature_names_out(["measured", "temp"]), model.coef_):
+    print(f"{name}: {coef:.5f}")
```
JRoybot: Name of this file is misleading... doesn't have any pass/fail criteria, what does it test?

JRoybot (dismissed):
All looks good.  One small comment about a file name.

scripts/util_helpers/test_specific_conductivity.py
ferrom-jaia: Yes, nice catch, I will update

PR #1168, opened 2025-05-07
feature/calibrate-bio-sensors-through-stm32

marshalllele:
Gives the ability to send calibration/compensation commands to the Atlas Scientific EC, DO, and pH chips via Liaison. 

To conduct a calibration:
1. Navigate to the bot's Liaison page
2. Go to Commander
3. In the "Message" dropdown, select jaiabot.sensor.protobuf.SensorRequest
4. Include a comment in the "Log comment" field
5. Include an integer in the "time" field
6. In the "calibration_type" dropdown, select "START_xx_CALIBRATION" to stop the data from being sent from all other Atlas chips, and update the current sensor to report at 1Hz
7. Press send
8. Under "calibration_type", select the type of calibration you'd like to perform. Include the value to calibrate to in the "calibration_value" field. 
9. Press send
10. Once all calibrations have been set for a chip, select "STOP_CALIBRATION" under the "calibration_type" field.
11. To set a compensation value, perform the same steps above under the "compensation_type" and "compensation_value" fields.

The BIO payload board will return the calibration confirmation of each chip when a calibration command is sent. These confirmation values can be verified against the datasheet for each OEM Atlas chip. 

Must be deployed in conjunction with https://github.com/jaiarobotics/stm32/tree/feature/calibrate-bio-sensors-through-stm32-for-stm32


ferrom-jaia (approved):
Looks good! Tested with Nick.

PR #1171, opened 2025-05-08
task/introduce-tasks-to-minimal-jcc/JAIA-1980

michael-jaia:
### Summary
This pull request brings back the waypoint panel so an operator can click on a waypoint, select a task, and modify the task parameters. The behavior of the task parameters is as follows:
1. Updating the task parameters of one task does not change the task parameters of another task.
2. Updating the task parameters of one task sets the default values for new tasks to be added.
3. When the task parameter values are saved as they are entered and checked against a min and or max.

### Implementation
task-parameters drawio

https://github.com/user-attachments/assets/f5f21749-1871-4251-b64b-000350ece20f

### Testing
1. Selecting a task renders the correct task parameters
2. Changing a task parameter does not change the task parameters of a previous task
3. Changing a task parameter sets the default for a new task

### Note
This is a bigger pull request than normal. It combines some of the sub tasks that were in progress prior to pausing the project.


src/web/context/JaiaContext.tsx
```diff
@@ -309,6 +335,69 @@ function handleAddWaypoint(mutableState: JaiaContextType, location: GeographicCo
     return mutableState;
 }
 
+/**
+ * Makes call to remove a waypoint from a mission
+ *
+ * @param {JaiaContextType} mutableState State object ref for making modifications
+ * @returns {JaiaContextType} Updated mutable state object
+ */
+function handleDeleteWaypoint(mutableState: JaiaContextType) {
```
JRoybot, line 359: There is an inconsistency between this pattern and `handleDeleteMission`.  
In `handleDeleteMission` the mission to be deleted is passed in as `missionID` based on the `.missionID` from the ` JaiaAction`    The action includes the `clickedWaypoint` attribute.  Is there a reason we are not putting the ID in the action and using it here instead of calling `getSelectedWaypoint()`

src/web/context/JaiaContext.tsx
```diff
@@ -469,6 +572,22 @@ function handleClickedPanelButton(mutableState: JaiaContextType, panelName: Pane
...
 
+/**
+ * Opens panel for the selected waypoint
+ *
+ * @param {JaiaContextType} mutableState State object ref for making modifications
+ * @param {SelectedWaypoint} clickedWaypoint Identifies which waypoint was clicked by operator
+ * @returns {JaiaContextType} Updated mutable state object
+ */
+function handleClickedWaypoint(mutableState: JaiaContextType, clickedWaypoint: SelectedWaypoint) {
+    jaiaGlobal.setSelectedWaypoint(clickedWaypoint);
```
JRoybot, line 626: shouldn't we check if the clicked waypoint is currently selected and deselect if it is?

src/web/types/jaia-system-types.ts
```diff
@@ -11,9 +11,39 @@ export interface SelectedNode {
     id: number;
 }
 
+export interface SelectedWaypoint {
+    waypointNum: number;
+    missionID: number;
```
JRoybot, line 21: a little concerned we have a slight circular reference because waypoints are inherently part of a mission in the data model and there is a chance of these IDs being out of sync with how the data in the data model is managed.  I presume this is included here because that relationship between the waypoint and a mission does not exist for the waypoint feature in the map?  Not sure of a better way to handle this but bringing it up to spur conversation....

src/web/context/JaiaContext.tsx
```diff
@@ -309,6 +335,69 @@ function handleAddWaypoint(mutableState: JaiaContextType, location: GeographicCo
...
+
+/**
+ * Updates the task associated with a waypoint based on the operator's selection
+ *
+ * @param {JaiaContextType} mutableState State object ref for making modifications
+ * @param {TaskType} taskType Name of the task selected
+ * @returns {JaiaContextType} Updated mutable state object
+ */
+function handleSelectTask(mutableState: JaiaContextType, taskType: TaskType) {
+    const waypoint = getWaypoint();
```
JRoybot, line 381: similar to comment for deleteWaypoint.... is there a reason we do not pass the waypoint in the action data?

src/web/context/JaiaContext.tsx
```diff
@@ -309,6 +335,69 @@ function handleAddWaypoint(mutableState: JaiaContextType, location: GeographicCo
...
+ *
+ * @param {JaiaContextType} mutableState State object ref for making modifications
+ * @param {TaskParameterPair} taskParameterPair The name of the input updated and its value
+ * @returns {JaiaContextType} Updated mutable state object
+ */
+function handleChangeTaskParameter(
+    mutableState: JaiaContextType,
+    taskParameterPair: TaskParameterPair,
+) {
+    const waypoint = getWaypoint();
```
JRoybot: same comment as above regarding waypoint

src/web/context/JaiaContext.tsx
```diff
@@ -320,6 +409,20 @@ function handleClosedDetails(mutableState: JaiaContextType) {
...
+/**
+ * Handles cleanup when a waypoint panel closes
+ *
+ * @param {JaiaContextType} mutableState State object ref for making modifications
+ * @returns {JaiaContextType} Updated mutable state object
+ */
+function handleClosedWaypointPanel(mutableState: JaiaContextType) {
+    jaiaGlobal.setSelectedWaypoint({ waypointNum: UNASSIGNED_ID, missionID: UNASSIGNED_ID });
+
+    mutableState.selectedWaypoint = jaiaGlobal.getSelectedWaypoint();
```
JRoybot, line 433: same comment regarding waypoint

src/web/data/tasks/task.ts
```diff
@@ -1,15 +1,37 @@
...
+import { jaiaGlobal } from "../jaia_global/jaia-global";
 
 export default class Task {
     private type: TaskType;
-    private diveParameters: DiveParameters;
-    private driftParameters: DriftParameters;
-    private constantHeadingParameters: ConstantHeadingParameters;
-    private stationKeepParameters: StationKeepParameters;
+
+    // Dive Parameters
```
JRoybot: I wish the data structure used for tasks enforced which parameters went with which TaskType, both missing paramaters for a task type and rogue parameters that should not be associated with a task type have been sources of errors that were hard to debug.  I had to write a very exhaustive tests for the previous TaskSettingsPanel to make sure it didn't produce bogus tasks.  I would like to discuss any options to reduce the potential for ill-formed tasks being created and saved.  Maybe use subclasses or unions?

src/web/data/tasks/task.ts
```diff
@@ -1,15 +1,37 @@
...
+    // Drift Parameters
+    private driftTime: number;
+
+    // Constant Heading Parameters
+    private heading: number;
+    private constantHeadingTime: number;
+    private speed: number;
+
+    // Parameter Constraints
+    MIN_DEPTH = 0;
```
JRoybot: Using MAX_DEPTH for the constraint is confusing because there is a task parameter of the same name.  Suggest using either "limit" or "constraing" in the names of these parameter constraints.

src/web/data/tasks/task.ts
```diff
@@ -18,40 +40,182 @@ export default class Task {
...
+     */
+    setParameter(taskParameterPair: TaskParameterPair) {
+        const key = taskParameterPair.key;
+        let value = taskParameterPair.value;
+        const mutableDefaults = { ...jaiaGlobal.getDefaultTaskParameters() };
+
+        switch (key) {
+            case TaskParameterKeys.MAX_DEPTH:
+                this.setMaxDepth(value);
+                mutableDefaults.maxDepth = value;
```
JRoybot: since `setMaxDepth` might override the `value` with the constraints I think this should be `this.maxDepth`

src/web/jcc/App.tsx
```diff
@@ -22,10 +28,49 @@ export default function App() {
...
+                <Panel />
             </JaiaContextProvider>
         </div>
     );
 }
+
+/**
+ * Controls the rendering of HubDetails and BotDetails
+ */
+function Details() {
```
JRoybot, line 41: Would this function be better off as a container with the BotDetails and HubDetails nested in the directory structure?

src/web/jcc/App.tsx
```diff
@@ -22,10 +28,49 @@ export default function App() {
...
+            return <BotDetails />;
+        default:
+            return <div></div>;
+    }
+}
+
+/**
+ *  Controls the rendering of JCC panels
+ */
+function Panel() {
```
JRoybot, line 61: Can this function get a more descriptive name?  "OverlayPanel" , "ActionPanel", "PopUpPanel"??

src/web/context/jaia-actions.ts
```diff
@@ -7,12 +7,17 @@ export enum JaiaActions {
     ASSIGN_MISSION = "ASSIGN_MISSION",
     AUTO_ASSIGN_MISSIONS = "AUTO_ASSIGN_MISSIONS",
     ADD_WAYPOINT = "ADD_WAYPOINT",
+    DELETE_WAYPOINT = "DELETE_WAYPOINT",
+    SELECT_TASK = "SELECT_TASK",
+    CHANGE_TASK_PARAMETER = "CHANGE_TASK_PARAMETER",
 
     CLOSED_DETAILS = "CLOSED_DETAILS",
+    CLOSED_WAYPOINT_PANEL = "CLOSED_WAYPOINT_PANEL",
```
JRoybot, line 15: Since we are going to be adding more panels, could we have a single action to close whatever transient panel is displayed?  Should be similar to how `CLICKED_PANEL_BUTTON` opens a panel.  We should add the panel name in the action data (or just use `jaiaContext.visiblePanel`, passing it is  more consistent with existing functions) 
 Again, would like to see the group of panels managed as "Panel" in the jcc app have a more descriptive name.

src/web/data/jaia_global/jaia-global.ts
```diff
@@ -1,27 +1,62 @@
...
+            this.selectedNode = clickedNode;
         }
     }
+
+    getSelectedWaypoint() {
+        return this.selectedWaypoint;
+    }
+
+    setSelectedWaypoint(clickedWaypoint: SelectedWaypoint) {
+        this.selectedWaypoint = clickedWaypoint;
```
JRoybot, line 56: I don't see anywhere in the call stack where the waypoint ID is checked to make sure it exists in the mission... not likely but what would happen if the selected waypoint got set to a waypoint that doesnt exist?

src/web/components/WaypointPanel/WaypointPanel.tsx
```diff
@@ -0,0 +1,137 @@
...
+                return "Station Keep";
+            case TaskType.CONSTANT_HEADING:
+                return "Constant Heading";
+            default:
+                return "";
+        }
+    };
+
+    const handleDeleteWaypointClick = () => {
+        jaiaDispatchContext({ type: JaiaActions.DELETE_WAYPOINT });
```
JRoybot, line 70: suggest including waypointID in the action data

JRoybot (commented):
I did a static review of the code.  It's a lot to absorb.  Made a few suggestions that may spur some discussions.  I will return to review again and build and test the branch before making a final approval or request for changes

src/web/context/JaiaContext.tsx
michael-jaia: Maybe down the road, but that is not how it is implemented in the current version of the JCC

src/web/data/tasks/task.ts
michael-jaia: Yes, good catch

src/web/jcc/App.tsx
michael-jaia: I like storing which panels render in App.tsx. I see App.tsx as the place for that type of control where the other directories should just worry about creating the component, not control when it renders.

src/web/jcc/App.tsx
michael-jaia: I left it general because all panels outside of the "details" panels will be included in the function. Of the three suggestions I like overlay panel the best, but I also feel like it doesn't add much description since all of the panels will be overlay panels.

src/web/data/jaia_global/jaia-global.ts
michael-jaia: I'm not sure if it is worth coupling the Mission class with JaiaGlobal. I like how JaiaGlobal just focuses on handling important UI data that comes from user interactions. That data should only come from the click event. If we get the wrong data part of me feels like we have bigger problems. This is just my opinion. Tell me if you think this is too loose. I could add docstrings to each of the functions to make this clear as well.

src/web/data/tasks/task.ts
michael-jaia: I thought the capitalization would distinguish them enough, but if that is not the case, we can use CONSTRAINT at the end of the variable names

src/web/data/tasks/task.ts
michael-jaia: I think we can find a way to do this. Let's talk this one over tomorrow.

src/web/context/JaiaContext.tsx
michael-jaia: The reason I did this is because context does not know which mission the operator is trying to delete unless data is passed via the action option. With a waypoint, only one waypoint can be deleted and that is the selected waypoint which is already stored in context. To me, passing the waypoint through the action here would be unnecessary since the data already lives in context. Feel free to push back on this

src/web/types/jaia-system-types.ts
michael-jaia: I'm not quite following the circular reference. I see the SelectedWaypoint as similar to a SelectedNode. It's not a core data model feature, it's more of a UI helper object.

src/web/context/jaia-actions.ts
michael-jaia: I think we can do this. I think this would be better as its own ticket.

src/web/context/JaiaContext.tsx
JRoybot: I understand the reasoning.  So in some cases the mission being deleted will not be the `missionIDInEditMode` stored in context?  I didn't consider that scenario.  My concern is there are multiple sets of data defining what mission is being edited and was concerned that a race condition may exist where they are not in sync.

src/web/jcc/App.tsx
JRoybot: OK

src/web/data/tasks/task.ts
```diff
@@ -1,15 +1,37 @@
...
+import { TaskParameterKeys, TaskParameterPair } from "../../types/jaia-system-types";
+import { jaiaGlobal } from "../jaia_global/jaia-global";
 
 export default class Task {
     private type: TaskType;
-    private diveParameters: DiveParameters;
-    private driftParameters: DriftParameters;
-    private constantHeadingParameters: ConstantHeadingParameters;
-    private stationKeepParameters: StationKeepParameters;
+
```
JRoybot, line 12: To accommodate bottom dives we need a way to indicate that.   In the old code it was just a boolean flag, however that contributed to the issues of having/missing parameters that do not match the type.  An alternative to that would be to have two distinct task types for interval dives vs bottom dives.  Either way we want to make sure we consider that when discussing how to manage the parameters.

src/web/context/JaiaContext.tsx
JRoybot: Is the only other way to delete waypoints deleting the entire mission?

michael-jaia: * So in some cases the mission being deleted will not be the missionIDInEditMode stored in context?
    * Yes, even when a mission is not in edit mode, the delete buttons still works in the missions panel.

* My concern is there are multiple sets of data defining what mission is being edited and was concerned that a race condition may exist where they are not in sync.
    * How I see it is there is only one property that defines what mission is being edited and that is `missionIDInEditMode`. This property is only updated in `JaiaContext` and then the data is transferred to the `JaiaGlobal` singleton so OpenLayers can view it. As we continue building out the waypoint panel, there will be a check to see if the waypoint belongs to the mission in edit mode. Whether that check is T or F will dictate what the panel renders (if the waypoint belongs to the mission in edit mode, then there will be an option to delete the waypoint, if it does not, then there will be no option to delete it, modify the task, or change the location).

michael-jaia:
> Assume we will be adding all the existing options like bottom dive toggle, position on map, constant heading on map.... even if they are not implemented does it make sense to add those items to the panel now?

We will be adding those options back, but to prevent this PR from growing even larger, I would prefer to add those in future work.

src/web/context/JaiaContext.tsx
JRoybot: Sounds good

src/web/data/jaia_global/jaia-global.ts
JRoybot: OK, might be overly paranoid on this one.  I agree would only happen if there was a bigger problem but would be a good catch if the bigger problem did occur.

src/web/types/jaia-system-types.ts
JRoybot: got it

src/web/data/tasks/task.ts
```diff
@@ -78,144 +81,111 @@ export default class Task {
...
+            return min;
         }
 
-        this.holdTime = holdTime;
+        return value;
     }
 
-    getDriftTime() {
-        return this.driftTime;
+    private updateDefaultTaskParameters() {
```
JRoybot, line 159: "default" parameters are just whatever was used last, not a true default.  Is that the desired behavior?

src/web/components/TaskParameters/TaskParameters.tsx
```diff
@@ -47,15 +47,20 @@ export default function TaskParameters(props: Props) {
 
 /**
  * Renders input fields for a dive task
+ *
+ * @notes
+ * Number(value).toString() remvoes the leading zero from input fields
```
JRoybot: since this is used in so many places would be a could thing to factor out into it's own helper function

src/web/data/jaia_global/jaia-global.ts
```diff
@@ -7,13 +7,19 @@ import {
...
 
 const defaultTaskParameters: TaskParameters = {
-    maxDepth: 10,
-    depthInterval: 10,
-    holdTime: 0,
-    driftTime: 0,
-    heading: 180,
-    constantHeadingTime: 30,
-    speed: 2,
+    dive: {
```
JRoybot, line 10: even if there is no functionality, should include the bottom dive flag for consistency with the protobuf types

src/web/data/jaia_global/jaia-global.ts
```diff
@@ -7,13 +7,19 @@ import {
...
+        max_depth: 10,
+        depth_interval: 10,
```
JRoybot, line 12: still curious to see how these parameters are handled when it is a bottom dive

JRoybot (commented):
This all looks good, left a few minor comments and questions.  Would like to have seen some tests included in the merge

src/web/data/jaia_global/jaia-global.ts
michael-jaia: When there is a bottom dive, the dive parameters are: 
```
{
    max_depth: 50,
    depth_interval: 50,
    hold_time: 0
}
```

src/web/data/jaia_global/jaia-global.ts
michael-jaia: No because the bottom dive field is going to be removed from the protobuf message

src/web/data/tasks/task.ts
michael-jaia: Yes

src/web/utils/input.ts
```diff
@@ -0,0 +1,10 @@
+/**
```
JRoybot, line 1: do you prefer keeping this as a standalone or do you think it could be combined with one function in `src/web/utils/conversions.ts` in some catchall like `src/web/utils/utils.ts`

JRoybot (approved):
Looks good!

src/web/utils/input.ts
michael-jaia: I was on the fence, but I am leaning yes because I can see this file being a good location for other function that deal with rounding and validating input values.

PR #1172, opened 2025-05-09
task/zoom-to-first-bot-with-GPS-fix/JAIA-1965

michael-jaia:
### Summary
Zooms map to first Bot with GPS fix so the operator does not need to scroll to the Bots on initial load. Future work will involve saving the center, zoom, and rotation of the map to local storage so the map will not need the zoom animation on every reload.

### Testing
https://github.com/user-attachments/assets/14fc8cd6-1bc3-4264-8f6d-5b582f03ec27


src/web/utils/constants.ts
```diff
@@ -3,3 +3,6 @@ export const DEFAULT_HUB_ID = 1;
 export const DETAILS_DECIMALS = 2;
 export const TEXT_OFFSET_RADIUS = 11;
 export const DATA_MODEL_POLL_TIME = 500; // milliseconds
+export const INITAL_ZOOM_DURATION = 3000; // milliseconds
+export const INITIAL_ZOOM = 12;
```
JRoybot: Should comment on the units of this.

JRoybot (changes requested):
Tested and ran this on LTS 22.04, built and ran.  I found the initial zoom to be very wide, I would like to see it almost 10 times smaller.  Not sure what the zoom level is but appears to be more than 1 km.  Please add comment on zoom level units.

PR #1174, opened 2025-05-13
task/develop-structure-for-buttons-that-apply-to-multiple-bots/JAIA-1978

michael-jaia:
### Summary
We established a structure for command buttons that apply to a singular Bot, however, we did not have a way to handle the fleet-wide buttons. These buttons need to be able to produce dialog boxes that describe which Bots can accept the command and which cannot. The content will vary between fleet-wide buttons to capture the different reasons a Bot may not be ready to handle the command. For each reason there should be an array that stores the Bot IDs that pertain to it. This can be seen at the top of the `ActivateAllButton` function. A `groupBotsByState` function will also be used to place each Bot into the correct array. The `ActivateAllDialog` then takes those arrays as Props and generates the correct messages. Those are the major differences between the single Bot button and a fleet-wide button.

### Testing
Unit tests cover: 
* All Bots ready for the activate command
* Some Bots ready for the activate command
* No Bots ready for the activate command


ferrom-jaia:
The command to activate should also not send if the bots status age has increased over our current limit of 30 seconds.

src/web/components/ActivateAllButton/__tests__/ActivateAllButton.test.tsx
```diff
@@ -0,0 +1,191 @@
...
+    // Click cancel button
+    const cancelButton = screen.getByText("Cancel");
+    await userEvent.click(cancelButton);
+    expect(screen.queryByText("Confirm")).toBeNull();
+
+    // Clean up
+    bots.getBots().clear();
+});
+
+test("Click activate all button with all Bots already activated", async () => {
```
JRoybot, line 94: Nice job testing all these combinations of conditions.  Because the structure of the test is very consistent this is a good candidate for a parameterized test.  The parameters could be composed of an array of `PortalBotStatus` and the expected outputs.  Would provide an easy way to add even more tests.

JRoybot (approved):
This looks great.  One suggestion on the test.

src/web/components/ActivateAllButton/__tests__/ActivateAllButton.test.tsx
michael-jaia: Jeff I'm going to hold off on forming the parameterized tests for now because I think it might be a little more custom than it appears. I just added a parameterized test for my other PR.

PR #1177, opened 2025-05-19
feature/discard-task-packet-jaia-1989

edsanville:
* Allow user to exclude/include task packets from consideration in the contour/interpolated drifts.
* The excluded task packets get exiled to their own "Excluded Packets" layer

Bonus stuff that I added to make my testing life easier.  These don't have to be part of this PR, the reviewers can decide:
* Persist the visibility of the measurement layers between browser sessions (localStorage)
* Persist the date range of the task packets between browser sessions

Missing:
* I have an "Include" and "Exclude" button, instead of the toggle switch.  I can convert this to a toggle switch, it's just slightly more involved given the way the task packet code is factored.


src/web/containers/TaskPacketPanel/TaskPacketPanel.tsx
```diff
@@ -3,16 +3,61 @@ import Icon from "@mdi/react";
...
+    const start_time_seconds = Math.round(task_packet.start_time / 1e6);
+    return `${task_packet.bot_id}_${start_time_seconds}`;
+}
+
 export function TaskPacketPanel(props: Props) {
-    let taskPacketData = props.taskPacketData;
-    delete taskPacketData?.title;
+    const task_packet = props.selectedTaskPacket;
+
+    // Probably want to refactor this into a type with optional name, value, units
```
JRoybot: Is this comment a TODO?  (Not sure if this team is in the habit of putting TODO on things they want to revisit)

src/web/containers/TaskPacketPanel/TaskPacketPanel.tsx
```diff
@@ -3,16 +3,61 @@ import Icon from "@mdi/react";
...
+        taskPacketData = {
+            // Snake case used for string parsing in task packet panel
+            bot_id: { value: task_packet.bot_id.toString(), units: "" },
+            depth_achieved: { value: dive.depth_achieved.toFixed(2), units: "m" },
+            dive_rate: { value: dive.dive_rate.toFixed(2), units: "m/s" },
+            bottom_dive: { value: dive.bottom_dive ? "Yes" : "No", units: "" },
+            start_time: { value: startTime.toLocaleString(), units: "" },
+            end_time: { value: endTime.toLocaleString(), units: "" },
+        };
+    } else if (props.type == "drift") {
```
JRoybot, line 43: If you did the refactor mentioned above would the name be used here instead of a literal?  And elsewhere to avoid the `drfit` you found in `CommandControl `?

src/web/containers/TaskPacketPanel/TaskPacketPanel.tsx
```diff
@@ -3,16 +3,61 @@ import Icon from "@mdi/react";
...
+        const drift = task_packet.drift;
+        const startTime = new Date(task_packet.start_time / 1000);
+        const endTime = new Date(task_packet.end_time / 1000);
+        taskPacketData = {
+            // Snake case used for string parsing in task packet panel
+            bot_id: { value: task_packet.bot_id.toString(), units: "" },
```
JRoybot, line 49: similarly, the units could come from a refactored `taskPacketData`

src/web/server/utils.py
```diff
@@ -0,0 +1,58 @@
+from typing import *
+from math import *
+from datetime import *
+
+
+def now_utime():
+    return int(datetime.now().timestamp() * 1e6)
+
+
+def utcnow():
```
JRoybot: suggest making these two function names similar now_utime & now_utc ?

src/web/shared/TaskPacketFeatures.ts
```diff
@@ -25,22 +25,24 @@ export function createDivePacketFeature(map: Map, task_packet: TaskPacket) {
     feature.setProperties({
         type: "dive",
         id: Math.random(),
+        task_packet: task_packet,
+        selected: false,
+        animated: false,
+        // Below data is now redundant, and we can refactor it away
```
JRoybot: should this comment be a TODO?

JRoybot (commented):
All looks good.  Left a few minor comments.  Did not build and test branch because have been working on 2.y and switching back and forth messes up my settings....

src/web/containers/TaskPacketPanel/TaskPacketPanel.tsx
edsanville: We can standardize on TODO, so it's easy to search.

src/web/containers/TaskPacketPanel/TaskPacketPanel.tsx
edsanville: The refactor I summarized would really be just an interface called maybe `DataRow`.  Quick example:

```
interface DataRow {
  name: string
  value: string
  units: string
}
```

That would make the function a lot more legible I think, especially the JSX.  I'm picturing something like pseudocode:

```
<div>
  rows.map((row) => {
    return <div>
      <div ...>{row.name}</div>
      <div ...>{row.value}</div>
      <div ...>{row.units}</div> 
    </div>
  }
</div>
```

edsanville: To avoid that "drfit" issue though, yeah, it might be advisable to have an enum for that somewhere.

src/web/containers/TaskPacketPanel/TaskPacketPanel.tsx
edsanville: Yes, that taskPacketData should absolutely be refactored into an interface.  I wanted to do that, but was afraid of "mission creep."

src/web/server/utils.py
edsanville: Looking at this function, it's actually dead code and never gets called.  Which is probably a good thing, since it returns the wrong answer.  The now_utime function is the one to use to get the true timestamp in microseconds.  I will remove the broken utcnow function.

There are a couple other places in the code where this sort of thing is done:
```
utime = datetime.datetime.utcnow().timestamp()
```

Which seems like it's the thing to do, but it actually doesn't work properly.  This is because `datetime.utcnow()`, for some unknown reason, returns a python "naive datetime" that has no timezone.  So, when calculating the timestamp, it assumes the system timezone, which results in an incorrect timestamp.  I will post to the demerits channel.

Good catch!

src/web/containers/TaskPacketPanel/TaskPacketPanel.tsx
JRoybot: I was asking... it is what I had done in the past but not sure what Matt prefers

JRoybot (dismissed):

src/web/server/app.py
```diff
@@ -243,11 +243,18 @@ def getPackets():
...
     startDate = parseDate(request.args.get('startDate', (datetime.now(timezone.utc) - timedelta(hours=14))))
     endDate = parseDate(request.args.get('endDate', ''))
-    return JSONResponse(jaia_interface.get_task_packets(start_date=startDate, end_date=endDate))
 
-@app.route('/jaia/v0/task-packets-count', methods=['GET'])
+    result = jaia_interface.task_packet_database.get_task_packets(start_date=startDate, end_date=endDate)
+    return JaiaResponse(result)
+
+@app.route('/jaia/v0/task-packets-version', methods=['GET'])
 def getPacketsCount():
```
ferrom-jaia: I think this function should get updated to reflect what it is doing.

JRoybot (dismissed):

src/python/pyjaia/src/pyjaia/utils.py
```diff
@@ -8,3 +10,44 @@ def myip():
...
+    Args:
+        task_packet (Dict): The task packet in dict form.
+
+    Returns:
+        str: The unique id for this task packet.
+    """
+
+    # Combine the bot_id and start_time (rounded to the nearest second)
+    bot_id = task_packet["bot_id"]
+    SECOND = 1_000_000
```
JRoybot, line 48: I did not know you could use numeric literals like that in python!

JRoybot (dismissed):
changes looks good

src/python/pyjaia/src/pyjaia/task_packet_database.py
```diff
@@ -7,47 +7,111 @@
...
+        db.execute('create index if not exists task_packets$bot_id on task_packets(bot_id)')
+        db.execute('create index if not exists task_packets$utime on task_packets(utime)')
+        db.execute('create index if not exists task_packets$included on included(included)')
+        db.commit()
+
+        return db
+    
+
+    def _update(self):
+        """Updates taskpacket files from disk, if enough time has passed.  THREAD UNSAFE.
```
JRoybot, line 59: are these comments regarding threading identifying follow on work or just a heads up in case this is used in a threaded environment?

src/python/pyjaia/tests/test_task_packet_database.py
JRoybot, line 1: for consistency purposes should put tests either in a `tests` folder or a `test` folder like the one in `rest_api`.  Ideally we would be using a test framework so these tests can be integrated into circleCI.  @ferrom-jaia

src/python/pyjaia/src/pyjaia/task_packet_database.py
ferrom-jaia: I think they are there so if a new developer comes in they know which function to use

src/python/pyjaia/src/pyjaia/utils.py
edsanville: Yeah, it's great for readability!

src/python/pyjaia/src/pyjaia/task_packet_database.py
edsanville: Yes, that's why.  The user should make sure they only call those functions when the `self.lock` is engaged.  I also used the underscore prefix to make people think twice before calling those functions.

src/python/pyjaia/tests/test_task_packet_database.py
edsanville: Yes, ideally we would have some good tests for CircleCI.  The two files in this directory were really just ad-hoc scripts to help while coding.  It was probably a bad idea to even call them "tests,"  but I also wanted to get them into the repository to help with dev work.

I can integrate a test framework and put together some real tests for the task packet database and/or the contour generation.  Should it be a separate ticket @ferrom-jaia ?

ferrom-jaia: Yes, I will make a follow ticket for this task.

ferrom-jaia: https://jaia-innovation.atlassian.net/browse/JAIA-1998

edsanville: If anyone has a favorite python test framework, we can use that.  I've got no experience with python test frameworks.

ferrom-jaia (approved):
Looks good! Tested in sim and everything checked out. Going to test in the field before releasing.

PR #1178, opened 2025-05-19
task/create-layer-switcher-menu/JAIA-1966

michael-jaia:
### Summary
Creates a layer switcher menu. Previously we used the [ol layerswitcher](https://github.com/walkermatt/ol-layerswitcher) library to manage the visible layer(s). The downside of this library is the lack of style customization. This pull request gives us the same layer switching flexibility with complete control over the style.

### Demo
[Video](https://jaiatech.sharepoint.com/:v:/s/ProductDevelopment/EanzitSr69hDvToLKE7XGVkBfLCfGpjZZggjAUf2uUmowQ?e=YWg8Vx) hosted on SharePoint (ping me for access if you do not have it)

### Testing
Unit tests cover:
* Displaying each layer
* Toggling between layers


src/web/types/context-types.ts
```diff
@@ -32,6 +32,20 @@ export interface BotAccordionStates {
...
 
+export const enum MapLayerGroupNames {
+    BASE_MAPS = "baseMaps",
+    BATHYMETRY = "bathymetry",
+    MEASUREMENTS = "measurements",
+    MISSION = "mission",
+}
+
+export interface MapLayerGroupStates {
+    baseMaps: boolean;
```
JRoybot, line 43: Had to look at the uses of this to realize it is the state of the accordion, why not call it an accordion state? MapLayerAccordionStates ?

JRoybot (commented):
This all looks good, suggested one name change.  Built branch and ran unit tests

src/web/types/context-types.ts
michael-jaia: Good call, I will update that

JRoybot (approved):

PR #1180, opened 2025-05-19
feature/ph-compensation-calculations

ferrom-jaia:
## Description
This PR introduces a temperature compensation function for pH measurements based on the Nernst equation, allowing raw pH readings taken at various temperatures to be normalized to a standard reference temperature of 25 °C.

The Nernst equation models how the voltage of a pH electrode changes with temperature:
```
E = E₀ - (2.303 * R * T / nF) * pH
```
From this, the temperature-dependent electrode slope is derived as:

```
slope ≈ 0.1984 × T (in Kelvin)
```
At 25°C (298.15 K), the slope is ~59.16 mV/pH — used as the calibration standard. As temperature increases, the electrode becomes more sensitive, affecting pH readings. This function adjusts the measured pH by scaling its deviation from pH 7.0 using the ratio of the slope at the measurement temperature to the slope at 25°C.

References:
* https://mantech-inc.com/faq/how-does-mantech-account-for-temperature-compensation-and-correction-in-ph-measurements/
* https://atlas-scientific.com/ph-temperature-calculator/
* https://www.horiba.com/usa/water-quality/support/technical-tips/bench-meters/automatic-temperature-compensation-in-ph-measurement/
* https://cdn.hach.com/7FYZVWYB/at/7jpb8qbgw4v62mvtc88vqg74/LIT2007-Temperature_Compensation_with_pH_Measurement.pdf

## Testing
* Added unit tests were added to validate the temperature compensation logic using trusted data from MANTECH and other sources.


src/lib/utils/ph_temperature_compensation.h
```diff
@@ -0,0 +1,31 @@
...
+ *      * https://www.horiba.com/usa/water-quality/support/technical-tips/bench-meters/automatic-temperature-compensation-in-ph-measurement/
+ *      * https://mantech-inc.com/faq/how-does-mantech-account-for-temperature-compensation-and-correction-in-ph-measurements/
+ */
+double temperature_compensated_ph(double measured_ph, double temperature_celsius)
+{
+    const double slope_at_25C = 59.16; // mV/pH at 25°C
+    double temp_kelvin = temperature_celsius + 273.15;
+    double slope = 0.1984 * temp_kelvin;
+
+    return 7.0 + (measured_ph - 7.0) * (slope / slope_at_25C);
```
edsanville: It looks like we should be multiplying by `slope_at_25C / slope`, shouldn't we?  The slope is the change in measured E per change in pH, so:

* for a high temperature, we have a high slope, (the measured E changes more per unit of true pH)
* the compensated pH should be closer to 7.0 than the measured pH, (not further away)
* for a low temperature, we have a low slope
* the compensated pH should be further from 7.0 than the measured pH

src/test/utils/test.cpp
```diff
@@ -29,6 +30,48 @@ BOOST_AUTO_TEST_CASE(test_conductivity)
...
+        // PH: 7.00 at 25 C
+        {7.12, 0.0, 7.00}, {7.09, 5.0, 7.00}, {7.06, 10.0, 7.00}, {7.04, 15.0, 7.00},
+        {7.02, 20.0, 7.00}, {7.00, 25.0, 7.00}, {6.98, 30.0, 7.00}, {6.98, 35.0, 7.00},
+        {6.97, 40.0, 7.00}, {6.97, 45.0, 7.00}, {6.97, 50.0, 7.00},
+        // PH: 10.01 at 25 C
+        {10.32, 0.0, 10.01}, {10.25, 5.0, 10.01}, {10.18, 10.0, 10.01}, {10.12, 15.0, 10.01},
+        {10.06, 20.0, 10.01}, {10.01, 25.0, 10.01}, {9.97, 30.0, 10.01}, {9.93, 35.0, 10.01}, 
+        {9.89, 40.0, 10.01}, {9.86, 45.0, 10.01}, {9.83, 50.0, 10.01},
+    };
+
```
edsanville, line 71: Following my previous comment:  If we're correcting a measured pH to a true pH, then we'd expect:

* For a low pH (<7.0) and a low temperature (< 25C), then the actual pH would be further from 7.0 (lower) than the measured pH, but the first test is the opposite:
  {3.775, 0.0, 4.00}

In summary, it looks to me like the `temperature_compensated_ph` function is adjusting backward as a result of the previous comment, and the tests are backward as well?  Unless I'm missing something...

src/lib/utils/ph_temperature_compensation.h
ferrom-jaia: Yes, I agree. Thank you for pointing this out.

src/test/utils/test.cpp
ferrom-jaia: Thank you Ed!

JRoybot (dismissed):
all looks good to me

edsanville (dismissed):
Looks good!

PR #1182, opened 2025-05-22
task/send-stop-command-single-bot

michael-jaia:
### Summary
As the name of the PR indicates, the goal was to send the stop command to a single Bot. Prior to doing that we needed to send a mission to a Bot so it could be in the correct state to receive a stop command. This PR includes the code to start a mission and stop a mission. To send a mission to the Bot we package the data from the data model into protobuf message format the Hub is expecting with helper function `package...ForHub` in the `Mission`, `Waypoint`, and `Task` classes.

### Testing
Unit tests cover all disabled codes related to the start mission button/dialog (except for `DOWNLOAD_QUEUE` which needs to be implemented first). 
* `NONE`
* `NO_COMMS`
* `MISSION_STATE`
* `NO_MISSION_ASSIGNED`
* `LOW_BATTERY`


src/web/components/StartMissionButton/__tests__/StartMissionButton.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+    expect(screen.getByText(messages.get(DisabledCodes.LOW_BATTERY)));
+    expect(screen.getByText("Close"));
+    expect(screen.getByText("Override"));
+});
+
+test("Start mission", async () => {
+    render(<StartMissionButton bot={bots.getBot(5)} mission={mockMission2} />);
+    const button = screen.getByRole("button", { name: "start-mission-individual-bot" });
+    await userEvent.click(button);
+    expect(screen.getByText("Confirm"));
```
JRoybot, line 104: could mock `setIsDialogVisible` to verify it is called.  (unlikely to ever fail but why not)

src/web/components/StartMissionButton/__tests__/StartMissionButton.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+test("Button click with Bot not assigned to mission", async () => {
+    render(<StartMissionButton bot={bots.getBot(3)} mission={mockMission1} />);
+    const button = screen.getByRole("button", { name: "start-mission-individual-bot" });
+    await userEvent.click(button);
+    expect(screen.getByText("Alert"));
+    expect(screen.getByText(messages.get(DisabledCodes.NO_MISSION_ASSIGNED)));
+    expect(screen.getByText("Close"));
+});
+
+test("Button click with critically low battery", async () => {
```
JRoybot, line 90: Is there any chance more than one of these conditions could co-exist?  If so should probably be tested.

src/web/components/TaskParameters/TaskParameters.tsx
```diff
@@ -89,7 +89,7 @@ function DiveParameters(props: SubProps) {
             <input
                 name={TaskParameterKeys.DRIFT_TIME}
                 type="number"
-                value={formatNumericalInput(props.task.getDrfitParameters().drift_time)}
+                value={formatNumericalInput(props.task.getDriftParameters().drift_time)}
```
JRoybot, line 92: Wow I saw Dr. Fit in Ed's review as well ... popular guy!

src/web/containers/BotDetails/BotDetails.tsx
```diff
@@ -23,26 +28,17 @@ import {
...
-import {
-    mdiPlay,
-    mdiStop,
-    mdiPower,
-    mdiDelete,
-    mdiRestart,
-    mdiSkipNext,
-    mdiRestartAlert,
-} from "@mdi/js";
+import { mdiPlay, mdiPower, mdiDelete, mdiRestart, mdiSkipNext, mdiRestartAlert } from "@mdi/js";
```
JRoybot, line 41: I think we should stick with the multi-line imports for things that end up getting wrapped in a standard width editor.

src/web/context/JaiaContext.tsx
```diff
@@ -681,12 +676,12 @@ function syncOpenLayers() {
...
 
 /**
- * Retrieves the Waypoint object that is currently selected
+ * Retrieves the Task object connected to the currently selected waypoint
  *
- * @returns {Waypoint} Access to Waypoint modifiers
+ * @returns {Task} Access to Task modifiers
  */
-function getWaypoint() {
+function getTask() {
```
JRoybot, line 683: This file is starting to get pretty  big.  If we could move some of the functions into other files it will probably be easier to maintain.  Thoughts?

src/web/data/missions/mission.ts
```diff
@@ -75,4 +84,32 @@ export default class Mission {
             );
         }
     }
+
+    packageMissionForHub() {
+        const missionPlan: MissionPlan = {
+            start: MissionStart.START_IMMEDIATELY,
+            movement: MovementType.TRANSIT,
+            goal: this.packageWaypointsForHub(),
```
JRoybot, line 92: What is the distinction between goal and waypoint?  I have always found the use of the two words confusing because they seem to mean the same thing.

src/web/data/tasks/task.ts
```diff
@@ -37,6 +38,7 @@ export default class Task {
     private isEnablePAM: boolean;
 
     constructor() {
+        this.type = TaskType.NONE;
         const defaults = jaiaGlobal.getDefaultTaskParameters();
         this.setDiveParameters(defaults.dive);
```
JRoybot, line 43: why do these get set for a task type of none... couldn't the defaults just be applied when the task type is selected?

JRoybot (commented):
Did a static review with a few comments.  Built the branch and am testing functionality

JRoybot (approved):
Previous review was mostly questions... approving if you want to merge as is.

michael-jaia:
> Tested branch and have question on how disabled buttons are intended to work. When the start button is disabled a dialog explaining why pops up when the button is pushed. In most other apps a disabled button can not be pushed and if there is an explanation as to why it is disabled it is typically displayed when the user hovers on the button. Some users might not think to push a button that is disabled and may never see the dialog. (comment applies to the stop button and disabled JCC buttons in general

 That is a good point. I will add it to the programming notes as something to review at the end with Matt and Michael.

src/web/components/StartMissionButton/__tests__/StartMissionButton.test.tsx
michael-jaia: By the text that appears after the click we know it is functioning correctly. We could prove it out further by saying for example before the click `expect(screen.getByText("Confirm").toBeNull` or something along those lines.

src/web/containers/BotDetails/BotDetails.tsx
michael-jaia: That pre-commit auto formatter made this change. I believe it is because I removed an import which meant it could fit all imports on the same line without violating the width rule

src/web/context/JaiaContext.tsx
michael-jaia: Yes, I want to wait closer to the end to do that.

src/web/data/missions/mission.ts
michael-jaia: goal is the term that is used on the C++/proto side. Waypoint is what we will use on the frontend. It would be nice to have them the same but I think it is important we stick with waypoint on the webside otherwise we will have goal data in the waypoint panel. With the data model, this mission.ts and bot.ts are the only places we should see the word goal (the two places where we need to interface between the server and the client)

src/web/data/tasks/task.ts
michael-jaia: I thought it would be nice to know that the moment a waypoint is created on the map, this property would not be undefined.

src/web/components/StartMissionButton/__tests__/StartMissionButton.test.tsx
michael-jaia: There is but only one will be displayed at a time (because of the return statements) which is the intended behavior.

src/web/data/missions/mission.ts
JRoybot: I would suggest changing C++/proto side but must be for a reason, guessing it is because waypoints in MOOS are just locations and don't have tasks.

PR #1186, opened 2025-05-27
task/increase-number-of-wpts

ferrom-jaia:
## Description

Increase number of waypoints a user is able to send at one time in the JCC.

## Test

This was tested in sim:
* 80 waypoints with no task
* 80 waypoints with bottom dive task
* 80 waypoints with dive task and hold


edsanville (dismissed):
Code looks good, and I was able to submit an 80-waypoint task.

PR #1197, opened 2025-05-29
task/add-back-start-all-missions-button

michael-jaia:
### Summary
This pull request re-instates the start all missions button. It is based on the structure used in the activate all button. While implementing the functionality around this button, parts of the activate all structure did not scale well. Individually defining each array connected to a DisabledCode became messy, so I transitioned the structure to a map keyed on the DisabledCodes that pairs with an array to contain Bot IDs.

### Testing
**Unit tests cover the following scenarios:**
* 0 Bots ready to start a mission
* 1 Bot ready to start a mission
* 2 Bots ready to start a mission
* All Bots ready to start a mission

**Simulation testing covers:**
* Verifying the aesthetics of the dialog box for various combinations of warnings and confirmations
* Sending the command and watching the Bots execute their missions


src/web/components/StartMissionButton/start-mission-messages.ts
```diff
@@ -1,10 +1,10 @@
 export enum DisabledCodes {
-    NONE = 0,
-    NO_COMMS = 1,
-    MISSION_STATE = 2,
-    NO_MISSION_ASSIGNED = 3,
-    DOWNLOAD_QUEUE = 4,
-    LOW_BATTERY = 5,
+    NONE = 1,
```
JRoybot, line 2: did these need to change because they are used to index the array?

src/web/components/StartAllMissionsButton/StartAllMissionsButton.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+import { Command, CommandType } from "../../types/protobuf-types";
+import { isCommandAvailable, sendBotCommand } from "../../utils/commands";
+import { microsecondsToSeconds } from "../../utils/conversions";
+import { MIN_BATTERY_PERCENT, NO_COMMS_STATUS_AGE, UNASSIGNED_ID } from "../../utils/constants";
+
+import "../../style/stylesheets/util.less";
+
+interface Props {
+    bots: Map<number, Bot>;
```
JRoybot, line 23: Is there a reason these need to be passed in as props instead of pulled directly from jaiaContext?

JRoybot (commented):
Looks good.  Reviewed code, couple of questions.  Built branch and ran sim, tested in browser, ran all npm tests.  Got a bunch of console errors indicating the jaiaAPI mock is not working...

src/web/components/StartAllMissionsButton/StartAllMissionsButton.tsx
michael-jaia: The component could use context, but none of the other buttons are using context, so I just wanted to stick with props. I think it keeps the buttons a little simpler as they only get the data they need.

src/web/components/StartMissionButton/start-mission-messages.ts
michael-jaia: They did not need to change to index the list, but when working on the BIO work, I ran into a pretty weird bug where 0 in an enum did not register in nano protobuf, so I was thinking it could just be a good practice to start enums at 1. Have you ever come across something like that?

JRoybot (approved):

src/web/components/StartMissionButton/start-mission-messages.ts
JRoybot: no I have not, but not surprised, nanoPB isn't a full implementation...

PR #1198, opened 2025-05-29
bug/protoc-python-and-salinity-requires

ferrom-jaia:
## Description

Updated protoc command to include sensor folder, salinity field is now optional, updated jaiabot_pressure_sesnor to point to the correct location for the proto message


marshalllele (approved):
Looks great!

PR #1202, opened 2025-05-29
task/add-back-jaia-button

michael-jaia:
### Summary
Integrates the Jaia button into the new panel structure. While working on this ticket, I combined the `TopButtonList` and `SideButtonList` components into `ButtonList`. If I did not do this, I would have repeated the core functions. To combine these components, we now pass a button list type (TOP or SIDE) to `ButtonList` to determine what format will render.

### Testing
The panel opens and closes in simulation. It fits on an iPad Mini size screen and displays the correct information. There can be a slight delay to the version loading because it is fetched from the server each time the panel is opened. I did not write any automated unit tests because most of the data is hard-coded. Down the line it could be beneficial to write some integration tests for panel renderings.

### Future Work
Obtain an SVG that does include the slogan Micro is More, so the button icon can be larger.


src/web/components/ButtonList/ButtonList.tsx
```diff
@@ -0,0 +1,90 @@
...
+        jaiaDisaptch({ type: JaiaActions.CLICKED_PANEL_BUTTON, panelName: panelName });
+    };
+
+    /**
+     * Provides the class name to style the button based on its selection state
+     *
+     * @param {PanelNames} panelName Name of panel associated with button
+     * @returns {void}
+     */
+    const getSelectedClassName = (panelName: PanelNames) => {
```
JRoybot, line 49: suggest adding "jaia-button" to the output of this function and then using it for all the buttons below instead of having to put the "jaia-button" literal on each one separately

src/web/components/JaiaAbout/JaiaAbout.tsx
```diff
@@ -0,0 +1,66 @@
...
+            setVersion(`${version.major}.${version.minor}.${version.patch}`);
+        }
+    };
+
+    return (
+        <div className="jaia-about">
+            <img src={JaiaLogo}></img>
+            <div className="jaia-about-row">
+                <div className="label">Website:</div>
+                <a href="https://www.jaia.tech" target="_blank" rel="noopener noreferrer">
```
JRoybot, line 42: Could we put all this information in a json file so we do not have to edit source code if something were to change?

JRoybot (changes requested):
Looks good, suggested a few changes, nothing major

michael-jaia:
> Built and ran the branch in a browser and reran all unit tests. All tests passed (known issue of console errors related to JaiaAPI mock). Noted About button icon is too small to read, is the "Future Work" above going to be part of this PR? JCC version in pop up is blank, assume it is because version under test is not a release.

Future work will be a different PR. You should have seen the version number...did you have the Hub or any Bots running in sim?

src/web/components/JaiaAbout/JaiaAbout.tsx
michael-jaia: Yes, but it is not a high enough priority to do it now.

PR #1204, opened 2025-05-29
task/align-activate-all-with-start-all

michael-jaia:
### Summary
After working through the start all missions button, we developed better ways to implement the fleet wide command buttons. This PR involves going back to the activate all button and applying those changes. Additionally, it includes adding the `NO_COMMS` condition to the activate button. 

### Testing
All previous tests passed. Furthermore, I added a test to verify the `NO_COMMS` condition for the activate button.


src/web/components/ActivateAllButton/ActivateAllButton.tsx
```diff
@@ -99,11 +99,34 @@ export default function ActivateAllButton(props: Props) {
...
         </div>
     );
 }
+
+/**
+ * Maps each disabled code to an empty array to prevent undefined behavior
+ *
+ * @returns {Map<DisabledCodes, number[]>} Disabled codes mapped to an empty array for Bot IDs
+ */
+function initBotReadyStates() {
```
JRoybot, line 115: The handling of `DisabledCodes` is very similar on a lot of components.  Different components track different subsets of the DisabledCodes.  Wondering if we could consolidate much of the code if we tracked all the `DisabledCodes` in all of them but just left the arrays of the ones that don't apply empty.  Not sure about the impact of this .... just a thought

src/web/components/ActivateAllButton/ActivateAllDialog.tsx
```diff
@@ -34,46 +29,51 @@ export enum DialogActions {
 export function ActivateAllDialog(props: DialogProps) {
     /**
      * Applies the base class "jaia-dialog" and appends "alert"
-     * if no Bots are in a state to receive the command
+     * if at least one Bot cannot receive the command to adjust spacing
```
JRoybot: same as other review... not sure what "to adjust spacing" means, maybe leave it out

src/web/components/ActivateAllButton/ActivateAllDialog.tsx
```diff
@@ -34,46 +29,51 @@ export enum DialogActions {
...
     const generateMessage = () => {
-        return [
-            generateSubMessage(props.availableBotIDs, TextTypes.AVAILABLE),
-            generateSubMessage(props.activatedBotIDs, TextTypes.ACTIVATED),
-            generateSubMessage(props.noCommsBotIDs, TextTypes.NO_COMMS),
-        ];
+        const message: string[] = [];
+
+        // Halve the length to only count keys
+        const disabledCodesLength = Object.keys(DisabledCodes).length / 2;
```
JRoybot, line 50: you could use the size property of the map and not have to divide

src/web/components/ActivateButton/__tests__/ActivateButton.test.tsx
```diff
@@ -30,10 +37,19 @@ test("Click activate button in disabled state due to mission state", async () =>
     expect(screen.getByText("Close")).toBeInTheDocument();
 });
 
-test("Click activate button in enabled state", async () => {
+test("Click activate button in disabled state due no comms", async () => {
```
JRoybot: "due to no comms" or "due to lack of comms"

JRoybot (changes requested):
Built and ran branch, tested JCC in browser.  Ran all unit tests.

Made a few minor comments.

src/web/components/ActivateAllButton/ActivateAllDialog.tsx
michael-jaia: I don't love that we have to divide by 2 but I want to get the length from enum. I think it could be a little confusing to have `disabledCodesLength = messages.size;` especially with how variable `message` is used in the file.

src/web/components/ActivateAllButton/ActivateAllButton.tsx
michael-jaia: I agree, I think there is some purging we can do between buttons. I want to save that for when they are all complete and we can do a thorough review.

PR #1207, opened 2025-06-03
task/make-lat-lon-coordinates-editable/JAIA-1963

michael-jaia:
### Summary
Allows operator to enter latitude and longitude points for a waypoint. As the operator edits the coordinates, the waypoint moves accordingly on the map.

### Implementation Details
Waypoint location data exists in both number and string form. We utilize the number type when saving to the data model and string type when working with user input. We need to use strings when working with user input to allow negative signs and decimal points. As the user enters a coordinate, we will check if the value can be converted to a number. If it can, we will update the data model with the numerical form of the user input.

### Tests
* To test the WaypointPanel with written tests, we need a larger integration test, otherwise, we run into the situation where we need to do some code gymnastics to please the React engine. Writing a larger integration test is not part of this ticket.
* A key function `validateCoordiantes` is accompanied by a set of unit tests
* Using the simulator, waypoints update as expected entering various inputs.


src/web/context/JaiaContext.tsx
```diff
@@ -395,6 +399,22 @@ function handleDeleteWaypoint(mutableState: JaiaContextType) {
...
 }
 
+/**
+ * Makes the calls to move a waypoint to a user set location
+ *
+ * @param {JaiaContextType} mutableState State object ref for making modifications
+ * @param {GeographicCoordinate} location New location of the waypoint
+ * @returns {JaiaContextType} Updated mutable state object
+ */
+function handleMoveWaypoint(mutableState: JaiaContextType, location: GeographicCoordinate) {
```
JRoybot, line 409: Will we be able to use this action when we re-enable drag-to-move?

JRoybot (dismissed):
Everything looks good.  Built and tested branch in browser and ran all unit tests

src/web/context/JaiaContext.tsx
michael-jaia: Yes that's the goal

PR #1208, opened 2025-06-03
task/add-back-bottom-dive-toggle/JIRA-1962

michael-jaia:
### Summary
This pull request includes the bottom dive toggle with the dive parameters displayed in the waypoint panel. When the toggle is switched on, the only parameter visible is drift time. Behind the scenes the parameters are configured as:
* max_depth=50
* depth_interval=50
* hold_time=0

### Testing
Exercised the toggle in simulation, logging the dive parameters. Similar to https://github.com/jaiarobotics/jaiabot/pull/1207, we need a full integration test to simulate the WaypointPanel in a test environment. I will add a ticket to JIRA for that work.


src/web/components/TaskParameters/TaskParameters.tsx
```diff
@@ -35,9 +49,25 @@ export default function TaskParameters(props: Props) {
         });
     };
 
+    /**
+     * Dispatches action to update the bottom dive toggle and the
+     * dive parameters
+     *
+     * @return {void}
+     */
+    const handleBottomDiveClick = () => {
```
JRoybot, line 58: random question... why do we use const arrow functions instead of regular functions in this context?

src/web/components/TaskParameters/TaskParameters.tsx
```diff
@@ -50,50 +80,79 @@ export default function TaskParameters(props: Props) {
...
+    }
 
-            <div>Hold Time</div>
-            <input
-                name={TaskParameterKeys.HOLD_TIME}
-                type="number"
-                value={formatNumericalInput(diveParameters.hold_time)}
-                autoComplete="off"
-                onChange={(evt) => props.onChange(evt)}
+    return (
```
JRoybot: matter of preference but I'd prefer to see this in an else clause , better yet create 2 sub functions for each dive type

src/web/data/tasks/task.ts
```diff
@@ -146,6 +127,15 @@ export default class Task {
         this.updateDefaultTaskParameters();
     }
 
+    /**
```
JRoybot: could this belong in `src/web/utils/input.ts`

JRoybot (dismissed):
All looks good, made a few suggestions, nothing big.  Ran and tested in browser

src/web/components/TaskParameters/TaskParameters.tsx
michael-jaia: It's a concise way to write a function within a function. It used to be more important in class components because it would capture `this` from the parent scope, instead of having to bind `this` to the function call.

src/web/components/TaskParameters/TaskParameters.tsx
michael-jaia: Switched to if/else. I like it as one function to keep the related JSX code together.

src/web/data/tasks/task.ts
michael-jaia: Yes, moving it over

PR #1213, opened 2025-06-04
bug/mission-in-edit-mode-should-be-on-top/JAIA-1994

JRoybot:
Bug Fix

- Added logic to properly stack the missions in their layer based on edit 
- Fixed existing logic to stack bots based on edit mode
-  added explicit stacking of the bots, hub and mission layers.

Testing

- verified behavior in browser


src/web/openlayers/features/waypoint-feature.ts
```diff
@@ -175,3 +176,23 @@ function getWaypointColor(missionID: number) {
...
 }
+
+/**
+ * Supplies the zIndex for waypoints and lines based on edit mode
+ *
+ * @param {number} missionID Used to determine zIndex for waypoints
+ * @returns {number} zIndex to be applied to waypoint
+ */
+function getWaypointZIndex(missionID: number) {
+    const selectedNode = jaiaGlobal.getSelectedNode();
```
michael-jaia: I think we can remove this line

src/web/openlayers/layers/vector/zindex.ts
```diff
@@ -0,0 +1,7 @@
+import { LayerTitles } from "../../../types/openlayers-types";
+
+export const openLayersZIndexes = new Map<LayerTitles, number>();
```
michael-jaia: I think we should change this to layerZIndexes because it only applies to layers not features.

michael-jaia (changes requested):
Jeff, code looks good. Added a couple small comments.

src/web/openlayers/features/waypoint-feature.ts
```diff
@@ -175,3 +176,34 @@ function getWaypointColor(missionID: number) {
...
+ * Supplies the zIndex for waypoints and lines based on edit mode
+ *
+ * @param {number} missionID Used to determine zIndex for waypoints and lines
+ * @param {number} waypointNum Used to determine zIndex for waypoints, leave unassigned for lines
+ * @returns {number} zIndex to be applied to waypoint and lines
+ */
+function getWaypointZIndex(missionID: number, waypointNum?: number) {
+    let waypointZIndex = 0;
+    // Provide proper mission stacking
+    if (missionID === missions.getMissionIDInEditMode()) {
```
michael-jaia, line 190: Can we combine this if statement with the bottom if statement?

src/web/openlayers/features/waypoint-feature.ts
```diff
@@ -175,3 +176,34 @@ function getWaypointColor(missionID: number) {
...
 }
+
+/**
+ * Supplies the zIndex for waypoints and lines based on edit mode
+ *
+ * @param {number} missionID Used to determine zIndex for waypoints and lines
+ * @param {number} waypointNum Used to determine zIndex for waypoints, leave unassigned for lines
+ * @returns {number} zIndex to be applied to waypoint and lines
+ */
+function getWaypointZIndex(missionID: number, waypointNum?: number) {
```
michael-jaia, line 187: Does waypointNum need to be an optional parameter?

michael-jaia (changes requested):

src/web/openlayers/features/waypoint-feature.ts
JRoybot: Yes because I wanted this function to be used for the waypoint icons and the waypoint lines instead of having 2 separate functions.  None of the code creating the line segments has a waypoint num available.  So the logic to stack individual waypoints is only applied to the waypoint icons.  all of the lines for a given mission will be at the same level and underneath all the waypoint icons

src/web/openlayers/features/waypoint-feature.ts
JRoybot: I don't think so because I wanted to stack the line segments and the waypointNum for those features is undefined.  The selected waypoint does not seem to get reset if the mission in edit is changed, so I needed to check both below.  Open to suggestions if you think of a better way.

michael-jaia: I just needed to play around with it, looks good how you have it

michael-jaia (approved):

PR #1214, opened 2025-06-05
task/add-back-tap-to-move/JAIA-1983

michael-jaia:
### Summary
Adds "Tap To Move" toggle to the waypoint panel. This allows an operator to adjust the position of a waypoint during the mission planning process.

### Sequence
1. An operator selects a waypoint and switches Tap to Move on
    * The `CLICKED_TAP_TO_MOVE` action is dispatched from the `WaypointPanel`
    * In the data model, the property `isMovable` is set to `true` for the selected waypoint
3. The operator clicks on the map to move the waypoint
    * The map registers a click
    * It checks to see if there is a selected waypoint and if that waypoint is movable
    * If it is movable, the waypoint will shift to the click point
    * It is not movable, a new waypoint will be created

### Testing
* Tested in simulation
    * Variable mission lengths
    * Multiple missions
    * Deleting waypoints
    * Clicking waypoints
    
* Unit tests will come in a separate ticket once the `WaypointPanel` integration test is complete


JRoybot (changes requested):
Code looks good.  Built and tested branch.

I think the ability to move a waypoint should be integrated with the mission in edit mode.  I do not think it is logical that I can move a waypoint from the waypoint panel and map for a mission that is not the mission in edit mode.  

I suggest either disable the tap to move toggle if the mission is not in edit mode and display a message why if they click it, or change the mission in edit mode to the mission of the waypoint if tap to move is toggled.

Of the two I prefer the first option

michael-jaia:
> Code looks good. Built and tested branch.
> 
> I think the ability to move a waypoint should be integrated with the mission in edit mode. I do not think it is logical that I can move a waypoint from the waypoint panel and map for a mission that is not the mission in edit mode.
> 
> I suggest either disable the tap to move toggle if the mission is not in edit mode and display a message why if they click it, or change the mission in edit mode to the mission of the waypoint if tap to move is toggled.
> 
> Of the two I prefer the first option

Jeff I agree, I was thinking we would finalize the implementation of edit mode in a separate ticket. For instance, right now when we start a mission, the mission remains in edit mode.

JRoybot (approved):
Agree we can follow up with more mission in edit mode functionality

PR #1215, opened 2025-06-05
task/add-back-help-window

michael-jaia:
### Summary
Brings back the help window into the JCC. Minor tweaks include:
* The window does not contain a close button to match the other panels
* The window style stays consistent between smaller and larger screen sizes
* The title does not scroll over the window

### Testing
* Opening and closing the help window along with other panels in the JCC

image


JRoybot (approved):
All looks good.  Ran and tested branch

PR #1219, opened 2025-06-06
task/add-back-mission-flags

michael-jaia:
### Summary
Adds back flags containing the mission ID above the first waypoint of each mission.

### Testing
Worked with adding, deleting, and toggling edit mode for multiple missions

image


JRoybot (approved):
All looks good, built and tested branch.

PR #1222, opened 2025-06-06
task/highlight-target-waypoint

michael-jaia:
### Summary
This pull request highlights the target waypoint on the map. It also switches edit mode off when a mission begins. 

### Testing
Ran multi-Bot missions watching the target waypoint and edit mode toggle off correctly.


src/web/components/StartAllMissionsButton/StartAllMissionsButton.tsx
```diff
@@ -83,19 +86,21 @@ export default function StartAllMissionsButton(props: Props) {
...
                 const startMissionCommand: Command = {
                     bot_id: botID,
                     type: CommandType.MISSION_PLAN,
-                    plan: props.missions
-                        .get(missionsManager.getMissionID(botID))
-                        .packageMissionForHub(),
+                    plan: props.missions.get(missionID).packageMissionForHub(),
                 };
-                sendBotCommand(startMissionCommand);
+                const res = await sendBotCommand(startMissionCommand);
```
JRoybot, line 100: what does `await` do if `sendBotCommand` is not declared as an `async` or return a promise?

src/web/components/StartMissionButton/StartMissionButton.tsx
```diff
@@ -91,7 +94,13 @@ export default function StartMissionButton(props: Props) {
                 type: CommandType.MISSION_PLAN,
                 plan: props.mission.packageMissionForHub(),
             };
-            sendBotCommand(startMissionCommand);
+            const res = await sendBotCommand(startMissionCommand);
```
JRoybot, line 97: same question regarding the use of await

src/web/containers/BotDetails/BotDetails.tsx
```diff
@@ -185,19 +185,19 @@ export default function BotDetails() {
                                             </td>
                                         </tr>
                                         <tr>
-                                            <td>Active Goal</td>
+                                            <td>Target Waypoint</td>
```
JRoybot, line 188: Nice.   We can put the G word away with the R word :-)

src/web/context/JaiaContext.tsx
```diff
@@ -367,6 +370,23 @@ function handleChangeMissionSpeeds(mutableState: JaiaContextType, missionSpeeds:
...
+ * Turns off edit mode upon starting a mission
+ *
+ * @param {JaiaContextType} mutableState State object ref for making modifications
+ * @param {number} missionID Checks sent mission against missionID in edit mode
+ * @returns {void}
+ */
+function handleSendMission(mutableState: JaiaContextType, missionID: number) {
+    if (missions.getMissionIDInEditMode() === missionID) {
+        missions.setMissionIDInEditMode(UNASSIGNED_ID);
+        mutableState.missionIDInEditMode = UNASSIGNED_ID;
```
JRoybot, line 387: do we want to clear the selected waypoint as well?

src/web/data/bots/bots.ts
```diff
@@ -115,11 +117,12 @@ class Bots {
         }
 
         if (botStatus.active_goal) {
-            missionStatus.activeGoal = botStatus.active_goal;
+            missionStatus.targetWaypoint = botStatus.active_goal;
+            missionLayer.updateFeatures();
```
JRoybot: I think this breaks an established pattern.  Earlier in this effort I thought you said we wanted to keep the data model and open layers decoupled. This is the first call to manipulate open layers directly from the data model.  I think everything else uses the `jaia-actions` and `JaiaContext` to update open layers.  

Are you sure this is how we want to do this?

src/web/style/openlayers/colors.ts
```diff
@@ -4,4 +4,5 @@ export enum OpenLayersColors {
     OUTLINE = "black",
     SELECT = "turquoise",
     TEXT = "black",
+    TARGET = "chartreuse",
```
JRoybot, line 7: who are you kidding... guys don't know chartreuse did you mean yellow ? :-D

JRoybot (changes requested):
Looks good.  One concern about updating the open layers directly from the data model

src/web/components/StartAllMissionsButton/StartAllMissionsButton.tsx
michael-jaia: `sendBotCommand` is not declared as `async` but it does return a promise

src/web/context/JaiaContext.tsx
michael-jaia: No, I just checked with the current JCC, it maintains the selected waypoint

src/web/data/bots/bots.ts
michael-jaia: Talked on the phone, made changes

src/web/utils/commands.ts
```diff
@@ -53,7 +53,7 @@ export function isCommandAvailable(commandType: CommandType, missionState: Missi
  * Passes a command message for a Bot to the jaiaAPI so it can reach the Hub for distribution
  *
  * @param {Command} command Command message to be sent to Bot
- * @returns {void}
+ * @returns {Promise} Response from sending command
```
JRoybot, line 56: I looked up and down the call chain and could not find anything that indicated this returns a promise.  I expected to see a type declaration including "Promise", an "async" function declaration or use of ".then" in a function.

Can you show me how to determine this is a promise?  (still learning....)

JRoybot (approved):

src/web/utils/commands.ts
michael-jaia: Discussed on the call

PR #1227, opened 2025-06-10
bug/2.y/autonproc-zero

aofolz:
I was running into issues building the jaiabot simulator. My autonproc was getting set to zero automatically which prevented the simulator from building. I modified build.sh to default to nproc = 1 instead of nproc = 0. This fixed the issue.


build.sh
```diff
@@ -10,21 +10,40 @@ fi
...
+
+    if [ "${MEMORY_NPROC}" -gt "${NPROC}" ]; then
+        JAIA_BUILD_NPROC=${NPROC}
     else
-        JAIA_BUILD_NPROC=$MEMORY_NPROC
+        JAIA_BUILD_NPROC=${MEMORY_NPROC}
     fi
 
-    echo "Auto nproc = $JAIA_BUILD_NPROC"
+    # Prevent 0 or negative value
```
michael-jaia: I'm wondering if we need this block. Will we catch the 0 cores problem here: 
```
if [ -z "${MEMORY_KB}" ]; then
        echo "[WARN] Unable to read memory info, defaulting to 1 process."
        MEMORY_NPROC=1
```

ferrom-jaia:
Ashley, can you add the output you are seeing when your run the build.sh?

For example, what is $NPROC equal to?

MEMORY_NPROC?
MEMORY_KB?

aofolz:
ubuntu@sufficient-rhino:~/jaiabot$ ./build.sh
[INFO] JAIA_BUILD_NPROC not set, calculating automatically...
[DEBUG] MemAvailable: 3384336 KB
[DEBUG] Memory per process: 2000000 KB
[DEBUG] Detected CPU cores: 4
[INFO] Auto nproc = 1
Configuring...
+ cmake ../..
MAJOR VERSION: 2
MINOR VERSION: 0
PATCH VERSION: 0
Project Revs Since Tag: 0

ferrom-jaia:
Thank you for the output!

Do we know the root cause of the problem? 

Can you go back to the previous code and post the output.

I would like to know why it was defaulting to 0.

aofolz:
The root of the issue was the amount of memory, CPUs, and disk space the instance has. Edited the original build.sh to output warnings if the instance did not have enough of the above. Below are outputs: One is the output of a machine that did not have enough space, the other one is a machine that had enough space so that it ran build.sh successfully. 

ubuntu@fleet-swift:~/jaiabot$ ./build.sh
Auto nproc = 0
Detected system: 1 CPU(s), 516 MiB RAM, 40333 MiB free disk
Warning: Your system has less than the required 4 CPUs to build the software.
Warning: Your system has less than the required 4 GiB of RAM.
Error: JAIA_BUILD_NPROC resolved to 0. Your system does not meet the minimum requirements to build this software.
ubuntu@fleet-swift:~/jaiabot$ 

ubuntu@fleet-swift:~/jaiabot$ ./build.sh
Auto nproc = 1
Detected system: 4 CPU(s), 3343 MiB RAM, 40387 MiB free disk
Warning: Your system has less than the required 4 GiB of RAM.
Configuring...
+ cmake ../..
MAJOR VERSION: 2
MINOR VERSION: 0
PATCH VERSION: 0
Project Revs Since Tag: 0

seantfish:
The markdown file was deleted instead of removed from the package. When looking to remove changes from a branch, use `git checkout origin/2.y -- file_to_restore.xyz` instead of removing the file (this happens with both `rm` and `git rm` as far as I know). @aofolz

build.sh
seantfish: To me, it seems that it would be better to do an explicit NPROC check since this snippet seems to just check memory. I might be wrong though.

edsanville: Yeah, the code block @michael-jaia mentioned makes sure the available memory is readable.  I think we need both blocks.

edsanville (dismissed):
Code looks good to me.

seantfish (dismissed):

ferrom-jaia (dismissed):

PR #1228, opened 2025-06-10
bug/2.y/docker-on-dev-not-needed

graceinfante:
When you're setting up the development environment you don't need Docker and it shows unnecessary errors.


scripts/setup-tools-build.sh
```diff
@@ -1,6 +1,8 @@
...
-
-script_dir=$(dirname $0)
-
-${script_dir}/setup-tools-build-nodocker.sh
-${script_dir}/setup-tools-build-docker.sh
+script_dir=$(dirname "$0")
+# Always run the non-docker setup
+"${script_dir}/setup-tools-build-nodocker.sh"
+# Only run docker setup if DEPLOY env var is set
+if [[ "$DEPLOY" == "true" ]]; then
```
ferrom-jaia: Instead of adding an argument, should we just rename the setup-tools-build-nodocker.sh to setup-tools-build.sh. This way we have two scripts:
setup-tools-build.sh
setup-tools-build-docker.sh

I do not think it is necessary to have 3 scripts. What do you think?

graceinfante: I think that makes sense.

ferrom-jaia: Sweet! When you get a chance, make those changes and push them up.

seantfish:
@graceinfante is this ready for review? (I set to draft yesterday since there were some pending changes)

ferrom-jaia (approved):
Looks great! Renaming nodocker script to setup-tools-build.sh as it is more readable.

PR #1232, opened 2025-06-12
task/add-back-rc-mode

michael-jaia:
###  Summary
Re-implemented RC mode into the JCC. The work started with establishing "modes"  to handle when the RC panel should be visible (because there are many ways to close it and I did not want the code to turn to spaghetti). After sending a command, a button will now dispatch the event `SENT_COMMAND` so the "mode" of the Bot can be updated. For the RC panel itself, we were able to reduce the amount of code by making reusable components such as `AnalogStick` that can function for throttle, rudder, or both.

### Testing
1. RC panel correctly appears when the RC button is clicked and when the stop button is clicked.
2. The Bot receives engineering commands with throttle and rudder data (verify with liaison localhost:30002 (Bot 1))

### Future Work
1. Dispatch `SENT_COMMAND` from all buttons that stop RC mode
4. Add back manual dive option
5. Add back Gamepad component for XBox compatibility
6. Highlight the Bot and button


src/web/types/jaia-system-types.ts
```diff
@@ -64,3 +64,9 @@ export enum CoordinateTypes {
     LAT = "LAT",
     LON = "LON",
 }
+
+export const enum BotModes {
+    NONE = 1,
```
JRoybot: I couldn't find any reference to `NONE` in the code.  Does it make sense that a Bot could be in a mode other than `MISSION` or `REMOTE_CONTROL`

Are the Modes limited to certain `MissionState`s ?  I.e only 
Is RC Mode going to be able to be used in Pre or Post Mission states?
If we plan to use NONE in the future would IDLE be a better name?

src/web/context/JaiaContext.tsx
```diff
@@ -515,6 +527,37 @@ function handleToggleBottomDive(mutableState: JaiaContextType) {
...
+            }
+            break;
+        case CommandType.REMOTE_CONTROL_TASK:
+            bot.setMode(BotModes.REMOTE_CONTROL);
+            break;
+        default:
+            bot.setMode(BotModes.MISSION);
+    }
+
+    mutableState.bots = bots.getBots();
```
JRoybot: isn't needed

src/web/context/JaiaContext.tsx
```diff
@@ -515,6 +527,37 @@ function handleToggleBottomDive(mutableState: JaiaContextType) {
...
+ */
+function handleSentMissionCommand(mutableState: JaiaContextType, botID: number, command: Command) {
+    const bot = bots.getBot(botID);
+
+    switch (command.type) {
+        case CommandType.MISSION_PLAN:
+            const movement = command.plan.movement;
+            if (movement === MovementType.TRANSIT) {
+                bot.setMode(BotModes.MISSION);
+            } else if (movement === MovementType.REMOTE_CONTROL) {
```
JRoybot, line 542: where is this set, I could not find it.

src/web/data/bots/bot.ts
```diff
@@ -23,6 +24,8 @@ export default class Bot {
     private wifiLinkQuality: number;
     private statusAge: number;
```
JRoybot: remove blank line

src/web/components/RemoteControlButton/RemoteControlButton.tsx
```diff
@@ -0,0 +1,127 @@
...
+}
+
+/**
+ * Packages the Bot data into a command to enter RC mode
+ *
+ * @param {Bot} bot Provides the ID and location
+ * @returns {Command} Command to enter RC mode
+ */
+function getEnterRCCommand(bot: Bot) {
+    // Bot requires location to be set if no recovery at waypoint
```
JRoybot, line 107: not sure what the comment means but this seems a little hokey to me.  If a bot can exist without a location , whatever is requiring should be able to handle it's absence.  Based on this code, it appears in this context location is irrelevant or we would need to set it to something more realistic than 0,0

src/web/components/RemoteControlPanel/RemoteControlPanel.tsx
```diff
@@ -0,0 +1,388 @@
...
+        switch (analogStickType) {
+            case AnalogStickTypes.SINGLE:
+                // Throttle
+                if (event.y > 0) {
+                    setThrottleDirection("FORWARD");
+                } else if (event.y < 0) {
+                    setThrottleDirection("BACKWARD");
+                }
+                // Rudder (with deadzone)
+                if (event.x * 100 > DEAD_ZONE_PERCENT) {
```
JRoybot: why is this logic necessary?  If the joystick movement is within the dead zone then the magnitude will be zero, if the magnitude is zero the direction is irrelevant.  Could just set it to Forward or Right or whatever direction the last sent command was.  I think it will be distracting if very small movements of the joystick changed something visible and if there is nothing visible, then even more reason to just set it to Forward/Right

src/web/components/RemoteControlPanel/RemoteControlPanel.tsx
```diff
@@ -0,0 +1,388 @@
...
+     * @param {number} position Where the user moved the analog stick
+     * @param {AnalogStickTypes} analogStickType Which analog stick moved
+     * @returns {void}
+     */
+    const setAnalogStickMagnitude = (position: number, analogStickType: AnalogStickTypes) => {
+        const absPosition = Math.abs(position) * 100;
+        const isNegative = position < 0;
+        let magnitude = 0;
+
+        if (absPosition > DEAD_ZONE_PERCENT && position < 50) {
```
JRoybot: I would like to see 50 and 95 specified as constants or even better part of an enumeration with DEAD_ZONE_PERCENT

src/web/components/RemoteControlPanel/RemoteControlPanel.tsx
```diff
@@ -0,0 +1,388 @@
...
+    /**
+     * Adds the throttle and rudder data to the Engineering command message
+     *
+     * @returns {Engineering} Throttle and rudder values to send to Bot
+     */
+    const packageCommand = () => {
+        let throttle = throttlePercentages.get(throttleMagnitude);
+        let rudder = rudderPercentages.get(rudderMagnitude);
+
+        if (throttleDirection === "BACKWARD") {
```
JRoybot, line 197: related to my comment above regarding stick movement in the dead zone.  If the logic above was removed this could be changed so it only checked if it was set to LEFT or BACKWARD and used a default of positive values

src/web/components/RemoteControlPanel/RemoteControlPanel.tsx
```diff
@@ -0,0 +1,388 @@
...
+            );
+        case ControlTypes.DIVE:
+            return <div className="remote-control-panel">{SelectMenu}</div>;
+    }
+}
+
+/**
+ * Designed to control the throttle and rudder of a Bot
+ */
+function AnalogStick(props: AnalogStickProps) {
```
JRoybot: Suggest moving this into it's own component module nested inside RemoteControlPanel (for now, could move if it ever had a use in a different panel)

src/web/components/RemoteControlPanel/RemoteControlPanel.tsx
```diff
@@ -0,0 +1,388 @@
...
+                stop={() => props.onAnalogStickStop()}
+            />
+        </div>
+    );
+}
+
+/**
+ * Allows the operator to switch between single and dual analog sticks
+ */
+function RCSelectMenu(props: RCSelectMenuProps) {
```
JRoybot: Suggest moving this into it's own component module nested inside RemoteControlPanel

src/web/components/StartMissionButton/StartMissionButton.tsx
```diff
@@ -94,11 +94,12 @@ export default function StartMissionButton(props: Props) {
                 type: CommandType.MISSION_PLAN,
                 plan: props.mission.packageMissionForHub(),
             };
-            const res = await sendBotCommand(startMissionCommand);
-            if (res.status === "ok") {
+            const response = await sendBotCommand(startMissionCommand);
+            if (response && response.status === "ok") {
```
JRoybot, line 98: I've commented on this before... prefer to see things defined as constants or enumerations instead of relying on string literals for equality checks, becomes prone to bugs by typo

Also is "===" appropriate against a literal?  We should be comparing values not locations.  Seems javascript figures it out but seems odd to me.

src/web/jcc/App.tsx
```diff
@@ -79,3 +81,20 @@ function Panel() {
             return <div></div>;
     }
 }
+
+/**
+ * Controls the rendering of the RemoteControlPanel
+ */
+function RemoteControl() {
```
JRoybot, line 86: Why is RC Panel managed differently than the others?  Is it not mutually exclusive with the others?

src/web/components/RemoteControlPanel/RemoteControlPanel.tsx
```diff
@@ -0,0 +1,388 @@
...
+                    />
+                    {SelectMenu}
+                    <AnalogStick
+                        analogStickType={AnalogStickTypes.RIGHT}
+                        handleAnalogStickMove={handleAnalogStickMove}
+                        onAnalogStickStop={onAnalogStickStop}
+                    />
+                </div>
+            );
+        case ControlTypes.DIVE:
```
JRoybot, line 255: Why not use a single joystick in Dive mode?  Is rudder movement allowed during a dive? (seems like a bad idea), if not does it make sense to display it's position in dive mode

JRoybot (changes requested):
I reviewed all the code, a few comments and questions.  Built and ran branch, tested in browser and ran all automated tests.

I was not able to get the bot to move forward or backward in single stick mode, just make it spin clockwise or counterclockwise.  

In dual stick mode, could only trigger a single stick at a time in a browser but was able to get the bot to move forward/back and to spin.

src/web/components/RemoteControlButton/RemoteControlButton.tsx
michael-jaia: It's a condition on the C++ side. It should not apply to RC mode because the operator is in control the entire time, so I just pass the default 0,0 like we did previously.

src/web/components/RemoteControlPanel/RemoteControlPanel.tsx
michael-jaia: Good call, I just took that logic from the previous code, I will investigate further.

src/web/components/RemoteControlPanel/RemoteControlPanel.tsx
michael-jaia: Yes

src/web/components/RemoteControlPanel/RemoteControlPanel.tsx
michael-jaia: I'm not totally following here, doesn't this code only check if it was set to BACKWARD or LEFT?

src/web/components/RemoteControlPanel/RemoteControlPanel.tsx
michael-jaia: Dive mode is just a start button then the state machine takes over.

src/web/components/StartMissionButton/StartMissionButton.tsx
michael-jaia: I can create a enum in one of the util files

michael-jaia: === is correct because it does a strict match (value and type)

src/web/context/JaiaContext.tsx
michael-jaia: In the RC panel component in the `packageCommand` function

src/web/jcc/App.tsx
michael-jaia: I wanted to keep the when it should render logic outside of the component like we do with the Bot and Hub details and the visible panels.

michael-jaia: It is not mutually exclusive. The RC panel can be visible while the details panel is open for example.

src/web/types/jaia-system-types.ts
michael-jaia: I was sort of using NONE as a default convention when defining enums. If it causes more confusion I can remove it. The modes are a client side convention to help us with panel renderings. For instance when a Bot does a RC dive, it is no longer in RC mission states but the RC panel should still be displayed.

michael-jaia (commented):
Jeff thanks for the review, I just responded to the comments. I need to go back and make some changes.

michael-jaia:
Not sure how my responses to your comments turned into a code review. This makes it hard to follow. Just look at the replies to your original comments.

src/web/types/jaia-system-types.ts
JRoybot: I think it does add confusion and would prefer to see it removed.  Even if the modes are only client side convention it seems there is some relationship to the different mission states.  Not sure if the relationship needs to be maintained or enforced by the code to keep things in sync

src/web/components/StartMissionButton/StartMissionButton.tsx
JRoybot: Thanks, I had a slight misunderstanding of the "===" operator

src/web/components/RemoteControlPanel/RemoteControlPanel.tsx
JRoybot: Ignore, not sure what I was thinking

src/web/types/jaia-system-types.ts
michael-jaia: Removed!

michael-jaia:
> I reviewed all the code, a few comments and questions. Built and ran branch, tested in browser and ran all automated tests.
> 
> I was not able to get the bot to move forward or backward in single stick mode, just make it spin clockwise or counterclockwise.
> 
> In dual stick mode, could only trigger a single stick at a time in a browser but was able to get the bot to move forward/back and to spin.

Commit: b396d4ced79708914a26d5409fd956725982656a cleans up the movement of the Bot. It is difficult to control the Bot with the single stick in sim because of the spinning (which exists in the current JCC).

JRoybot (dismissed):
Nice Job on hte updates

PR #1237, opened 2025-06-17
task/refactor-rally-points-part-1/JAIA-1984

michael-jaia:
### Summary
This pull request adds back the ability to place a rally point on the map. When the rally button is selected, the `mapMode` changes to `RALLY`. This is how we signal to the OpenLayers click event handler to add a rally point instead of a waypoint. This mode is reset when clicking on the rally button again, clicking a command button, or a panel button. These buttons disptach click events where we update the `mapMode` accordingly.

### Testing
1. Rally points appear on the map with numbers incrementing by 1
2. Clicking other panel buttons after having the rally button selected stops map clicks from adding rally points
3. Clicking the start mission button exits the `mapMode` from `RALLY`

### Future Work (Part 2)
1. Implement the RallyPanel
2. Make all command buttons dispatch the action `CLICKED_COMMAND_BUTTON`


src/web/components/Map/Map.tsx
```diff
@@ -58,6 +58,11 @@ export default function Map() {
             return;
         }
 
+        if (jaiaGlobal.getMapMode() === MapModes.RALLY) {
```
JRoybot: I think this should be above the if on line 56.  MapMode should supercede waypoint mode

src/web/context/jaia-actions.ts
```diff
@@ -26,5 +27,7 @@ export enum JaiaActions {
     CLICKED_EDIT_MISSION = "CLICKED_EDIT_MISSION",
     CLICKED_TAP_TO_MOVE = "CLICKED_TAP_TO_MOVE",
     CLICKED_PANEL_BUTTON = "CLICKED_PANEL_BUTTON",
+    CLICKED_COMMAND_BUTTON = "CLICKED_COMMAND_BUTTON",
```
JRoybot: hmmmm, not a fan of this method to control the map mode.  I would like to open up a broader conversation.

I feel this is going to lead to unmanageable workflow when we implement more of the command buttons.  My concern is those buttons are already complicate enough since they launch dialogs with additional buttons that will dispatch events to the Context.  I think whatever action results from the command button and it's associated dialog should result in a single event dispatch and the handler for that event set the map mode accordingly.  

The number events we handle in this reducer has grown considerably since we started and I am concerned that we will have so many events and handlers and independent state variables that in some cases could be in illogical combinations it will be impossible to avoid putting the variables in illogical states.

This probably should be managed as a true state machine with defined states that define the combination of what we have in these independent state variables.

src/web/utils/style.ts
```diff
@@ -106,3 +112,15 @@ export function scrollMissionsList() {
         }
     }, SCROLL_DELAY);
 }
+
+/**
+ * Sets the cursor type for the map
+ *
+ * @param {Cursor} cursor Type of cursor to display on map
+ */
+export function setMapCursor(map: Map, cursor: Cursors) {
```
JRoybot: I think this should be integrated into jaia-global.ts since it is Map related.  I don't like common utility modules depending on large component modules like Map.  Should only be the other way around.  It should probably be integrated with mapMode at the same time

src/web/context/JaiaContext.tsx
```diff
@@ -515,6 +530,21 @@ function handleToggleBottomDive(mutableState: JaiaContextType) {
...
+ *
+ * @param {JaiaContextType} mutableState State object ref for making modifications
+ * @param {GeographicCoordinate} location Where to add the rally point
+ * @returns {JaiaContextType} Updated mutable state object
+ */
+function handleAddRallyPoint(mutableState: JaiaContextType, location: GeographicCoordinate) {
+    rallyLayer.addRallyPoint(location);
+    jaiaGlobal.setMapMode(MapModes.NONE);
+    mutableState.mapMode = jaiaGlobal.getMapMode();
+    setMapCursor(map, Cursors.DEFAULT);
```
JRoybot: I think changing the cursor style should be part of setMapMode and should be part of the jaia-global state

src/web/openlayers/layers/vector/zindex.ts
```diff
@@ -5,3 +5,4 @@ export const layersZIndexes = new Map<LayerTitles, number>();
 layersZIndexes.set(LayerTitles.BOT_LAYER, 3);
 layersZIndexes.set(LayerTitles.HUB_LAYER, 2);
 layersZIndexes.set(LayerTitles.MISSION_LAYER, 1);
+layersZIndexes.set(LayerTitles.RALLY_LAYER, 1);
```
JRoybot: shouldn't all our layers have distinct ZIndex s?

src/web/openlayers/layers/vector/rally-layer.ts
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * Places the rally icon on the map layer
+     *
+     * @param {GeographicCoordinate} location Where to place the rally point
+     * @returns {void}
+     */
+    addRallyPoint(location: GeographicCoordinate) {
+        const rallyNum = this.getNextRallyNum();
+        this.getVectorLayer().getSource().addFeature(generateRallyFeature(location, rallyNum));
+        this.rallyNums[rallyNum - 1] = rallyNum;
```
JRoybot, line 29: I think Python is the only language I have seen that got arrays right but starting index at 1 instead of 0 so we don't have to do this every time we want to access a numbered array element.

src/web/types/openlayers-types.ts
```diff
@@ -14,3 +15,9 @@ export enum MapFeatureTypes {
     WAYPOINT = "WAYPOINT",
     WAYPOINT_LINE = "WAYPOINT_LINE",
 }
+
+export enum MapModes {
+    NONE = "NONE",
```
JRoybot: I think it means we are either going to select an existing feature or create a new waypoint... maybe a better name would be "DEFAULT", NONE implies we aren't doing any map interactions

JRoybot (commented):
I reviewed the code and have a bunch of high level comments and concerns.

src/web/context/JaiaContext.tsx
```diff
@@ -67,6 +72,7 @@ export interface JaiaAction {
     botAccordionName?: BotAccordionNames;
     mapLayerAccordionName?: MapLayerAccordionNames;
     panelName?: PanelNames;
+    mapMode?: MapModes;
```
JRoybot: Assuming this stays it should be initialized in the constructor.

src/web/context/JaiaContext.tsx
```diff
@@ -734,6 +764,47 @@ function handleClickedPanelButton(mutableState: JaiaContextType, panelName: Pane
...
+    return mutableState;
+}
+
+/**
+ * Handles a click to a command button. Used to reset map modes.
+ *
+ * @param {JaiaContextType} mutableState State object ref for making modifications
+ * @returns {JaiaContextType} Updated mutable state object
+ */
+function handleClickedCommandButton(mutableState: JaiaContextType) {
```
JRoybot: Seems like this could be handled by the function below

src/web/components/Map/Map.tsx
JRoybot: Better yet use a switch statement on MapMode

src/web/context/JaiaContext.tsx
michael-jaia: (in JaiaGlobal constructor)

src/web/components/Map/Map.tsx
michael-jaia: Moved map mode ahead of waypoint mode. Switch statement will be added with the measure tool.

src/web/openlayers/layers/vector/zindex.ts
michael-jaia: Yeah it will help remove any ambiguity.

src/web/context/jaia-actions.ts
ferrom-jaia: Ya, that sounds like a good plan. We could use one of these state machine libs to help us: https://stately.ai/docs/xstate, https://github.com/matthewp/robot, https://www.statedesigner.com/

ferrom-jaia: Maybe Redux can help us here as well

ferrom-jaia: We should create a ticket for this. This is something we can add after the refractor.

src/web/context/JaiaContext.tsx
```diff
@@ -201,14 +202,8 @@ function jaiaReducer(state: JaiaContextType, action: JaiaAction) {
...
-        case JaiaActions.CLICKED_PANEL_BUTTON:
-            return handleClickedPanelButton(mutableState, action.panelName);
-
-        case JaiaActions.CLICKED_COMMAND_BUTTON:
-            return handleClickedCommandButton(mutableState);
-
-        case JaiaActions.CLICKED_MAP_MODE_BUTTON:
-            return handleClickedMapModeButton(mutableState, action.mapMode);
+        case JaiaActions.CLICKED_BUTTON:
+            return handleClickedButton(mutableState, action.buttonType, action.buttonName);
```
JRoybot, line 206: Much nicer!

src/web/context/JaiaContext.tsx
```diff
@@ -737,60 +731,37 @@ function handleClickedTapToMove(mutableState: JaiaContextType) {
...
     }
 
-    jaiaGlobal.setMapMode(MapModes.NONE);
-    mutableState.mapMode = jaiaGlobal.getMapMode();
-    setMapCursor(map, Cursors.DEFAULT);
-
-    return mutableState;
-}
+    jaiaGlobal.setMapMode(mapMode);
+    setOpenLayersCursor(cursor);
```
JRoybot, line 761: why don't we manage the cursor in jaiaGlobal like the mapMode?

JRoybot (dismissed):
This looks good, one question about the cursor style to make sure I understand

PR #1243, opened 2025-06-19
task/refactor-data-offload-queue

michael-jaia:
### Summary
Adds back data offload all button and the data offload queue. The previous implementation included all of the queue functionality on the web side since it was developed prior to the queue being built on the Hub side. Now we use the queue information from the Hub and display it in the JCC. This reduces logic on the web side and creates a better user experience for multiple clients.

### Kaizen
1. Moves the duplicated accordion theme that removes transition to a utility file so it can be used by all of the accordions.
2. The last data offload that completes will hold on to the percentage. If that percentage is 100, we stop displaying it.

### Testing
image


JRoybot (dismissed):
Did a static review of the code, all looks good.  Nice job!
Did not build or run test due to base branch conflict with my system

JRoybot (approved):

PR #1244, opened 2025-06-20
task/update_working_with_jaia_software

graceinfante:
updated page120 with Twomey's formatting suggestions.


src/arduino/jaiabot_runtime/jaiabot
```diff
@@ -0,0 +1 @@
+/home/ubuntu/jaiabot/build/arm64/include/jaiabot
```
ferrom-jaia: Do we need this?

src/doc/markdown/notes.txt
```diff
@@ -0,0 +1,29 @@
+1. Troubleshooting vs Trouble Shooting
```
ferrom-jaia: Do we need this file?

src/doc/markdown/page120_working_with_jaia_software.md
```diff
@@ -1,124 +1,109 @@
...
 ```
-6. Launch the simulator in a separate terminal.
+2. Launch the simulator in a separate terminal.
 ```
 cd jaiabot/config/launch/simulation
 # Set the simulation to run 4 bots at a time warp of 5
 ./generate_all_launch.sh 4 5
 ./all.launch
 # Should look like: [all] XX:XX:XX: All processes running
 ```
```
ferrom-jaia: Can we split the above into the different commands? This will allow the user to use the copy button (instead of highlighting and copying)
Move the comments outside of the code blocks.

ferrom-jaia (changes requested):
Looks good!

What do you think about:

Can we split the code blocks into the different commands? This will allow the user to use the copy button (instead of highlighting and copying)
Move the comments outside of the code blocks.

ferrom-jaia:
What do you think about the titles below?

Setting Up the Operating System
Preparing the Workspace
Launching the Simulator
Troubleshooting the Simulator
Modifying the Code
Creating an SSH Key Pair (MacBooks)
Deploying the Code
Debugging Deployment Issues

ferrom-jaia (approved):
Looks good!

PR #1246, opened 2025-06-23
Rebase developer friendly jcc on 2.y

JRoybot:
### Summary
This pull request adds the `developer-friendly-jcc` changes on top of `2.y`. `developer-frindely-jcc` started on `1.y`. As we near the end of the life of `1.y`, it is time to make the switch. 

**This branch should not be merged into 2.y until all of the developer-friendly-jcc tasks are complete.**  Target branch of the PR is just for reference.  Instead this branch will become the feature integration branch for the developer-friendly-jcc Epic, replacing the 1.y based developer-friendly-jcc branch


src/web/components/Map/__tests__/Map.test.tsx
```diff
@@ -59,13 +59,15 @@ test("Select and deselect Bot and Hub icons on map", () => {
     // Bot
     bots.setBot(botStatusMock1);
     act(() => {
-        map.dispatchEvent(mapBrowserEventMock);
+        const mockEvent = createMapBrowserEventMock();
```
michael-jaia: Can we create the mock event once and then use the same one throughout the different tests?

src/web/tests/setup-tests.ts
```diff
@@ -1 +1,30 @@
 import "@testing-library/jest-dom";
+
+// setupTests.ts
+if (typeof PointerEvent === "undefined") {
```
michael-jaia: Do you still have the error output from not having this code? I follow most of it, but the last line seems a little hand wavy. I'm wondering if there is an alternative path that reads a little smoother.

michael-jaia (commented):
Hey Jeff, just left a couple comments on some of the changes related to the map tests.

src/web/tests/setup-tests.ts
JRoybot: I tried some simpler solutions but kept running into one more issue .... I got this solution from chatGPT, there may be a better solution but I do not know how to do it.

Simpler mocks ran into issues setting parameters because the real PointerEvent is a subclass of MouseEvent, which has a lot read-only parameters, which could not be overridden... If you want more I can share the entire chatGPT conversation

JRoybot: This is the error simpler mocks ran into...
```TypeError: Cannot set property clientX of [object MouseEvent] which has only a getter
        at Function.assign (<anonymous>)

       5 |     constructor(type: string, props: PointerEventInit = {}) {
       6 |       super(type, props);
    >  7 |       Object.assign(this, props);
         |              ^
       8 |     }
       9 |   }
      10 |

      at new MockPointerEvent (tests/setup-tests.ts:7:14)
      at createMapBrowserEventMock (tests/__mocks__/openlayers/events/map-browser-click.mock.ts:60:25)
      at components/Map/__tests__/Map.test.tsx:62:52
      at callback (node_modules/react/cjs/react.development.js:2512:16)
     ```

File: src/web/components/Map/__tests__/Map.test.tsx (hunk 1 above)
Comment (JRoybot) [reply src/web/components/Map/__tests__/Map.test.tsx]: Funny I had asked that question to chatGPT when it suggested this... it's response was 
```Why you should not reuse the same event:
Event objects can be mutated internally by OpenLayers or your app logic (e.g. .stopPropagation(), .preventDefault(), or metadata set during propagation).

React or OpenLayers event handlers may treat reused events as "already handled", especially if any internal flags are set.

Stateful mocks become brittle — if one test mutates the shared event, other tests may behave unpredictably or fail.

```

PR #1249, opened 2025-06-24
task/2.y/dev-friendly/saving-mission-part-1/JAIA-1865

JRoybot:
This branch adds functionality to save & load missions to/from localStorage.  Functions to save and load individual missions, save and load mission sets, list missions in a mission set and delete a mission from a mission set were created.

Unit tests were created to exercise and verify the new functionality.

During development we identified confusion due to the overloaded use of the term `missions` so used `MissionSet` to refer to the class, `missionSet` to refer to the singleton of the class and kept with `missions` where referring to internal Maps of missions

This is only a partial implementation of the Save and Load Mission panels, the intent is to merge in the name changes before proceeding with the React code to create the panels


src/web/data/missions/missionSet.ts
michael-jaia, line 1: Let's call this `mission-set.ts`. I also think we should re-name the folder to `mission_set`. To clarify the naming conventions, the files will take the structure `hello-world` to match the other `ts` files and the directories will use `hello_world` to match the other directories in the project.

src/web/data/missions/mission.ts
```diff
@@ -7,12 +7,14 @@ import {
...
 } from "../../types/protobuf-types";
 import Waypoint from "../waypoints/waypoint";
+import Task from "../tasks/task";
 
 export default class Mission {
     private missionID: number;
     private waypoints: Waypoint[];
     private speeds: Speeds;
     private repeats: number;
+    private name: string;
```
michael-jaia: Can you bump up name above waypoints to match the order of the methods in the file

src/web/data/missions/mission.ts
```diff
@@ -115,4 +125,23 @@ export default class Mission {
 
         return goals;
     }
+
+    /**
+     * Creates a mission object from serialized mission data
+     *
+     * @param {any} serializedMission Serialized Mission Data to transform to mission object
```
michael-jaia: NIT: lowercase D in data and capitalize M in mission object. Same thing in the @returns line

src/web/data/missions/mission.ts
```diff
@@ -115,4 +125,23 @@ export default class Mission {
...
+     * @returns {Mission} mission Resulting mission object
+     */
+    static fromJSON(serializedMission: any): Mission {
```
michael-jaia: Nice function, super easy to follow!

src/web/data/missions/missionSet.ts
```diff
@@ -0,0 +1,213 @@
...
+import { Speeds, GeographicCoordinate } from "../../types/protobuf-types";
+
+class MissionSet {
+    private missions: Map<number, Mission>;
+    private nextMissionID: number;
+    private missionIDInEditMode: number;
+    private missionSpeeds: Speeds;
+    private name: string;
+
+    constructor() {
```
michael-jaia: Should we have a default mission name even if it is just an empty string?

src/web/data/missions/missionSet.ts
```diff
@@ -0,0 +1,213 @@
...
+
+    /**
+     * Saves an individual mission to local storage
+     *
+     * @param {string} name Name to use when storing mission in local storage
+     * @param {number} missionID Mission ID of mission to be saved
+     *
+     * @notes not used at this time, provided for future capabilities
+     */
+    saveMission(name: string, missionID: number) {
```
michael-jaia: I think it might be clearer to name these function `saveMissionLocalStorage`, `loadMissionLocalStorage`, and `saveMissionSetLocalStorage`. I know it adds some length but I think it will be clearer to new developers.  I'm almost wondering if we want the local storage functions in a separate file to avoid the near name clashes.

src/web/data/missions/missionSet.ts
```diff
@@ -0,0 +1,213 @@
...
+        }
+    }
+
+    deleteAllMissions() {
+        this.missions.clear();
+        this.setMissionIDInEditMode(UNASSIGNED_ID);
+        this.setNextMissionID(1);
+    }
+
+    /**
```
michael-jaia: NIT: Missing `@returns {void}`

src/web/data/missions/missionSet.ts
```diff
@@ -0,0 +1,213 @@
...
+
+        const newMission = Mission.fromJSON(JSON.parse(jsonMission));
+        // Save to the mission set
+        this.addMission(newMission);
+    }
+
+    /**
+     * Saves all the current missions as a mission set to local storage
+     *
+     * @param {string} name : Name to use for storing the mission set
```
michael-jaia: NIT: Remove `:`

src/web/data/missions/missionSet.ts
```diff
@@ -0,0 +1,213 @@
...
+     */
+    saveMissionSet(name: string) {
+        // Apply the name to the mission set and all the missions
+        this.setName(name);
+        this.missions.forEach((mission) => {
```
michael-jaia: Do the individual missions need a name?

src/web/data/missions/missionSet.ts
```diff
@@ -0,0 +1,213 @@
...
+            mission.setName(name);
+        });
+        // Convert missions Map to an array before using stringify
+        const missionsArray = Array.from(this.missions.entries());
+        // Read the saved mission sets from localStorage (or start fresh)
+        const existing = JSON.parse(localStorage.getItem("missionSets") || "{}");
```
michael-jaia: Can we call this variable missionSets?

src/web/data/missions/missionSet.ts
```diff
@@ -0,0 +1,213 @@
...
+        // Save to the mission set
+        this.addMission(newMission);
+    }
+
+    /**
+     * Saves all the current missions as a mission set to local storage
+     *
+     * @param {string} name : Name to use for storing the mission set
+     */
+    saveMissionSet(name: string) {
```
michael-jaia: Should we save each mission set as its own key?

src/web/data/missions/missionSet.ts
```diff
@@ -0,0 +1,213 @@
...
+    /**
+     * Replaces the current mission set with those from a saved set
+     *
+     * @param {string} name String Name of saved set to retrieve
+     * @returns {boolean} False if the mission set was not found
+     */
+    loadMissionSet(name: string) {
+        // Get all saved mission sets
+        const allSets = JSON.parse(localStorage.getItem("missionSets") || "{}");
+        const data = allSets[name];
```
michael-jaia: Maybe we name this targetMissionSet?

michael-jaia (commented):
Hey Jeff, I just did a walk through of the PR. I left a few comments and open-ended questions to get your thoughts. Nice work.

src/web/data/missions/missionSet.ts
JRoybot: If you feel they do I could append the mission ID to the mission-set name when storing them...  I don't think we want to allow them to be named individually through the interface, I think that would make things overly complicated.

src/web/data/missions/missionSet.ts
JRoybot: I was under the impression that the line `missionSets[name] = {` is using the name as a key.... I could be wrong.  If you look at the test, I tested saving multiple sets and retrieving them by name.

src/web/data/missions/missionSet.ts
michael-jaia: I was thinking we could just remove the name property from the Mission class

src/web/data/missions/missionSet.ts
michael-jaia: This line: `localStorage.setItem("missionSets", JSON.stringify(missionSets));` in `saveMissionSet` looks to me like it writes all the mission sets instead of just the one being saved.

michael-jaia: But that may be a good thing to avoid naming collisions between other parts of the code that may use local storage.

src/web/data/mission_set/mission-set.ts
```diff
@@ -0,0 +1,218 @@
...
+
+        // Rebuild mission set from saved entries
+        if (Array.isArray(targetMissionSet.missions)) {
+            targetMissionSet.missions.forEach(([id, missionObj]: [number, any]) => {
+                const mission = Mission.fromJSON(missionObj);
+                this.addMission(mission);
+            });
+        }
+
+        // Restore other fields
```
michael-jaia: Should we also restore nextMissionID and missionIDInEditMode?

src/web/data/missions/missionSet.ts
JRoybot: Sure, I don't see why we need it...

src/web/data/mission_set/mission-set.ts
JRoybot: That is done in the call to deleteAllMissions

src/web/data/missions/missionSet.ts
JRoybot: mission name was part of left over code to load and store individual missions instead of mission sets.  Logic was removed

src/web/data/mission_set/mission-set.ts
michael-jaia: Do we want also want to do it on the load. For instance, if we load a mission set with missions 5 and 6, I think we want nextMissionID to be 7.

JRoybot: That is handled by the call to addMission

PR #1250, opened 2025-06-24
bug/ec-driver-warning-in-sim

edsanville:
So, the pipeline during runtime (non-simulation) for salinity data goes like this:
1.  `jaiabot_as-ezo-ec.py` sends the comma-delimited data over UDP
2.  `jaiabot_atlas_scientific_ezo_ec_driver` receives the comma-delimited data, parses it, then publishes the salinity message over Goby

During simulation, the simulated comma-delimited line of data is missing one of the new fields, and so the  `jaiabot_atlas_scientific_ezo_ec_driver` cannot process it.

I added in the missing field during simulation.

### DIFF
--- a/src/bin/simulator/simulator.cpp
+++ b/src/bin/simulator/simulator.cpp
@@ -424,8 +424,8 @@ void jaiabot::apps::SimulatorTranslation::process_nav(const CMOOSMsg& msg)
         salinity += salinity_distribution_(generator_);
 
         auto msg = jaiabot::protobuf::SalinityData();
+        // We only set the raw values here, because the derived values are calculated elsewhere, after the data comes in from the sensor.
         msg.set_conductivity_raw(45000.0);
-        msg.set_salinity_raw(salinity);
         msg.set_salinity_raw(salinity);
         msg.set_total_dissolved_solids(0.0);
 
--- a/src/bin/simulator/simulator.cpp
+++ b/src/bin/simulator/simulator.cpp
@@ -49,6 +49,7 @@
 #include "jaiabot/messages/jaia_dccl.pb.h"
 #include "jaiabot/messages/low_control.pb.h"
 #include "jaiabot/messages/sensor/pressure_temperature.pb.h"
+#include "jaiabot/messages/sensor/salinity.pb.h"
 #include "jaiabot/messages/simulator.pb.h"
 #include <goby/middleware/gpsd/groups.h>
 #include <goby/middleware/protobuf/gpsd.pb.h>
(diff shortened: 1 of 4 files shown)


JRoybot (dismissed):
Looks goof!, verified in browser

edsanville:
Re-requesting review.  The root of the problem was actually a missing field in the simulated comma-delimited line of data from `jaiabot_simulator`.  So, I added the missing field, and added the health check back in.

JRoybot (dismissed):
Even better!!

src/bin/simulator/simulator.cpp
```diff
@@ -426,10 +426,11 @@ void jaiabot::apps::SimulatorTranslation::process_nav(const CMOOSMsg& msg)
         std::string time = "";
         std::string conductivity = "45000.0";
         std::string dissolved_solids = "0.0";
+        std::string specific_gravity = "0.0";
 
         // date_string, conductivity, dissolved solids, salinity
         ss << std::setprecision(std::numeric_limits<double>::digits10) << time << ","
-           << conductivity << "," << dissolved_solids << "," << salinity;
+           << conductivity << "," << dissolved_solids << "," << salinity << "," << specific_gravity;
```
ferrom-jaia: I think we are missing conductivity_raw

ferrom-jaia (changes requested):
We want to add conductivity_raw I think

edsanville:
> We want to add conductivity_raw I think

Shall we move to a protobuf message while we're at it, or keep it comma-delimited for now?

ferrom-jaia:
> > We want to add conductivity_raw I think
> 
> Shall we move to a protobuf message while we're at it, or keep it comma-delimited for now?

Yes, that seems like the right strategy

edsanville:
Comms between the python and C++ driver now use the SalinityData protobuf message, rather than a comma-delimited line of text.

src/bin/simulator/simulator.cpp
```diff
@@ -422,17 +423,14 @@ void jaiabot::apps::SimulatorTranslation::process_nav(const CMOOSMsg& msg)
...
-        std::string conductivity = "45000.0";
-        std::string dissolved_solids = "0.0";
-
-        // date_string, conductivity, dissolved solids, salinity
-        ss << std::setprecision(std::numeric_limits<double>::digits10) << time << ","
-           << conductivity << "," << dissolved_solids << "," << salinity;
+        auto msg = jaiabot::protobuf::SalinityData();
+        msg.set_conductivity_raw(45000.0);
+        msg.set_salinity_raw(salinity);
+        msg.set_salinity_raw(salinity);
```
ferrom-jaia: Is one of these suppose to be salinity? (set_salinity, set_salinity_raw)

edsanville: Whoops!  Actually, we only set the raw values here, because the code that calculates the derived values is elsewhere.  It's just an accidental duplicate line.

ferrom-jaia (approved):
Looks good! Tested in sim!

PR #1253, opened 2025-06-25
task/add-back-rally-panel

michael-jaia:
### Summary
Builds off new rally point implementation to add a panel containing three buttons:
1) Go To Rally Point
2) Delete
3) Close
Previously sending Bots to a rally point would delete all missions. This implementation intentionally does not enact that behavior so an operator can quickly move all Bots without having to recreate their current mission set.

### Considerations
Prior to creating the `GoToRallyButton` I tried using `StartAllMissionsButton` but incorporating the rally logic and text made the component too clunky.

### Testing
1) Selecting a rally point displays the correct panel
2) Clicking the delete button removes it from the map and closes the panel
3) Clicking the Go To Rally Point button leads produces the correct dialog box and when confirmed sends the the location to the Bots for transit.

image


src/web/context/jaia-actions.ts
```diff
@@ -16,9 +16,12 @@ export enum JaiaActions {
     CHANGE_TASK_PARAMETER = "CHANGE_TASK_PARAMETER",
     TOGGLE_BOTTOM_DIVE = "TOGGLE_BOTTOM_DIVE",
     ADD_RALLY_POINT = "ADD_RALLY_POINT",
+    DELETE_RALLY_POINT = "DELETE_RALLY_POINT",
```
JRoybot, line 19: Rather than separate actions for each button could we use an action parameter to specify which Rally Button was pressed similar to CLICKED_BUTTON actions?  Could this all be handled by CLICKED_BUTTON and dispatched to the appropriate handler based on the parameter in the action?

src/web/context/JaiaContext.tsx
```diff
@@ -520,11 +535,41 @@ function handleToggleBottomDive(mutableState: JaiaContextType) {
  */
 function handleAddRallyPoint(mutableState: JaiaContextType, location: GeographicCoordinate) {
     rallyLayer.addRallyPoint(location);
+    setOpenLayersCursor(Cursors.DEFAULT);
```
JRoybot, line 579: I understand why setting the mapMode and Cursor needs to be done in 2 functions but since each mapMode has a defined Cursor we could wrap these calls in a function in this module that called both of these functions insuring we always set the cursor correctly for each mapMode

src/web/components/GoToRallyButton/GoToRallyButton.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+import Bot from "../../data/bots/bot";
+
+import { SelectedRallyPoint } from "../../types/jaia-system-types";
+import {
+    Command,
+    CommandType,
+    MissionStart,
+    MovementType,
+    Speed,
```
JRoybot, line 20: Speed is not used

src/web/components/GoToRallyButton/GoToRallyButton.tsx
```diff
@@ -0,0 +1,188 @@
...
+        for (const [botID, bot] of props.bots.entries()) {
+            if (isCommsDropped(bot.getStatusAge())) {
+                updatedBotReadyStates.get(DisabledCodes.NO_COMMS).push(botID);
+            } else if (
+                !isCommandAvailable(CommandType.START_MISSION, bot.getMissionStatus().missionState)
+            ) {
+                updatedBotReadyStates.get(DisabledCodes.MISSION_STATE).push(botID);
+            }
+
+            // Download queue
```
JRoybot, line 66: Fix comment

src/web/context/JaiaContext.tsx
```diff
@@ -520,11 +535,41 @@ function handleToggleBottomDive(mutableState: JaiaContextType) {
...
+ * Makes call to delete a rally point from the rally layer
+ *
+ * @param {JaiaContextType} mutableState State object ref for making modifications
+ * @returns {JaiaContextType} Updated mutable state object
+ */
+function handleDeleteRallyPoint(mutableState: JaiaContextType) {
+    rallyLayer.deleteRallyPoint(mutableState.selectedRallyPoint.id);
+    mutableState.selectedRallyPoint = { id: UNASSIGNED_ID };
+    mutableState.visiblePanel = ButtonNames.NONE;
+    return mutableState;
```
JRoybot, line 595: why aren't any map layers refreshed?

src/web/openlayers/layers/vector/rally-layer.ts
JRoybot, line 1: This layer is managed very differently than others... why aren't we keeping the Rally Point data in the data model somewhere and have this layer pull data from there?  

Shouldn't all layers have a zIndex assigned?

JRoybot (commented):
Had time to do a static review of the code, did not build or test it.  Some basic comments about the way the Rally Points are handled.

src/web/context/JaiaContext.tsx
michael-jaia: rallyLayer.deleteRallyPoint makes a call to update the layer directly

src/web/openlayers/layers/vector/rally-layer.ts
michael-jaia: We could, I was not sure if it was worth the overhead. Looking at it now, we should put them in the data model. The rally z-index came in the first rally PR.

PR #1254, opened 2025-06-26
task/2.y/add-back-comms-circles

michael-jaia:
### Summary
This pull request adds back the circles around the Hub that display its communication range with the Bots.

### Testing
Circles can be viewed in simulation. The circles maintain the correct size when the map zoom level changes.
image


JRoybot (dismissed):
All looks good!

PR #1257, opened 2025-07-01
bug/jdv-delete-jaia-1167

edsanville:
Also add dataclasses_json as a dependency for the venv


JRoybot (dismissed):
Looks good

ferrom-jaia:
Is there an apt package we could use for this dependency: dataclasses-json

edsanville:
> Is there an apt package we could use for this dependency: dataclasses-json

Apparently yes, which is surprising.  Will switch to that one.

JRoybot (dismissed):

ferrom-jaia (dismissed):
Nice. Button is disabled if no log is selected to delete.

PR #1258, opened 2025-07-01
bug/jdv-graph-zoom-should-only-display-task-packets-within-time-window-jaia-503

edsanville:
Only display task packets from within the timeRange window.


JRoybot (approved):
All Looks good

ferrom-jaia (approved):
Nice! Tested in JDV.

Mission had multiple task packets.

Zoomed into multiple time zones and the task packets displayed as expected.

PR #1241, opened 2025-06-18
task/2.y/add-config-option-for-camera-jaia-1919

edsanville:
Camera config option, targeted to the `2.y` branch.

PR #1261, opened 2025-07-02
Add socat to runtime tools

tsaubergine:
This is causing the docker sim to fail.

Also in https://github.com/jaiarobotics/jaiabot/pull/1241/files#diff-78917d91dba0eafcab2a0e0dcf8a1fe147f83959ae016f04e3a1d781cb6f13ddL6 but we can easily fix this now rather than wait until that is reviewed.


ferrom-jaia (approved):
Looks good!

PR #1264, opened 2025-07-07
feature/send-botstatus-to-tak-server

OverArchMC:
Sends BotStatus and TaskPacket messages to the TAK interface via CoT messages.


src/python/requirements.txt
seantfish, line 1: It looks like the requirements.txt is organized alphabetically, not sure if it matters.

PR #1265, opened 2025-07-07
First Waypoint Start Fix 

bengulezian:
Changed so the simulation would start from first waypoint and not the last.


seantfish:
@bengulezian Is this still planned?

michael-jaia:
Closing until code builds without errors.

PR #1273, opened 2025-07-11
task/prevent-credential-push

seantfish:
This PR introduces entries in the main .gitignore to prevent addition of files with extensions known to be associated with secret credentials. It also adds a pre-commit husky hook to prevent committing files containing certain phrases (case-sensitive) common to secret credential files.

How to test:
- Check text files with certain phrases are prevented from being committed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,11 +17,8 @@ __pycache__
 
 # SSH keys
 id_ecdsa
-id_ecdsa.pub
 id_ed25519
-id_ed25519.pub
 id_rsa
-id_rsa.pub
 
 # GPG keys
 *.gpg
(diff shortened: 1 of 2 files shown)


tsaubergine:
I don't see any reason to block committing public keys.

id_ecdsa.pub
id_ed25519.pub
id_rsa.pub

seantfish:
> I don't see any reason to block committing public keys.
> 
> id_ecdsa.pub id_ed25519.pub id_rsa.pub

Good catch, I agree. Mindlessly added those to the list. I'll remove them.

seantfish:
Removed the GPG ignores too.

ferrom-jaia (approved):
Looks good!

PR #1278, opened 2025-07-17
feature/detect-bot-tail-overheating

OverArchMC:
Detects if the thermistor temperature has exceeded 100 degrees Celsius. If above that threshold, the bot enters IN_MISSION__PAUSE__RESOLVE_BOT_TAIL_OVERHEATING state and waits for temperature to drop below the threshold before resuming its mission.

### DIFF
--- a/.husky/_/.gitignore
+++ b/.husky/_/.gitignore
@@ -1,2 +1,1 @@
 *
-!.gitignore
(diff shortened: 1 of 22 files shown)


ferrom-jaia:
Closing for now, no clear need on this currently

PR #1267, opened 2025-07-08
feature/xbee-simulator-jaia-2019

seantfish:
## Summary

This pull request introduces a Python-based XBee simulator to enable XBee-based networking in the Jaiabot simulation.

PR #1295, opened 2025-08-05
Task/2.y/implement hub to hub intervehicle link/jaia-2038/jaia-2042

tsaubergine:
**This PR builds off the XBee Simulator PR (https://github.com/jaiarobotics/jaiabot/pull/1267) so it would be best to merge that first**

# Hub2Hub comms link

- Use UDP driver to implement an intervehicle pub/sub link with messages sent to the other hubs in the inventory file (over the CloudHub VPN for real systems or localhost for simulated ones). The message used is [Hub2HubData](https://github.com/jaiarobotics/jaiabot/blob/task/2.y/implement-hub-to-hub-intervehicle-link/jaia-2038/src/lib/messages/jaia_dccl.proto#L618-L640). The only implemented `contents` field is  sending the BotStatus, the rest are placeholders for future work.

# Remove hub_id from drivers, Rework subscription logic
 
Based on Silent Swarm testing, the XBee radios work fine by spoofing the same serial ID for multiple radios. We can then use the same serial number for all hubs. This PR takes advantage of that feature to substantially simplify the subscription logic and general multi-hub logic.

- For local (XBee / Wifi) links, the bots will subscribe (Command, Engineering messages) to the hub(s) on start, **and every 1 minute (configurable) thereafter (new!)**. One or more hubs will receive and acknowledge this message. If a hub doesn't receive it, it will receive a later subscription when in range. Thus, the longest time to wait for a subscription when a bot comes in range is 1 minute. For Iridium (and future Cellular) links, we will subscribe to just the CloudHub, just on startup. This reduces excess data usage (which would lead to substantial airtime charges for Iridium), and is acceptable given that: 1) we only have one hub (CloudHub), and 2) we can fairly confidently assert that we don't need to reboot the CloudHub during operations.
- The intervehicle (re)subscription logic (on the bots) is written into `jaiabot_comms_manager` (new!)
- When a hub receives a subscription (to Command), it will reply with its subscriptions for the bots (BotStatus, TaskPacket, Engineering messages). as the hubs currently do. **Hubs will no longer subscribe to bots on startup based on the inventory file (new!)** . **Hubs will no longer persist their subscriptions across reboots (new!)**. These simplifications are both enabled by the fact that all bots will regularly subscribe to the (only) Hub Xbee serial ID. This also means that hubs will not spend XBee radio comms bandwidth sending subscriptions over and over for 1 hour to bots that are not currently in use (but are in the Inventory). This cost savings is obviously offset partially by the continual resubscribe loop, but this only 2 messages per bot per minute (1 from the bot, 1 from the hub).
- The Hub that is "in control" is now defined as the last hub to send a Command message. The Hub ID is now simply included in the Command (new!).

This sequence diagram shows this new behavior (where N=1 minute currently):

# Minor

- Hub log directory moved to `/var/log/jaiabot/hub/<hub_id>` from `/var/log/jaiabot/hub`. This makes for cleaner logging on single-machine simulation and will set up for easier log management from multiple hubs. This mirrors how the bot data are stored.

# Testing

## Testing performed

- Simulation with Wifi
  - Two hubs, two bots, launched together - WORKS
  - Start with one hub, add another hub later - WORKS
  - Reboot hub - WORKS
  - Start with one bot, add another bot later - WORKS
  - Reboot bot - WORKS
- Simulation with XBee
  - Two hubs, two bots, launched together - WORKS
  - Start with one hub, add another hub later - WORKS
  - Reboot hub - WORKS
  - Start with one bot, add another bot later - WORKS
  - Reboot bot - WORKS
- Flat fleet with XBee
  -  Two hubs, two bots, 1 CloudHub launched asynchronously - WORKS
  - Reboot hub - WORKS
  -  Reboot bot - WORKS

Cloudhub receiving hub2hub data from other real hubs:
image

## Needs testing
- In water with XBee.
- 
# Documenting
- Updated [Communications Markdown](https://github.com/jaiarobotics/jaiabot/blob/task/2.y/implement-hub-to-hub-intervehicle-link/jaia-2038/src/doc/markdown/page50_communications.md) with changes.

# Future work

See Epic: https://jaia-innovation.atlassian.net/browse/JAIA-2037


tsaubergine:
Still needs testing on hardware, but is otherwise ready for review given that it has been tested in simulation for both wifi and Xbee.

ferrom-jaia (approved):
Looks great! Tested on the bench with two bot and two hubs. Before this is released it will be use in the field with 5+ bots and three hubs on.

PR #1297, opened 2025-08-06
task/2.y/add-back-undo-button

Kait211:
Summary
This feature allows the user to undo the last 10 mission planning actions in JCC. It does not include or affect any Task modifications.

Implementation
Implements an UndoButton component that reverts the most recent mission planning action by dispatching JaiaActions.UNDO_LAST_ACTION. If no history remains, a notification is shown to the user. The button also ensures the waypoint panel is closed to prevent stale UI after an undo.

Testing
* Put a single way point then undo
* Put eleven waypoints and then undo


michael-jaia:
Closing PR until we merge track pod code

PR #1298, opened 2025-08-07
bug/2.y/details-panels-no-comms-health-change

KanzGiwa:
### Summary
On the JCC interface, when the simulator is stopped, the status age increments. When the status age is 30 seconds or greater, the bot colors in the left-hand panel should shift from green to gray and the Status Age row should turn red.

### Testing

1. Launch the JCC
2. Wait for the bots to connect (left-hand panel should be green)
3. Click on a botID panel in the left-hand panel
4. Expand the Quick Look tab
5. Double check to make sure Status Age is 0s
6. Stop the simulator
7. Monitor Status Age and confirm that it's incrementing 
8. Once Status Age is 30 seconds or greater, Status Age row should be red and all the Bot colors on the left-hand panel should be grey

Screenshot_7-8-2025_102657_localhost


.husky/_/pre-commit
```diff
@@ -17,10 +17,10 @@ credentials_phrases=(
 )
```
michael-jaia: Why are these changes included in the PR?

src/web/containers/BotDetails/bot-details.ts
```diff
@@ -21,12 +21,23 @@ export function getStatusAgeClassName(portalStatusAge: number) {
...
+
+    return "";
+}
+
+export function isBotDisconnected(portalStatusAge: number) {
+    const healthFailedTimeout = 30;
+    const statusAgeSeconds = convertMicrosecondsToSeconds(portalStatusAge);
+
+    if (statusAgeSeconds >= healthFailedTimeout) {
+        return "disconnected"; // gray color
```
michael-jaia: Remove comment

src/web/containers/NodeList/NodeList.tsx
```diff
@@ -62,7 +62,11 @@ export default function NodeList() {
         const selectedClass =
             selectedNode.type === nodeType && selectedNode.id === nodeID ? "selected" : "";
 
-        return `node-item ${nodeTypeClass} ${faultLevelClass} ${selectedClass}`;
+        const bot = nodeType === NodeTypes.BOT ? jaiaContext.bots.get(nodeID) : null;
```
michael-jaia: Is it possible to remove this line?

michael-jaia (changes requested):

src/web/containers/NodeList/__tests__/NodeList.test.tsx
```diff
@@ -52,10 +52,10 @@ test("Verfiy all Nodes are displayed correctly", () => {
     expect(nodes).toHaveLength(4);
     expect(nodes.map((div) => div.textContent)).toEqual(["HUB", "1", "2", "5"]);
     expect(nodes.map((div) => div.className)).toEqual([
-        "node-item hub-item faultLevel0 ",
-        "node-item bot-item faultLevel0 ",
-        "node-item bot-item faultLevel1 ",
-        "node-item bot-item faultLevel2 ",
+        "node-item hub-item faultLevel0  ",
```
JRoybot, line 56: We shouldn't need two blanks at the end of the class string.  Can we identify the cause and correct this?

JRoybot (changes requested):

src/web/containers/NodeList/__tests__/NodeList.test.tsx
KanzGiwa: There was only one blank there before, and when there was an error on CircleCI because there was an extra space at the end of each CSS class name. I updated the test to expect the extra space. Would you prefer that I do it a different way?

JRoybot: Ideally we would identify the code that is adding the trailing spaces and fix it there, quick look I didn't see it.  Check with Michael if he would like you to look for the root cause.  Probably shouldn't be any trailing spaces... the fact that it changed may help point to the source.

KanzGiwa:
Hey @JRoybot based on Micheal's ticket, the icons are supposed to turn gray and nothing modifying the Health accordion was mentioned in the ticket.

JRoybot (dismissed):

src/web/containers/NodeList/__tests__/NodeList.test.tsx
michael-jaia: The empty spaces are added from the selected and disconnected class when the ternary statement returns false, I don't see a problem with it

PR #1300, opened 2025-08-08
task/2.y/add-back-measure-tool

michael-jaia:
### Summary
This pull request adds back the measure tool. Instead of seeing the distance in the side bar, a panel now displays the measured distance.

image

### Implementation
The measure functionality uses the OpenLayers Draw Interaction. The code is based on the example (https://openlayers.org/en/latest/examples/measure.html). The distance updates are handled by OpenLayers to prevent the React code from updating excessively (you can see how quickly the updates occur when using a mouse to control the measure tool).

### Testing
* Measure distances shorter than 1 km
* Measure distances longer than 1 km
* Close the measure tool by clicking the measure button
* Close the measure tool by clicking any other panel, command, or map mode button
* Open the panel and expect the distance to read 0 meters


JRoybot (dismissed):
Reviewed all code, built and tested the branch.  Seems to work well.  One question on desired behavior.  The tool behaves like a polygon draw tool.  Subsequent clicks on the map create a series of dashed lines, similar to a waypoint route.  The displayed distance appears to be the sum of all the dashed lines displayed.  Clicking on the last endpoint changes the lines from dashed to solid.  Clicking the endpoint a 3rd time erases the lines and resets the distance.

I was not expecting all of that behavior.  Is that intended?

michael-jaia:
> Reviewed all code, built and tested the branch. Seems to work well. One question on desired behavior. The tool behaves like a polygon draw tool. Subsequent clicks on the map create a series of dashed lines, similar to a waypoint route. The displayed distance appears to be the sum of all the dashed lines displayed. Clicking on the last endpoint changes the lines from dashed to solid. Clicking the endpoint a 3rd time erases the lines and resets the distance.
> 
> I was not expecting all of that behavior. Is that intended?

Yes, that is the expected behavior. It maintains the functionality of the current line tool + the ability to double click to keep the line visible on the map until the next click.

src/web/components/Map/Map.tsx
```diff
@@ -36,6 +36,15 @@ export default function Map() {
      * @returns {void}
      */
     const handleMapClick = (event: MapBrowserEvent<PointerEvent>) => {
+        switch (jaiaGlobal.getMapMode()) {
+            case MapModes.RALLY:
+                handleAddRallyPoint(event.coordinate);
+                return;
+            case MapModes.MEASURE:
+                // Measurement clicks handled by measure layer
```
ferrom-jaia: Should we indicate where this code lives?

ferrom-jaia: ex the path?

src/web/style/openlayers/colors.ts
```diff
@@ -7,4 +7,5 @@ export enum OpenLayersColors {
     TARGET = "chartreuse",
     INNER_COMMS = "green",
     OUTER_COMMS = "rgba(192,0,0)",
+    MEASURE_LINE = "black",
```
ferrom-jaia: This works for now, but not for every map overlay

ferrom-jaia (commented):
Looks great! I left a couple of comments/questions below.

src/web/style/openlayers/colors.ts
michael-jaia: Good call, I will pick a color that works better with the satellite tiles

JRoybot (approved):

PR #1302, opened 2025-08-12
task/2.y/dev-friendly-jcc/save-mission-panel-2/jaia-1865

JRoybot:
Create dialog for managing mission sets in local storage.  A single dialog was created that allows the user to load and save mission sets from/to local storage as well as to delete mission sets from local storage.  The primary dialog includes buttons that may open secondary dialog windows to confirm actions or alert of errors.

Functionality to manage the local storage has already been tested as part of PR #1249 

Testing was done using the simulator and a browser.


src/web/components/LoadSaveMissionSet/DeleteMissionSetButton/DeleteMissionSetButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+        if (!listSavedMissionSets().includes(props.saveName)) return DisabledCodes.FILE_NOT_FOUND;
+        return DisabledCodes.CONFIRM;
+    };
+
+    /**
+     * Displays dialog before loading the mission set
+     *
+     * @returns {void}
+     */
+    const onButtonClick = async () => {
```
michael-jaia: I don't think async does anything here

src/web/components/LoadSaveMissionSet/DeleteMissionSetButton/DeleteMissionSetButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+    };
+
+    /**
+     * Checks the mission set and applies the appropriate disable code
+     *
+     * @returns {DisabledCodes} The applicable disabled code based on the mission set conditions
+     */
+    const getDisabledCode = () => {
+        if (!listSavedMissionSets().includes(props.saveName)) return DisabledCodes.FILE_NOT_FOUND;
+        return DisabledCodes.CONFIRM;
```
michael-jaia: Should this be DisabledCodes.NONE?

src/web/components/LoadSaveMissionSet/DeleteMissionSetButton/DeleteMissionSetButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { DisabledCodes } from "./delete-messages";
+
+import { Icon } from "@mdi/react";
+import { Button } from "@mui/material";
+import { mdiDelete } from "@mdi/js";
+
+import { DialogActions } from "../../../types/context-types";
+
+interface Props {
+    saveName: string;
```
michael-jaia: What do you think about missionSetName?

src/web/components/LoadSaveMissionSet/DeleteMissionSetButton/DeleteMissionSetButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * Closes the dialog and then acts based on the button clicked
+     *
+     * @param {DialogActions} dialogAction Indicates which button was clicked
+     * @returns {void}
+     *
+     */
+    const onDialogClose = (dialogAction: DialogActions) => {
+        setIsDialogVisible(false);
```
michael-jaia: clear save name

src/web/data/mission_set/mission-set.ts
```diff
@@ -140,7 +140,7 @@ class MissionSet {
                 this.addMission(mission);
             });
         }
-
+        this.missionIDInEditMode = UNASSIGNED_ID;
```
michael-jaia, line 143: Make default value ""

src/web/components/LoadSaveMissionSet/DeleteMissionSetButton/DeleteMissionSetButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { DialogActions } from "../../../types/context-types";
+
+interface Props {
+    saveName: string;
+}
+/**
+ * Produces the delete mission set button.
+ * It manages the alert/confirm dialog that appears when clicking on the button.
+ */
+export default function LoadMissionSetButton(props: Props) {
```
JRoybot: Fixing this function name, no idea how this ever worked

src/web/components/LoadSaveMissionSet/DeleteMissionSetButton/DeleteMissionSetButton.tsx
```diff
@@ -61,6 +61,7 @@ export default function LoadMissionSetButton(props: Props) {
         setIsDialogVisible(false);
         if (dialogAction === DialogActions.CONFIRMED) {
             missionSet.deleteFromLocalStorage(props.saveName);
+            props.saveName = "";
```
JRoybot: This does not have the desired effect, will probably be removed.

src/web/components/LoadSaveMissionSet/LoadSaveMissionSetDialog.tsx
```diff
@@ -22,27 +24,9 @@ export function LoadSaveMissionSetDialog(props: DialogProps) {
...
-                    if (e.key === "Enter") {
-                    }
-                }}
-            ></input>
-        </div>
-    );
-
-    // Mission rows
+    const clearSaveName = () => {
+        setSaveName("");
```
JRoybot: I don't think clearing the local state variable without calling mission-setName has the desired effect.  When a mission is saved to local storage the save name is stored in the mission-set.  With this implementation the save name the dialogs is using gets cleared but if the mission panel is closed and then re-opened the save name is back to what it is stored in `missionSet`  

As I said in a chat last week.  If a mission set has a saved name I think it would be useful to keep that as part of it's data so if the user goes to save it again, even if he deleted the previous version of it from local storage, the system should pre-fill the dialog with the previously used name.

I think it might work better if the delete mission set, just reset the saveName to what is returned by missionSet.getName() instead of "".  That way if it was the same name as the current misison set it will not be cleared but if the user deleted a different mission set it would get reset to "", which is now the default, or the name that was last used to save it.  

I will code this up and try it.

src/web/data/mission_set/mission-set.ts
```diff
@@ -162,6 +162,7 @@ class MissionSet {
         delete allMissionSets[name];
 
         localStorage.setItem("missionSets", JSON.stringify(allMissionSets));
+        this.setName("");
```
JRoybot: I do not think we want this

1. User saves current missionSet, which will save the name so next time he goes to save it, the name is remembered.
2. User deletes a missionSet from local storage -- Even if the name of the missionSet that was deleted is different than the current on the system forgets the saved name
3. User goes to save the mission set again and had to retype the name.  This could be annoying if someone is planning missions ahead of test or other scenarios.

michael-jaia:
Note before merging: increase Stored Mission Sets box size when no mission sets are saved

PR #1303, opened 2025-08-12
bug/iso-upgrade-to-2.y-error-fix

marshalllele:
When adding the fleet configuration file to the 2.y ISO, the following error message is outputted
```
jaia@pequod:~$ jaia admin fleet update_iso private_jaia/fleet_config/fleet4.cfg Downloads/jaiabot_updates_noble_2.0.0~beta3_arm64.iso
genisoimage: Error: '(null)' and './rr_moved' have the same Rock Ridge name 'rr_moved'.
Unable to sort directory
```

@tsaubergine I believe that adding the `-m rr_moved` parameter to the `genisoimage` call in jaiabot/scripts/fleet/jaia-fleet-config-update-iso.sh forces the call to ignore any pre-existing file/directory named "rr_moved." Do you know if this is the best way to go about this? Adding that parameter makes the process work seamlessly with the instructions.


tsaubergine:
I can't replicate this - what version of genisoimage are you using?

`genisoimage -version`

I get `genisoimage 1.1.11 (Linux)` (Ubuntu 24.04 amd64)

marshalllele:
`1.1.11` here as well. We can reproduce consistently on Pequod in the office. It's also running 24.04. I'm not totally sure what's causing it on our end, but this change fixed the issue.

tsaubergine (approved):

tsaubergine:
I certainly don't see this causing any issues as rr_moved is written by genisoimage.

PR #1304, opened 2025-08-12
task/2.y/add-back-drift-icons

michael-jaia:
### Summary
This pull request adds back the drift layer to display drift vectors on the map. The structure follows the pattern of the dive layer. There are six different drift icons; the more dashes on the drift icon, the faster the drift speed. 

| Drift Speed (m/s) | Dashes |
| ---                        | ---        |
| 0-0.5                    | 1           | 
| 0.5-1.0                 | 2           | 
| 1.0-1.5                 | 3           | 
| 1.5-2.0                 | 4           | 
| 2.0-2.5                 | 5           | 
| 2.5+                     | 6           | 

image

### Testing
* Drift icon renders after drift task
* Click drift icon and see task packet panel with correct data and flashing icon
* Close task packet panel and icon stops flashing
* Select another task packet (dive or drift) and see the panel open and the clicked icon flash


KanzGiwa:
The testing instructions are confusing and difficult to understand how to start the drift tasks.

src/web/openlayers/layers/zindex.ts
```diff
@@ -7,4 +7,5 @@ layersZIndexes.set(LayerTitles.HUB_LAYER, 5);
 layersZIndexes.set(LayerTitles.MISSION_LAYER, 4);
 layersZIndexes.set(LayerTitles.RALLY_LAYER, 3);
 layersZIndexes.set(LayerTitles.DIVE_LAYER, 2);
+layersZIndexes.set(LayerTitles.DRIFT_LAYER, 2);
```
JRoybot, line 11: Is it OK to have both dive and drift layers with the same Z index?

JRoybot (dismissed):
Built and tested branch, reviewed code all looks good.  Approved but left one question

src/web/openlayers/layers/zindex.ts
michael-jaia: I think so since they are "equals". I did not see an issue with the task packets rendered in testing. If we do see a weird overlapping behavior (which I do not expect), we can make the change.

PR #1306, opened 2025-08-13
task/log-file-documentation

aofolz:
this is a directory to help users find the data they need in the jdv.


michaelrockJAIA:
we should add a section explaining that this file was generated and kept up to date since version X.XX.XX. We have changed the naming scheme in the past and when we start to have multipe sorces of the same field we may change the names to include the sensor used to collect the data

ferrom-jaia (approved):
Looks great!

PR #1310, opened 2025-08-14
feature/svp-pipeline

aofolz:
Scripts that align h5 data, filter it for noise, and calculate SVP using gsw toolbox.


ferrom-jaia (approved):
Nice readme. Merging in as it is contained in its own directory. Will be utilized in future work.

PR #1313, opened 2025-08-19
task/2.y/add-back-hub-connection-error

michael-jaia:
### Summary
This pull request adds back the Hub <-> client connection error message. The warning message becomes visible when we cannot hit the status endpoint.

image

### Testing
1. Start/stop the web server and see the connection warning appear/disappear
2. Start/stop the simulator and see the connection warning appear/disappear


src/web/jcc/App.less
```diff
@@ -28,3 +28,21 @@ body {
     height: 100%;
     width: 100%;
 }
+
+#connection-warning {
```
JRoybot, line 32: I find the text in the warning dialog for to be too small relative to the rest of the text in JCC.  If I set the zoom level so that the rest of JCC looks good the text is tiny, if I zoom in to make the text readable the buttons and labels in other parts of the app look too big

JRoybot (changes requested):
Built and tested branch.  Left comment regarding dialog text size.

JRoybot (approved):
Re Ran the branch and set zoom level to 100% and text looked good.

PR #1316, opened 2025-08-22
bug/echo-driver-needs–to-point-at-ama5-in-24.04

marshalllele:
We were seeing the warning 

`HEALTH__DEGRADED
WARNING__NOT_RESPONDING__JAIABOT_ECHO_DRIVER` 

on JaiaBot-PAMs that are running 2.0.0 and 24.04. We ran into this same problem when trying to access the RPiCam on `/dev/ttyAMA3`. The fix was updating `jaiabot/src/python/echo/echo.py` so that the driver would look at `/dev/ttyAMA5` instead. 

To prevent this issue from popping up in the future, this PR includes a UDEV rule which creates a symlink to whichever UART line the pam stack is hooked up to. It does this by looking for the PAM stack's specific iomem_base attribute.


ferrom-jaia (approved):
Looks good!

PR #1319, opened 2025-09-02
feature/offline-maps

edsanville:
Offline maps, including:

* NOAA ENC, Open Street Maps, and satellite imagery (downloaded from the web and then uploaded to the hub as tiles), 
* GeoTIFF uploads to the hub (which are converted to tiles for the hub's tile server).
* Ability to delete offline map layers from the hub
* JCC shows disk usage for each layer, and warns about estimated disk space taken by new layers


ferrom-jaia (approved):
Looks great, tested in sim. Functionality meets expectations.

PR #1320, opened 2025-09-04
task/2.y/developer-friendly-jcc/add-back-csv-kmz-download-buttons/JAIA-2034

JRoybot:
Added download CSV and KMZ buttons to the `DataOffloadPanel`

Most of the functional logic was in place from 2.y baseline.  Functions from `src/web/shared` copied to `utils` folder and updated as needed.

Tested by running missions in JCC using the simulator and exporting the task packet data as both CSV and KMZ files.  CSV files checked in Excel.  KMZ files checked by unzipping them and examining contents in an editor and uploading them to Goggle Earth and verifying the display.

image


src/web/components/DataOffloadPanel/DataOffloadPanel.tsx
```diff
@@ -1,16 +1,63 @@
 import DataOffloadQueue from "../DataOffloadQueue/DataOffloadQueue";
 import ExpandMoreIcon from "@mui/icons-material/ExpandMore";
 import { Accordion, AccordionDetails, AccordionSummary } from "@mui/material";
+import { Button } from "@mui/material";
+
+// Utilities
```
michael-jaia, line 5: Jeff we want to stay away from the shared folder. Anything you use from the shared folder should be re-written and placed in the most logical location given the new file structure

src/web/components/DataOffloadPanel/DataOffloadPanel.tsx
```diff
@@ -1,16 +1,63 @@
...
+
+        kmlDocument.setTaskPackets(taskPackets.getTaskPackets());
+
+        let fileDate = new Date();
+        // Use the date of the first task packet, if present
+        if (taskPackets.getTaskPackets()[0]?.start_time !== undefined) {
+            fileDate = new Date(taskPackets.getTaskPackets()[0].start_time / 1e3);
+        }
+
+        const dateString = fileDate.toISOString();
```
michael-jaia: I don't think we need the `dateString` variable since it is only used once

src/web/components/DataOffloadPanel/DataOffloadPanel.tsx
```diff
@@ -1,16 +1,63 @@
...
+     */
+    const handleDownloadCSV = async (event: React.MouseEvent<HTMLButtonElement>) => {
+        const csvFilename = getCSVFilename(taskPackets.getTaskPackets());
+        downloadToFile(await getCSV(taskPackets.getTaskPackets()), "text/csv", csvFilename);
+    };
+
     return (
         <div className="jaia-panel data-offload-panel">
             <div className="jaia-panel-title">Data Offload</div>
+            <div className="button-row">
```
michael-jaia, line 40: We should have the function names follow the same pattern either handleDownloadTYPE or handleClickedDownloadTYPE

src/web/utils/download.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @param {string} data Contents written to file
+ * @param {string} mimeType Informs the browser of the type of data being sent
+ * @param {string} fileName Name given to the downloadable file
+ * @returns {void}
+ */
+export function downloadToFile(data: string, mimeType: string, fileName: string) {
+    const blob = new Blob([data], { type: mimeType });
+
+    var link = window.document.createElement("a");
```
michael-jaia: No var

src/web/utils/download.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Starts a browser download of a file with binary data contents
+ *
+ * @param {string} name Name given to the downloadable file
+ * @param {BlobPart} data Contents written to file
+ * @returns {void}
+ */
+export function downloadBlobToFile(name: string, data: BlobPart) {
+    let a = document.createElement("a");
```
michael-jaia: Let's name to `link` for consistency. Plus how the "click" occurs. I don't think one is better than the other, but they should be consistent

michael-jaia (changes requested):

src/web/components/DataOffloadPanel/DataOffloadPanel.tsx
JRoybot: OK, was trying to reuse what I could

PR #1322, opened 2025-09-09
task/2.y/add-back-depth-contour-map

michael-jaia:
### Summary
This pull request renders the contour map. Ed designed the contour processing to take place on the backend and made the data available through the contours endpoint. We poll this endpoint every 10 seconds (in unison with the task packet endpoint poll) to gain access to the GeoJSON data on the client side. We then use OpenLayers to convert the data to features and render the contour map.

### Testing
image


JRoybot (dismissed):
Did a static review of code, all looks good.  Built and ran branch all worked.  Would like to see warning suppressed if possible.

michael-jaia:
Contour map after implementing the changes to prevent unnecessary warnings from filling the console.

image

PR #1325, opened 2025-09-10
task/2.y/refactor-move-waypoint-structure/JAIA-2076/

JRoybot:
Moved the `isMoveable` property from `waypoint` to the `selectedWaypoint` in  `jaiaGlobal`  to be more consistent with our model of tracking UI state in `jaiaGlobal`.  

tested in browser

### DIFF
--- a/src/web/context/JaiaContext.tsx
+++ b/src/web/context/JaiaContext.tsx
@@ -31,7 +31,6 @@ import {
     TaskType,
 } from "../types/protobuf-types";
 import { DATA_MODEL_POLL_TIME, UNASSIGNED_ID } from "../utils/constants";
-import { compareWaypoints } from "../utils/comparisons";
 import { MapModes } from "../types/openlayers-types";
 import { MapFeatureTypes } from "../types/openlayers-types";
 import {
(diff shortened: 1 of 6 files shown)


src/web/context/JaiaContext.tsx
```diff
@@ -876,8 +879,9 @@ function handleClickedEditMission(mutableState: JaiaContextType, missionID: numb
  * @returns {JaiaContextType} Updated mutable state object
  */
 function handleClickedTapToMove(mutableState: JaiaContextType) {
-    const waypoint = getWaypoint();
-    waypoint.setIsMovable(!waypoint.getIsMovable());
+    mutableState.selectedWaypoint.isMoveable = !mutableState.selectedWaypoint.isMoveable;
+    jaiaGlobal.setSelectedWaypoint(mutableState.selectedWaypoint);
+    mutableState.selectedWaypoint = jaiaGlobal.getSelectedWaypoint();
```
michael-jaia: I think we could remove this line

michael-jaia (commented):
Jeff this is a much cleaner implementation. As we texted about, I think we should not change the tap to move toggle state when a waypoint is clicked.

src/web/context/JaiaContext.tsx
JRoybot: put it there just for consistency with some other code, agree it is not needed

PR #1328, opened 2025-09-16
Task/2.y/add checksum to major upgrade and fix no disk space bug

tsaubergine:
## Work performed

1. Fixes bug where major_upgrade ran out of space on former writable "overlay" partition. This is due likely to undefined behavior (see https://docs.kernel.org/filesystems/overlayfs.html) on the writable "overlay" partition when the overlayfs was still mounted. Even though files are deleted, and the fs is `sync`'d, the disk space still wasn't available (some of the time - this seemed to be unpredictable. I guess that's undefined behavior for you!)

The new solution is cleaner and less clever but adds a reboot (not a big deal on a relatively lengthy process already):
- Install do-major-upgrade.sh as new systemd service on rootfs (underlay) partition.
- Disable overlayfs
- Reboot into formerly read-only rootfs (underlay) partition.
- (Automatically, using systemd from the service installed in the first step) run do-major-upgrade.sh again which finishes the installation now that the overlayfs isn't mounted at all. This makes the second pass closer to the use of this script when upgrading an AWS AMI, which is nice.

2. Switched from `wget` to using `get_url` Ansible module with a checksum (SHA256) calculated (and added to the existing `version.txt`) when the major_upgrade ISO is created for both the rootfs and bootfs tarball. This should hopefully catch corrupted downloads as seen on one bot on a recent fleet upgrade. Three retries are made before failing completely.

*Same changes made to `1.y`* in https://github.com/jaiarobotics/jaiabot/pull/1329

## Testing

Upgrading 3-bot 1-hub VirtualBox fleet from 1.20 to this test/2.y branch by 

0. Mimic full overlay fs by writing dummy file to disk:
```
cd /
sudo fallocate -l 6G largefile.img
```
1. Upgrading hub and 1 bot.
2. Using upgraded hub to upgrade 2 other bots.

**Needs testing on Raspi fleet**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -178,6 +178,7 @@ filter-template-debian-builds: &filter-template-debian-builds
     branches:
       only:
         - "2.y"
+        - "task/2.y/add-checksum-to-major-upgrade"
 
 # which branches to run the normal source build
 filter-template-normal-builds: &filter-template-normal-builds
@@ -185,6 +186,7 @@ filter-template-normal-builds: &filter-template-normal-builds
     branches:
       ignore:
         - "2.y"
+        - "task/2.y/add-checksum-to-major-upgrade"
 
 # which branches to run the pi-zero filesystem builds
 filter-template-pi0-rootfs-builds: &filter-template-pi0-rootfs-builds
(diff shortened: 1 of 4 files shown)


ferrom-jaia (approved):
Looks great. Tested with a fleet of 6 bots in various states (1.y vs 2.y).

Upgrade worked as expected.

PR #1329, opened 2025-09-16
Task/1.y/add checksum to major upgrade and fix no disk space bug

tsaubergine:
1. Identical to https://github.com/jaiarobotics/jaiabot/pull/1328 but for `1.y`. Needs to be in `1.y` so `1.y` hubs can correctly and reliably upgrade themselves and `1.y` bots.


ferrom-jaia (approved):
Looks great. Tested with a fleet of 6 bots in various states (1.y vs 2.y).

Upgrade worked as expected.

PR #1330, opened 2025-09-17
feature/jdv-path-search

edsanville:
* Type into search bar to find data series in your logs
* Search is based on the human-readable name, human-readable description, and path of the series within the HDF5 file
* User can switch between search and browse
* Search defaults to the 20 most recently plotted series, in order of recency
* Browse is the old GUI, where you drill down through the HDF heirarchy


.husky/_/pre-commit
```diff
@@ -1,9 +1,2 @@
...
-
-# clang-format=hooks
-scripts/git-hooks/clang-format-hooks/git-pre-commit-format
-
-# prettier
-scripts/git-hooks/prettier/write.sh
-
-exit 0
+#!/usr/bin/env sh
+. "$(dirname "$0")/h"
```
JRoybot: Wasn't this already fixed on a different branch?

src/web/jdv/client/src/PathSelector.tsx
```diff
@@ -25,54 +79,173 @@ export default class PathSelector extends React.Component {
...
+
+        let results: {[path: string]: number} = {};
+
+        for (const query_part of query.toLowerCase().split(" ")) {
+            if (query_part.length < 2) {
+                continue;
+            }
+
+            for (const keyword_entry of this.state.series_descriptor_keyword_index) {
+                // TODO: Use a binary search here since the index is sorted
```
JRoybot: Is this still TODO?

src/web/jdv/client/src/PathSelector.tsx
```diff
@@ -25,54 +79,173 @@ export default class PathSelector extends React.Component {
...
+                    path={mdiMagnify}
+                    size={1}
+                    style={{ verticalAlign: "middle" }}
+                ></Icon>
+                <input
+                    className="padded"
+                    type="text"
+                    placeholder="Search"
+                    value={this.state.search_text}
+                    onChange={(e) => {
```
JRoybot: suggest moving functional code to a const function to keep the html/React code clean

src/web/jdv/client/src/PathSelector.tsx
```diff
@@ -25,54 +79,173 @@ export default class PathSelector extends React.Component {
...
+                        this.search(e.target.value)
+                    }} />
+                <button
+                    className="padded button"
+                    title="Clear"
+                    onClick={(e) => {
```
JRoybot: same

JRoybot (commented):
I did a review of the code and left a few picky comments, regarding logic in the html/React code, left a few comments but applies generally.  As part of the JCC Refactor, Twomey has stressed he does not like to see any logic embedded in the render code, prefers functions defined for the logic.  I was not able to test the code so am only providing comments at this time.  Will try to exercise the code and provide a full review

JRoybot (dismissed):
Tested with sim logs and found it very easy to find common plots.  Great Job on this!

.husky/_/pre-commit
edsanville: Yes, I'm confused about this file... I can remove it from the commit.

src/web/jdv/client/src/PathSelector.tsx
edsanville: Good point, it's already done!

src/web/jdv/client/src/PathSelector.tsx
edsanville: I moved the event handler out into a local closure, instead of a JSX-inline lambda function.

JRoybot (dismissed):

src/web/jdv/server/path_descriptors.py
```diff
@@ -0,0 +1,143 @@
...
+
+    if len(message_type_components) < 1:
+        l.warning(f'Invalid path: {path}')
+        return ''
+
+    components[0] = components[0].split('.')[-1]
+    return '/'.join(components)
+
+
+def get_by_path(path: str) -> PathDescriptor | None:
```
marshalllele, line 313: Is there anyway to get this search function to find sub-strings in a field name? For example, searching "current" would yield the "/jaiabot::arduino_to_pi/jaiabot.protobuf.ArduinoResponse/vcccurrent" in the results.

marshalllele (changes requested):

ferrom-jaia (approved):
Looks good! Tested in sim.

PR #1332, opened 2025-09-18
bug/sim-bot-jump-last-wpt

AdeX11:
## Summary
In simulation, when the bots start their first mission, they jump to the last waypoint instead of the first.

## Solution
Instead of setting the lat_origin and lon_origin to point towards the last goal in the mission plan, I set it to point towards the first goal(index 0).

## Testing
Used the simulator and sent the bots on a mission. I confirmed that they started at the first waypoint instead of the last.


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #1333, opened 2025-09-22
task/2.y/add-back-undo-button-JR/JAIA-2050

JRoybot:
Implemented an Undo and Redo capability for JCC.

Initial plan was to just provide buffering of mission data, however during implementation and testing it became clear it would be better to provide a complete solution to capture the entire state of the application so the scope was expanded a bit.

Features of history buffer implementation

- Utilizes a circular buffer with a size that can be set in the constructor
  - using MAX_HISTORY in constants.ts
- Generic class so can buffer any type
- Includes descriptions of what can be undone/redone
- Includes helper functions to support rendering the buttons
    - canUndo, canRedo, peekUndoDescription, peekRedoDescription

To facilitate implementation and make the code easier to maintain parts of JaiaContext were refactored, refactoring included

- Unified parameter profile of all action handlers so they could be easily mapped
- Replaced long Switch in jaiaReducer with mapped configuration for each action
   - map includes handler and flag if action is undoable
- Once functionality was tested JaiaContext was broken up to more manageable set of files
   - All types moved to `src/web/types/context-types.ts`
   - actionConfig moved to it's own file
   - action handlers grouped and moved to individual files under `src/web/context/handlers`

New Features Added

- Undo and Redo buttons with helper text describing the action 
- Handlers for the Undo and Redo actions
  - unique actions created instead of using CLICKED_BUTTON to keep undo/redo code isolated
- Defined type for snapshots of application state including items from JaiaContextType and parts of the data model
  - Restores all aspects of application state insuring consistent state after undo/redo
- uses cloneDeep to keep snapshots in history buffer isolated from future changes
- integrated pushing of states into jaiaReducer
- undo/redo supported for the following actions
   - ADD_MISSION
   -DELETE_MISSION
  -DUPLICATE_MISSION
  -DELETE_ALL_MISSIONS,
  -ASSIGN_MISSION
  -AUTO_ASSIGN_MISSIONS
  -CHANGE_MISSION_SPEEDS,
  -LOAD_MISSION_SET
  -ADD_WAYPOINT
  -DELETE_WAYPOINT
  -MOVE_WAYPOINT
  -SELECT_TASK
  -CHANGE_TASK_PARAMETER
  -TOGGLE_BOTTOM_DIVE

Other changes required to support clean undo/redo
- Prevented Map from generating invalid ADD_WAYPOINT events that cluttered the history buffer

Testing

- Unit test to validate historyBuffer class behavior
- Unit test to insure all JaiaActions are mapped to handlers
- All existing tests pass
- Extensive testing and debugging in Browser

Undo-Redo-Flow png


michael-jaia:
Hey Jeff, I ran into this problem that caused the JCC to crash:

I created a single mission, changed the tasks, did some undo and redos, and then undid everything until no waypoints remained. When I clicked on the map to create a new mission, the error occured.
 
image

michael-jaia:
https://github.com/user-attachments/assets/5abe819b-7c9a-4c6c-9d59-5987426c803c

michael-jaia:
Jeff I also think it would be good to have a diagram to go with the PR for review and future development purposes.

src/web/utils/history-buffer.ts
```diff
@@ -0,0 +1,146 @@
...
+        this.size = 1;
+    }
+
+    /**
+     * Clear the entire history
+     * @param {T} initialValue Optional new initial value, will use original if not provided
+     * @returns {T} current value of history buffer (initialValue)
+     */
+    reset(initialValue?: T) {
+        this.buffer.fill(undefined);
```
michael-jaia: Jeff what was the thinking behind filling the array with undefined versus creating a new array ?

src/web/utils/history-buffer.ts
```diff
@@ -0,0 +1,146 @@
...
+     * @param {T} value New state to push onto history buffer
+     * @param {string} description Description of action creating new state
+     *
+     * @notes Provide value should be cloned if mutable to prevent
+     *        corruption of history
+     */
+    push(value: T, description: string) {
+        const entry = { value, description };
+
+        // Trim redo path if we've undone some entries
```
michael-jaia, line 53: I'm a little confused by this block of code

src/web/utils/history-buffer.ts
```diff
@@ -0,0 +1,146 @@
...
+        this.head = 0;
+        this.size = 1;
+        return this.getPresent();
+    }
+    /**
+     * Push a new value onto the history buffer
+     * @param {T} value New state to push onto history buffer
+     * @param {string} description Description of action creating new state
+     *
+     * @notes Provide value should be cloned if mutable to prevent
```
michael-jaia: The comments are helpful about cloning to prevent corruption of history...did you consider just doing the cloning in the function

src/web/utils/history-buffer.ts
```diff
@@ -0,0 +1,146 @@
...
+     */
+    getPresent() {
+        return this.buffer[this.index].value;
+    }
+
+    /**
+     * Provides description of the last action for undo button tooltip
+     * @returns {string} Description of action to be undone
+     */
+    peekUndoDescription() {
```
michael-jaia, line 122: Should we return "" instead of undefined?

src/web/utils/history-buffer.ts
```diff
@@ -0,0 +1,146 @@
...
+     */
+    canUndo() {
+        return this.index !== this.head;
+    }
+
+    /**
+     * Checks if there is anything to redo
+     * @returns {boolean} True if redo can be performed
+     */
+    canRedo() {
```
michael-jaia, line 151: Could this just be this.index !== tail?

src/web/utils/history-buffer.ts
JRoybot: No real reason, just didn't consider creating a new array

src/web/utils/history-buffer.ts
JRoybot: Yes this was a tricky but very important aspect to get right.  When we go back in history using undo we leave some future states in the buffer that we can use redo to get back to.  If a new state is pushed we need to trim the history so we do not redo into a bad state. This is the case if the current index is not the tail of the buffer.  We trim the size to eliminate the bad future states from being used.  The buffer is never empty, so size is always between 1 and capacity.

this seemed to be the simplest way to get that.
`(this.index - this.head + 1 + this.capacity) % this.capacity || this.capacity;`

`this.index - this.head + 1 + this.capacity) % this.capacity` results in a value 0 .. capacity-1
`|| this capacity` handles the case when it evaluates to 0, which would be an empty buffer, instead it should be capacity for a full buffer.

I am open to other formulas, but this works.

src/web/utils/history-buffer.ts
JRoybot: yes that might be a little cleaner

src/web/utils/history-buffer.ts
JRoybot: hmmmm, I thought there was an edge case that needed it but looking at it, that case may have been eliminated by other changes... let me test and get back to you.

(at one point I initialized the buffer to be empty but that created lots of problems integrating it into JaiaContext so made the decision to add an initial value and never let the buffer empty ... this may be left over from an earlier version

src/web/utils/history-buffer.ts
JRoybot: I did consider that.  I decided to leave it to the consumer code of the history buffer for 2 reasons.

1.  Because this is generic, it could be instantiated with scalars, which do not need to be cloned (see unit test)
2.  I wanted the cloning to be explicit in the consumer code so it was clear how and what is being cloned.

src/web/utils/history-buffer.ts
michael-jaia: Ok that is good to know. I'm wondering if `this.size = index + 1` would work?

JRoybot: no because head is not always 0, once the buffer is full the head moves forward with each new push

michael-jaia (dismissed):
Awesome work

PR #1347, opened 2025-10-06
task/update-gps-data-rate-from-5-MHz-to-1-MHz

marshalllele:
Updates the gps data rate from 5 MHz to 1 MHz. This reduces the amount of bad data the chip sends over the line.


ferrom-jaia (approved):
This was tested on multiple fleets. Looks good!

PR #1225, opened 2025-06-09
task/add-back-take-control

michael-jaia:
### Summary
Integrates take control functionality with the individual activate button. The necessary data for making take control decisions existed in the `jaiaAPI` singleton. A utility function `isControllingClient` was added to `utils/commands.ts` to help know when the take control dialog needs to be displayed prior to sending a command. This will provide us the structure to integrating take control with the other command buttons.

### Testing
1. Opened two instances of the JCC
2. Activated Bot 1 in tab 1
    * Saw expected confirmation dialog
4. Switched to tab 2
5. Activated Bot 2 in tab 2
    * Saw expected take control dialog, then activate dialog
6. Restart the simulation (not the server)
7. Activated Bot 1 in tab 1
    * Saw expected take control dialog, then activate dialog

PR #1348, opened 2025-10-06
task/2.y/finish-take-control-rollout/JAIA-2112

michael-jaia:
### Summary
In PR https://github.com/jaiarobotics/jaiabot/pull/1225 we created the take control structure. This pull request applies that structure to the buttons that send commands to the Bots.

The buttons that require take control include:
* Activate/Activate All
* Start/Start All
* Stop/Stop All
* Data Offload/Data Offload All
* Next Task
* Go To Rally
* Enter RC Mode

### Testing
Opened two instances of the JCC and switched control between the clients by sending commands to the Bots.

image


src/web/components/__buttons__/DataOfffloadAllButton/DataOffloadAllButton.tsx
```diff
@@ -63,9 +66,15 @@ export default function DataOffloadAllButton(props: Props) {
      *
      * @returns {void}
      */
-    const handleClick = () => {
-        setIsDialogVisible(true);
-        groupBotsByReadyState();
+    const handleClick = async () => {
+        const hasControl = await isControllingClient();
+
+        if (!hasControl) {
```
JRoybot, line 71: This logic is repeated nearly identically in the other buttons as well.  Would be nice if we could extract what is common into a utility to reduce code and simplify maintenance

src/web/components/__buttons__/DataOfffloadAllButton/DataOffloadAllButton.tsx
```diff
@@ -88,6 +97,22 @@ export default function DataOffloadAllButton(props: Props) {
...
     };
 
+    /**
+     * Closes the take control dialog. If control is taken, the command
+     * dialog will appear.
+     *
+     * @param {DialogActions} dialogAction The action taken by the operator
+     * @returns {void}
+     */
+    const onTakeControlClose = (dialogAction: DialogActions) => {
```
JRoybot: Similarly, this logic is fairly common among the buttons, could extract what is common.

JRoybot (changes requested):
Reviewed all code, built and tested branch. 

Left a few comments regarding common code.  Would like to see the common code extracted to a utility so it isn't repeated and will be simpler to maintain.

Exercising the branch in the browser, it all seemed to work.  However I do think there was 2.y value in the warning you are not in control before trying to issue a command.  Currently you do not know you are not in control until you try to do something.

src/web/components/__buttons__/DataOfffloadAllButton/DataOffloadAllButton.tsx
michael-jaia: Due to slight differences in the handleClick functions between buttons, I am going to keep this function in each button file for now (i.e. some buttons need to dispatch an action, some do not)

JRoybot (approved):
Changes look great.  Nice job!

PR #1351, opened 2025-10-07
bug/jdv-plots-jitter

marshalllele:
Removes the jitter when scrubbing through certain datasets in JDV. Also aligns the columns and headers of the data table at the bottom of the page so that the data is easier to read while scrubbing through.


ferrom-jaia (approved):
Looks good tested in sim.

PR #1352, opened 2025-10-08
bug/fleet-config-message-missing-comms-field-in-1.y

marshalllele:
Migrating new FleetConfig message from 2.y into 1.y so that a major upgrade works without modification to the fleet config file.


ferrom-jaia (approved):
Looks good. Needed to ensure upgrades proceed going from 1.y to 2.y.

PR #1353, opened 2025-10-08
bug/jdv-log-time-range-reset

edsanville:
When loading another log, the system doesn't reset the tMin and tMax, so the timebar at the bottom had the wrong range.  This fixes that.


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #1355, opened 2025-10-09
bug/jdv-remove-old-bots-on-map

edsanville:
Reset timestamp to null when loading new logs, and update the bot layer


ferrom-jaia (approved):
Looks good! Tested in sim.

PR #1359, opened 2025-10-14
task/2.y/add-back-rc-dive/JAIA-2096

JRoybot:
Added Dive mode functionality to the Remote Control Panel.

Tested in browser with simulator, verified parameters are passed correctly by inspecting at breakpoint and verified bot did dive.

Layout of controls is an initial concept, we can tweak it if needed.


src/web/components/RemoteControlPanel/Dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,30 @@
+import { createTheme, MenuItem, Select, SelectChangeEvent, ThemeProvider } from "@mui/material";
```
michael-jaia: I think I like the name Dashboard better. What do you think?

src/web/components/RemoteControlPanel/RemoteControlPanel.tsx
```diff
@@ -39,26 +48,41 @@ const rudderPercentages = new Map<number, number>([
...
 
         return () => {
             clearInterval(rcCommandInterval);
         };
     }, [throttleDirection, throttleMagnitude, rudderDirection, rudderMagnitude]);
 
+    /**
+     * Sends RC commands periodically unless in dive mode.
+     */
+    const handleRcInterval = () => {
```
michael-jaia: NIT: RC

src/web/components/RemoteControlPanel/RemoteControlPanel.tsx
```diff
@@ -39,26 +48,41 @@ const rudderPercentages = new Map<number, number>([
...
         }, RC_COMMAND_TIMEOUT);
 
         return () => {
             clearInterval(rcCommandInterval);
         };
     }, [throttleDirection, throttleMagnitude, rudderDirection, rudderMagnitude]);
 
+    /**
+     * Sends RC commands periodically unless in dive mode.
+     */
```
michael-jaia, line 81: NIT: @returns {void} + remove period

src/web/components/RemoteControlPanel/DiveControls/DiveControls.tsx
```diff
@@ -0,0 +1,98 @@
...
+    botId: number;
+    handleRCDiveCommand: () => void;
+}
+
+/**
+ * Creates set of inputs for RC dives
+ */
+export function DiveInputs(props: DiveInputsProps) {
+    const handleInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+        props.onChange?.(e.target.name, +e.target.value);
```
michael-jaia: What does the + do here? I'm also thinking we may not need the question mark

src/web/components/RemoteControlPanel/RemoteControlPanel.tsx
```diff
@@ -213,17 +237,62 @@ export default function RemoteControlPanel(props: RemoteControlPanelProps) {
         return command;
     };
 
+    const handleRCDiveChange = (name: string, value: number) => {
```
michael-jaia, line 249: I'm wondering if a switch case might be better here for readability. We only have four parameters so it should not be too big.

src/web/components/RemoteControlPanel/RemoteControlPanel.tsx
```diff
@@ -213,17 +237,62 @@ export default function RemoteControlPanel(props: RemoteControlPanelProps) {
...
             rudderDirection={rudderDirection}
             rudderMagnitude={rudderMagnitude}
         />
     );
 
+    const RCDiveInputs = (
+        <DiveInputs rcDiveParameters={rcDiveParameters} onChange={handleRCDiveChange} />
+    );
+
+    const RCDiveCommand = (
```
michael-jaia: I don't think we need variables for these components since they are only used once

michael-jaia (changes requested):
Jeff, looks good, just left a few comments

src/web/components/RemoteControlPanel/DiveControls/DiveControls.tsx
JRoybot: shorthand way to cast sting as number...

src/web/components/RemoteControlPanel/RemoteControlPanel.tsx
JRoybot: OK, will do.  Was trying to be as concise as possible.

michael-jaia (dismissed):

PR #1363, opened 2025-10-17
bug/photo-timestamps-are-not-correct

marshalllele:
This branch was merged with the following branches for easier testing. These will all be merged into 2.y together:
[task/2.y/add-config-option-for-camera-jaia-1919
](https://github.com/jaiarobotics/jaiabot/pull/1241) by @edsanville 
[task/test-deploy
](https://github.com/jaiarobotics/jaiabot/pull/1366) by @ferrom-jaia 

Includes a date as a string in each CameraCommand message. The python driver then receives the date on the Pi 0 and sets the system clock according to the message.

Also cleans up the way that videos are started and stopped to prevent file corruption.

Tested on a vehicle by disconnecting the Hub's router from the internet, then sending both a CameraCommand message via the Bot's Liaison page, and sending the Bot on a dive, then watching the date change on the Pi 0.


ferrom-jaia (dismissed):
Looks great! 

Tested on fleet 1 bot 2

and fleet 6 bot 9

edsanville:
It looks good, but I notice that timezones aren't being handled.  I'm not sure if this will matter, or what timezone the Pi Zero is set to by default (probably UTC?), but it may result in some anomalies when copying the files to a computer with a different timezone?

I wonder if a Unix timestamp might work better, to dodge the timezone issue.

ferrom-jaia (approved):
Looks great. Static code review. Nick tested on the flat fleet.

ferrom-jaia:
> It looks good, but I notice that timezones aren't being handled. I'm not sure if this will matter, or what timezone the Pi Zero is set to by default (probably UTC?), but it may result in some anomalies when copying the files to a computer with a different timezone?
> 
> I wonder if a Unix timestamp might work better, to dodge the timezone issue.

Yes, we should use unix timestamp:

```
using namespace std::chrono;

auto now   = system_clock::now();
auto secs  = floor<std::chrono::seconds>(now);
std::int64_t epoch_s = secs.time_since_epoch().count();
```

Let's roll with what we have for now, but I will make a follow on ticket just because we do not have a lot of time for testing currently.

PR #1364, opened 2025-10-21
task/upload-download-to-disk/JAIA-2132

JRoybot:
Implemented the save and load to file functions.  These were called Export and Import to make it distinct form save load from local storage.  

Functions related to loading and storing mission sets were moved from `data/mission_set/mission-set.ts` to `components/MissionSetStorage/mission-set-storage.ts` and grouped with other helpers from `utils/local-storage.ts` .  

Common code between Load/Import and Save/Export factored out to re-usable components.

Goal was to make the Import and Load functions as similar as possible.  However due to the asynchronous nature of using the system file tool the pattern used for Load did not makes sense for Import.  Reading and parsing a file is too heavyweight and asynchronous to put in a React Reducer.  Existing code for Load from Local Storage was refactored so the parsing is done prior to send the action to the `jaiaReducer` and a snapshot of the parsed mission set is sent to the reducer as data.  This pattern was then re-used for the Import function.

Existing unit tests for mission set Saving and Loading were updated to use the refactored code.


src/web/components/MissionsPanel/MissionSetStorage/mission-set-storage.ts
```diff
@@ -0,0 +1,202 @@
...
+ * @param {string} saveName The key of the mission set to retrieve
+ *
+ * @returns {MissionSetSnapshot} Snapshot of mission set
+ *
+ * @notes Called by UI code, snapshot is sent to the reducer/action handler
+ */
+export function loadSnapshotFromLocalStorage(saveName: string) {
+    const allMissionSets = JSON.parse(localStorage.getItem("missionSets") || "{}");
+    const targetSet = allMissionSets[saveName] || {};
+    const missionsArray = Array.isArray(targetSet.missions)
```
michael-jaia, line 41: Jeff I'm thinking this code block might be easier to follow with an if statement. I'm having trouble seeing how the missionsArray gets populated.

src/web/components/MissionsPanel/MissionSetStorage/mission-set-storage.ts
```diff
@@ -0,0 +1,202 @@
...
+                resolve(null);
+                return;
+            }
+            try {
+                const targetSet = JSON.parse(await file.text());
+                if (!targetSet) {
+                    resolve(null);
+                    return;
+                }
+                const missionsArray = Array.isArray(targetSet.missions)
```
michael-jaia, line 164: Same as above comment

src/web/components/MissionsPanel/MissionSetStorage/mission-set-storage.ts
JRoybot: It first tries to get all saved missions sets from local storage with the key "missionSets".  There may not be any saved in which case the pars will be falsey so we use an empty set.  It then tries to get the specific mission set using the saveName, and that could also be falsey .  So I think it would require 2 ifs.  Once we know we have the mission set we are looking for it then just maps what is saved, an integer, and a mission to an array.

NOTE:  When save to local storage was developed we converted the missions Map stored in missionSet to an array and that process saves the key as a number and the mission.  The key would have been the mission ID of the mission.  However when the mission set is retrieved from local storage (or from a file, since I reused logic) the mission set in the data model is rebuilt using addMission so the missionIDs in the rebuilt mission are not the same as those in the saved mission set.  For that reason we do not really need to store the mission ID that was used as a key into the original map, we never use  it.  Things might be simpler if we changed what is being stored to be just the array of missions and throw the key away.  That would be a bigger change and would affect tests as well.  

Before changing this logic, we should discuss if you want to make a bigger change in how the mission sets are stored....

PR #1365, opened 2025-10-21
task/classify-bottom-dives-based-on-ascent-types

AdeX11:
### Summary
Previously, the model used maximum acceleration to classify bottom types. However, this metric did not reliably capture differences in dive behavior. The approach has now been updated to use ascent type, which offers a more behaviorally grounded and consistent signal. Specifically, powered ascents are expected to occur after soft bottom impacts, while unpowered ascents typically follow hard bottom impacts.

### Solution
Updated the classification logic to set a dive packet’s bottom type to Soft if a powered ascent occurs during the dive; otherwise, it is classified as Hard.

### Testing
Tested on multiple simulated dives. Verified that bottom type assignments align with ascent behavior and that no misclassifications or runtime errors occur.


ferrom-jaia (approved):
Looks great!

We will perform in water tests to validate.

PR #1368, opened 2025-10-24
task/apply-corrections-for-precise-depth-readings

marshalllele:
Removes the requirement for a vehicle to go vertical before a dive can begin. Instead, the vehicle checks that its motor stops spinning - verified by the arduino_to_pi.motor field == 1500.

Calculates the depth of the pressure sensor when a dive begins (D_0). This depth is calculated using the estimated distance along the vehicle from the pressure sensor to the waterline (`pressure_sensor_to_waterline`, configurable in `jaiabot/config/templates/bot/jaiabot_mission_manager.pb.cfg.in`), and the vehicle's pitch. TODO: Compensate for the vehicle's roll as well. 

Adds a compensation for the depth of the pressure sensor compared to the depth of the tail in order to more accurately measure the true water depth. Uses the distance from the pressure sensor to the end of the standoffs (`pressure_sensor_to_tail`, configurable in `jaiabot/config/templates/bot/jaiabot_mission_manager.pb.cfg.in`). This compensation assumes that the vehicle is vertical when it hits the seafloor.

There are two depth fields in the `pressure_adjusted` message - `sensor_depth` and `depth`. `Sensor_depth` adds D_0 and the adjusted pressure reading for its measurement. `depth` adds D_0, the adjusted pressure reading, and `pressure_sensor_to_tail` for its measurement. The `depth_achieved` reported is set to the maximum `depth` reached during the dive. 

Tested pressure readings in sim by sending the bot on a dive and saving the `pressure_adjusted`  measurements, then comparing the two maximum depth measurements against the `depth_achieved` in the dive packet. 

Tested the updated `dive_prep` state in sim by forcing the motor command to read either 1500 or 1600 - the vehicle immediately started its dive when this was set to 1500, but waited for the `dive_prep` timeout when set to 1600.


ferrom-jaia (approved):
Looks great!

Did a static review of the code and the data that was collected in the log.

Tested on bot 16 fleet 6

PR #1369, opened 2025-10-27
task/2.y/black-box-survey-tool

michael-jaia:
### Summary
This pull request creates the foundation for planning a survey with multiple JaiaBots. The sequence of steps is as follows:
1. Click the survey button
2. Add a start location to the map
3. Add a end location to the map
4. Drag to create the grid (use finger on touch screens and ctrl + mouse drag on laptops)
5. Set the task
6. Save

### Testing
Tested on tablet and computer working through the creation of various surveys

### Future Work
* Start and end survey tasks
* SRP settings
* Moving backwards in the planning steps


src/web/utils/input.ts
```diff
@@ -59,3 +60,26 @@ export function validateCoordinate(lat: string, lon: string) {
...
     return [lat, lon];
 }
+
+/**
+ * Converts a TaskType to a UI friendly string
+ *
+ * @param {TaskType} taskType Task name to be formatted
+ * @returns {string} Name of the task
+ */
+export function formatTaskMenuItem(taskType: TaskType) {
```
JRoybot: Consider a more general purpose utility to transform an enumerated value into a human readable string.  I did this for UNDO/REDO to transform descriptions to readable text.  It could be made generic and put in utils to be used for any enumerated type.  It is a lot less boilerplate code to maintain, does not need to be updated when enumerations change.

`src/web/context/handlers/history-handlers.ts`

```
function getActionDescription(action: JaiaActions) {
    return action
        .toLowerCase()
        .split("_")
        .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
        .join(" ");
}
```

src/web/components/Map/Map.tsx
```diff
@@ -101,6 +108,49 @@ export default function Map() {
...
+        let nextState: GridPlanningStates;
+
+        switch (gridPlan.getState()) {
+            case GridPlanningStates.ACCEPTING_MISSION_START_LOCATION:
+                gridPlan.setMissionStart(location);
+                gridLayer
+                    .getVectorLayer()
+                    .getSource()
+                    .addFeature(generateSurveyEndpoint(location, true));
+                nextState = GridPlanningStates.ACCEPTING_MISSION_END_LOCATION;
```
JRoybot, line 129: Is the consensus that a strictly sequential planning process is the best way?  Alternatively we could provide a menu to select which step you want to perform.  This may come in handy if, for example, the user goes through all the steps and then wants to move the start location.  Another option would be to provide Back and Forward buttons to allow the user to go back to a previous step, this is fairly common in websites that walk a user through a long process

src/web/context/handlers/waypoint-handlers.ts
```diff
@@ -117,32 +114,16 @@ export function handleChangeTaskParameter(mutableState: JaiaContextType, action:
  * @param {JaiaContextType} mutableState State object ref for making modifications
  * @returns {JaiaContextType} Updated mutable state object
  */
-export function handleToggleBottomDive(mutableState: JaiaContextType) {
-    const task = getWaypoint().getTask();
-
-    if (task.getIsBottomDive()) {
-        task.setIsBottomDive(false);
+export function handleToggleBottomDive(mutableState: JaiaContextType, action: JaiaAction) {
+    if (action.task.getIsBottomDive()) {
```
JRoybot: Couldn't we just toggle this without the if?

`action.task.setIsBottomDive(!action.getIsBottomDive())`

src/web/context/handlers/waypoint-handlers.ts
```diff
@@ -117,32 +114,16 @@ export function handleChangeTaskParameter(mutableState: JaiaContextType, action:
  * @param {JaiaContextType} mutableState State object ref for making modifications
  * @returns {JaiaContextType} Updated mutable state object
  */
-export function handleToggleBottomDive(mutableState: JaiaContextType) {
-    const task = getWaypoint().getTask();
-
-    if (task.getIsBottomDive()) {
-        task.setIsBottomDive(false);
+export function handleToggleBottomDive(mutableState: JaiaContextType, action: JaiaAction) {
```
JRoybot, line 117: passing in the task is a lot cleaner

src/web/data/mission_set/mission-set.ts
```diff
@@ -2,7 +2,7 @@ import Mission from "./mission";
 import { UNASSIGNED_ID } from "../../utils/constants";
 import { Speeds } from "../../types/protobuf-types";
 
-class MissionSet {
+export class MissionSet {
```
JRoybot: I couldn't find another reference to this, is export needed?

src/web/data/survey_planner/grid-plan.ts
```diff
@@ -0,0 +1,122 @@
...
+
+export class GridPlan {
+    private missionStart: GeographicCoordinate;
+    private missionEnd: GeographicCoordinate;
+    private numOfLanes: number;
+    private laneSpacing: number;
+    private pointSpacing: number;
+    private surveyTask: Task;
+    private state: GridPlanningStates;
+    private missions: Map<number, Mission>;
```
JRoybot, line 29: We've declared this Map of missions in a number of places, maybe we should declare it in mission-set.ts

src/web/openlayers/features/survey/survey-endpoints.ts
```diff
@@ -0,0 +1,52 @@
...
+import surveyEndIcon from "../../../style/icons/survey-end.svg";
+
+/**
+ * Creates a point on the map to indicate the start or end of a survey
+ *
+ * @param {GeographicCoordinate} location Lat/lon of start or end point
+ * @param {Boolean} isStart Whether or not this location is the survey start
+ * @returns {Feature} Survey start or end point to display on map
+ */
+export function generateSurveyEndpoint(location: GeographicCoordinate, isStart: boolean) {
```
JRoybot: Might be clearer to use an enumeration with values of START and END instead of boolean

src/web/openlayers/features/survey/survey-endpoints.ts
```diff
@@ -0,0 +1,52 @@
...
+    return feature;
+}
+
+/**
+ * Creates the style to be applied to the start/end points of a survey
+ *
+ * @param {boolean} isStart Determines which icon to use
+ * @returns {Style} Style to be applied to a survey endpoint
+ */
+function generateSurveyEndpointStyle(isStart: boolean) {
```
JRoybot: see above

src/web/openlayers/controls/touches.ts
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * Updates the isPanning property based on the operator's finger movements
+     *
+     * @param {number} numOfFingers Number of fingers currently touching screen
+     * @returns {void}
+     *
+     * @notes
+     * Zero indicates the operator has lifted their fingers from the screen
+     */
+    updateFingers(numOfFingers: number) {
```
JRoybot, line 21: Don't care if the number of fingers is 1, 3, 4, ?

src/web/openlayers/layers/vector/survey/grid-layer.ts
```diff
@@ -0,0 +1,262 @@
...
+        if (!this.centerLine) {
+            return;
+        }
+
+        const lanes = [];
+
+        this.layerSource.clear();
+
+        let distFromCenter = 0;
+        if (gridPlan.getNumOfLanes() % 2 === 0) {
```
JRoybot: suggest adding a comment indicating this is for even number of lanes

src/web/openlayers/layers/vector/survey/grid-layer.ts
JRoybot, line 1: I don't think we need the /survey folder for this, just put it with the other layers.

src/web/openlayers/layers/vector/survey/grid-layer.ts
```diff
@@ -0,0 +1,262 @@
...
+     * missions to be processed by the rest of the app
+     *
+     * @param {boolean} modifyDataModel Stores the missions in the data model
+     * @returns {void}
+     *
+     * @notes
+     * finalizeGrid(true) only needs to be called once. Multiple calls will
+     * result in duplicate missions.
+     */
+    finalizeGrid(modifyDataModel?: boolean) {
```
JRoybot: should this have a default value of false ?

src/web/context/handlers/mission-handlers.ts
```diff
@@ -149,3 +154,53 @@ export function handleLoadMissionSet(mutableState: JaiaContextType, action: Jaia
...
+            for (const [missionID, mission] of gridPlan.getMissions()) {
+                const waypoints = mission.getWaypoints();
+                for (let i = 0; i < waypoints.length; i++) {
+                    if (i === 0 || i === waypoints.length - 1) {
+                        continue;
+                    }
+                    waypoints[i].setTask(cloneDeep(gridPlan.getSurveyTask()));
+                }
+            }
+            missionSet.setMissions(gridPlan.getMissions());
```
JRoybot: should this be cloned?  Is there any chance the reference can be re-used by the survey planning tools?

src/web/context/action-configs.ts
```diff
@@ -71,15 +72,23 @@ export const actionConfigs: Map<JaiaActions, ActionConfig> = new Map([
...
+        { handler: handleChangeGridPlanningState, tracked: false },
+    ],
 
     // Waypoint & Task Actions
     [JaiaActions.ADD_WAYPOINT, { handler: handleAddWaypoint, tracked: true }],
     [JaiaActions.DELETE_WAYPOINT, { handler: handleDeleteWaypoint, tracked: true }],
     [JaiaActions.MOVE_WAYPOINT, { handler: handleMoveWaypoint, tracked: true }],
     [JaiaActions.SELECT_TASK, { handler: handleSelectTask, tracked: true }],
-    [JaiaActions.CHANGE_TASK_PARAMETER, { handler: handleChangeTaskParameter, tracked: true }],
+    [JaiaActions.CHANGE_TASK_PARAMETER, { handler: handleChangeTaskParameter, tracked: false }],
```
JRoybot: Playing with JCC, I see this as a loss of functionality.

src/web/context/action-configs.ts
```diff
@@ -71,15 +72,23 @@ export const actionConfigs: Map<JaiaActions, ActionConfig> = new Map([
...
     [JaiaActions.TOGGLE_BOTTOM_DIVE, { handler: handleToggleBottomDive, tracked: true }],
 
+    // Survey Actions
+    [JaiaActions.SURVEY_SELECT_TASK, { handler: handleSelectTask, tracked: false }],
+    [JaiaActions.SURVEY_TOGGLE_BOTTOM_DIVE, { handler: handleToggleBottomDive, tracked: true }],
```
JRoybot: Because none of the other survey tool actions are undoable, this has the effect of deleting the entire survey if it is undone.  Video of bug available, too big to attach here

JRoybot (changes requested):
Reviewed all the code and left several questions and suggestions.  Built and ran the tool using simulation and browser, noted some oddities.

michael-jaia:
> Does it make sense that the user can edit survey missions individually using the Missions Panel and Waypoint Panel

Yes, in the case where a waypoint or two needs to be shifted because of an obstacle or something

michael-jaia:
> Should the survey planning tool auto assign bots to missions so they do not have to go into the Missions Panel and do it?

Yes. Update: FIXED

michael-jaia:
> image There still seems to be some odd layering of the missions such that they produce a weave pattern, suspect it is result of how they are created from the center line out.

Yeah I was thinking the same. To fix, I'm thinking the array needs to be reordered at somepoint to move left to right.

src/web/components/Map/Map.tsx
michael-jaia: No strictly sequential is just for the first pass

src/web/data/survey_planner/grid-plan.ts
michael-jaia: I'm not totally following. My thinking was to have a separate map of missions for grid planning that gets saved to the mission set singleton when the planning is complete.

src/web/openlayers/layers/vector/survey/grid-layer.ts
michael-jaia: I was thinking for when we have different types of survey missions such as polygon, maybe it is too early, does that change how you see it?
 I also thought it helped match with the survey folder in the features directory

src/web/context/handlers/mission-handlers.ts
michael-jaia: Good call! I just added gridPlan.reset() to the survey cleanup in `handleMapModeChange` + did a clone for good measure

src/web/context/action-configs.ts
michael-jaia: Made it false temporarily for consistency

michael-jaia:
> Created a survey mission with dive tasks , assigned to bots and started missions. Observed bots executed dives at the start rendesvous, I do not think that is intended

Yes thank you, I forgot to fix that! Update: FIXED

src/web/context/action-configs.ts
michael-jaia: Added a new action SURVEY_CHANGE_TASK_PARAMETER that is not tracked. CHANGE_TASK_PARAMETER is back to tracked.

src/web/openlayers/controls/touches.ts
michael-jaia: 3 and 4 are handled by the OS and treated as pans (they do not trigger draw interactions). If number of fingers is 1, then that will be handled by the draw interaction (we really just need to safe guard against two finger movements)

src/web/data/survey_planner/grid-plan.ts
JRoybot: My comment was to define this as a named type or interface since we basically defined an identical type in several places.

src/web/openlayers/layers/vector/survey/grid-layer.ts
JRoybot: It's not a big deal, if you think there will be more files to put in here, fine.  Since it is not with the survey folder I don't see a lot of benefit in just making a folder for that reason.

PR #1370, opened 2025-10-27
task/fix-pyjaiaprotobuf-and-web-dependencies

edsanville:
The web apps and pyjaiaprotobuf package were rebuilding every time the `build.sh` script was run.  This fixes that, so they only rebuild when the real dependencies have changed.  This speeds up dev iteration for those of us using `build.sh`.


JRoybot (approved):
Changes look good.  I built the branch twice and verified the web apps were not rebuilt the second time.

ferrom-jaia (approved):
Looks great! Works as expected.

PR #1371, opened 2025-10-27
feature/new-gps-navigation-algorithm

edsanville:
Use the GPS heading uncertainty and the distance to the next waypoint to issue EvHeadingUncertaintyExceeded and EvHeadingUncertaintyRecovered messages, processed by the state machine to enter and leave the pause::ReacquireGPS state.

I also did some reorganizing of code to separate files (events.h, states.h, MissionManager.h, MissionManagerStateMachine.h, etc).  I can revert these changes if necessary.  I only made these changes to help me understand the structure of the `jaiabot_mission_manager` app.


edsanville:
Oh, there may be an issue with the Trail state, since it doesn't have a goal per se.

I believe we're going to want to check for the `trail` message, and if it's there we can use the `range` field instead of the distance to next goal.  I can make these changes if it's confirmed that this makes sense.

edsanville:
New PR without refactoring posted.

PR #1372, opened 2025-10-30
task/2.y/integrate-offline-maps-into-dev-friendly

michael-jaia:
### Summary
This pull request folds Ed work (https://github.com/jaiarobotics/jaiabot/pull/1319) into the developer-friendly-jcc codebase. Changes include:
* Merging `HubMapDownloader` into `OfflineLayerManager`
* Removing the need for the subscribe-notify structure

### Testing
1. Upload and view a geoTIFF in the JCC
2. Download map tiles and display them in the JCC
3. Remove map tiles


src/web/context/handlers/data-model-handers.ts
JRoybot, line 19: During review I realized these constants are not referenced anywhere else so did not need to be exported, could change that in this PR if desired

src/web/utils/file.ts
JRoybot, line 1: Might be able to refactor `loadSnapshotFromFile()` in `MissionsPanel/MissionSetStorage/mission-set-storage.ts` when that gets revisited to simplify the way missions are handled.

src/web/components/SettingsPanel/LayerSwitcherMenu/LayerSwitcherMenu.tsx
```diff
@@ -218,7 +217,83 @@ export default function LayerSwitcherMenu() {
...
+                </Accordion>
             </ThemeProvider>
         </div>
     );
 }
+
+/**
+ * Renders the list of offline layers that the operator can toggle on/off
+ */
+function OfflineLayersAccordionDetails(props: Props) {
```
JRoybot, line 239: I found this part of the code hard to understand.  Why can't the accordion states be managed like other accordions and maintain the state in JaiaContext?

src/web/openlayers/layers/layers.ts
```diff
@@ -1,4 +1,7 @@
+import TileLayer from "ol/layer/Tile";
```
JRoybot: None of these 3 added imports are ever referenced in the code.

JRoybot (commented):
I reviewed the code, didn't see anything that stood out as incorrect, but did not fully understand the details of all of it.  Built and ran the branch.  I used the Save Map to Hub feature and then disconnected the computer from the network to see how the map would render.   

Based on limited testing

1. Didn't seem the entire map area was supported by the downloaded map, appearded to be a smaller area.
2. I had Open Street Maps as the base layer and got a mix of Open Street and NOAA ENC tiles.

Unclear how to use the upload GeoTIFF feature, assume something must be saved from another app to use this feature.

I would like to see a demo of how to use this before approving.

PR #1373, opened 2025-10-31
bug/multiple-tpv-sky-per-cycle

ferrom-jaia:
## Description
We were getting 3 TPV messages per 200ms. This caused our log files to grow large. The issue stemmed from gpsd receiving nmea strings with varying levels of data/information. GPSD would then create multiple TPV messages with varying levels of information. To fix this we moved from nmea to ubx protocol. This allowed us to get a dump of the gps data and pipe it to gpsd. 

## Tests
Tested on fleet 2 bot 2 and hub 1


marshalllele (approved):
Looks good. Tested on Fleet 1 Bot 1 and Fleet 1 Hub 1. Verified that only 5 GPS points are being logged each second, as opposed to 15. Also, found that log files increase by ~0.5 MB / min less then before.

PR #1374, opened 2025-10-31
bug/no-time-sync-with-gps

ferrom-jaia:
## Description

Updating backup date script to include using gps time when appropriate.

Logic:

* On boot set the clock to what is in the time keeper file. 
* On a loop check if the system clock is synchronized with ntp
  * If it is then set the time keeper file
  * If it is not then check if gps has a valid time
    *  If it does then set the time keeper file with gps time
    * Set the system clock once if the clock has not been synchronized with ntp

## Tests
Fleet 2 bot 2


marshalllele (approved):
Looks great! Tested on the flat fleet (F1B1 + F1H1).

PR #1376, opened 2025-11-03
task/2.y/add-back-simulation-banner/JAIA-2136

JRoybot:
Added the simulation banner to JCC

The SimulationBanner component sets up it's own state variable and polling to monitor the metadata and displays **Simulation** if in sim mode, returns null otherwise.

Moved declaration of Metadata used by JaiaAbout to protobuff-types

image


michael-jaia (dismissed):
Nice work Jeff, I just added some minor cleanup.

PR #1379, opened 2025-11-05
task/2.y/sub-surface-current-measurements/JAIA-2149

marshalllele:
Adds feature where JaiaBot will automatically record an estimated subsurface current vector if:

1. There is exactly one hold during a dive
2. That hold is at least 10 seconds long (can be configured in mission_manager config file)
3. The dive task has a drift of at least 15 seconds

The subsurface current vector is sent back to the hub inside of the task packet message, and then displayed inside of a dive's task packet panel in JCC. The subsurface current vector fields will only display in the JCC if there is a valid measurement reported. 

Tested in simulator, to be tested on the water.


ferrom-jaia (approved):
Looks good! 

Static code review. 

Nick tested on the water fleet 2 bot 13 and hub 1

PR #1380, opened 2025-11-05
task/2.y/integrate-start-end-tasks-to-survey-planning/JAIA-2148

michael-jaia:
### Summary
This pull request adds the ability for modifying the start of the mission set task and end of the survey task when mission planning. This feature is needed for surf zone surveys where a constant heading is required at the start or end to punch through the waves.

### Testing
Walked through multiple survey planning scenarios without any unexpected behaviors by the tool

### Upcoming Work
Allow operator to drag on screen to set constant heading direction


src/web/components/Map/Map.tsx
```diff
@@ -137,13 +137,13 @@ export default function Map() {
                 nextState = GridPlanningStates.ACCEPTING_GRID_DRAWING;
                 break;
             case GridPlanningStates.ACCEPTING_GRID_DRAWING:
-                // No state change on accidental clicks
-                nextState = GridPlanningStates.ACCEPTING_GRID_DRAWING;
-                break;
+                return;
```
JRoybot, line 142: Use break for consistency.

src/web/components/WaypointPanel/TaskParameters/TaskParameters.tsx
JRoybot, line 50: I think we will need to figure out how to send the SURVEY_CHANGE_TASK_PARAMETER for changes associated with the  survey so when we integrate the survey actions with Undo.  The way it is now breaks the undo pretty badly because this is being tracked for both survey and regular missions.

src/web/components/SurveyPlanner/SurveyPlanner.tsx
```diff
@@ -211,46 +250,118 @@ function GridConfigs(props: Props) {
                 </div>
             </div>
             <div className="button-row">
-                <button onClick={() => props.handleSetTaskClick()}>Set Task</button>
+                <div onClick={() => props.handleSetTaskClick()}>
+                    <Icon path={mdiArrowRight} />
```
JRoybot: I think it would be useful to add some helper text, it might not be immediately clear what the Arrow button does.

src/web/data/survey_planner/grid-plan.ts
```diff
@@ -25,6 +27,8 @@ export class GridPlan {
     private laneSpacing: number;
     private pointSpacing: number;
     private surveyTask: Task;
+    private startTask: Task;
```
JRoybot, line 30: Is it worth considering making this an array, or dictionary that can be indexed for the 3 different task uses, it would allow of the redundant code to be generalized.

JRoybot (changes requested):
Nice Job!.  Built and tested branch with simulator.  Left a few minor suggestions and questions.

src/web/components/Map/Map.tsx
michael-jaia: Jeff I use return because I want to exit the function before dispatching in those situations. Would suggest a different structure or keep as is?

src/web/components/SurveyPlanner/SurveyPlanner.tsx
michael-jaia: Are you thinking about a little text like "Next planning step" when a user hovers over the arrow?

src/web/data/survey_planner/grid-plan.ts
michael-jaia: Are you talking about the getters and setters here? If there are other instances it could be a good change, if not I don't think it is necessary. (I did like being able to call gridPlan.getStartTask() and gridPlan.getEndTask(), it felt like it made some of the code easier to follow).

src/web/components/Map/Map.tsx
JRoybot: Got it, for now keep it as is.  (depending on how it goes these might end up dispatching events to support undo)

src/web/components/SurveyPlanner/SurveyPlanner.tsx
JRoybot: Yes, or better yet use the next value  in the grid state enum  to form specific strings  like "Enter End Point"  (might be overkill)

src/web/data/survey_planner/grid-plan.ts
JRoybot: I guess it is a matter of style.  
There are several places where a switch or if then else if else if blocks were needed because of the discrete setters and getters.  My preference is always to consolidate code where possible.

Not a big deal, just wanted to trigger the thought process....

JRoybot (dismissed):
Thanks for addressing my comments.  No show stoppers.  Approved to merge when you are ready.

PR #1382, opened 2025-11-05
bug/2.y/no-camera-mode-parameter-crashes-sim

marshalllele:
add try/except block to config/gen/common/__init__.py


ferrom-jaia (approved):
Looks good!

PR #1390, opened 2025-11-07
bug/csv-download-lockup

edsanville:
JDV was locking up when user tried to download a CSV when the plots were at default timeRange (full timeRange).  This fixes that.


ferrom-jaia (approved):
Looks great!

Tested with test data in jdv

PR #1393, opened 2025-11-09
task/2.y/update task packet panel parameters/JAIA-2159

marshalllele:
Shows bottom type in the task packet panel if there was a bottom dive. 

Only shows average depth if we have exactly one hold during a dive. Because of this, bottom dives will also output an average depth.


ferrom-jaia:
We should only show subsurface current date when there is exactly one hold as well

ferrom-jaia (approved):
Looks good!

PR #1396, opened 2025-11-12
task/add-back-constant-heading-map-interactions/JAIA-2152

michael-jaia:
### Summary
This pull request adds back the "Select on Map" feature for setting a constant heading task.

image

image

### Testing
* Edit mode and tap to move work in sync with select on map
* The next connecting line starts at the end of the constant heading projection
* Select on map works for survey start and end tasks


src/web/components/Map/Map.tsx
```diff
@@ -153,6 +160,55 @@ export default function Map() {
...
+     * based on click location
+     *
+     * @param coordinate Location of click on map
+     * @returns {void}
+     */
+    const handleConstantHeadingSelectClick = (coordinate: Coordinate) => {
+        let startLocation;
+        let task;
+
+        if (jaiaGlobal.getMapMode() === MapModes.SURVEY_CONSTANT_HEADING_SELECT) {
```
JRoybot: because this is only called from within the switch on map mode I think it would be cleaner to pass the map mode in as a parameter or collapse the two cases of the switch since there is no difference in them

src/web/components/Map/Map.tsx
```diff
@@ -53,6 +54,12 @@ export default function Map() {
             case MapModes.SURVEY_PLANNING:
                 handleSurveyPlanningClick(event.coordinate);
                 return;
+            case MapModes.SURVEY_CONSTANT_HEADING_SELECT:
```
JRoybot, line 59: see comment below.

src/web/components/Map/Map.tsx
```diff
@@ -153,6 +160,55 @@ export default function Map() {
...
+                key: TaskParameterKeys.HEADING,
+                value: params.constant_heading,
+            },
+            {
+                key: TaskParameterKeys.CONSTANT_HEADING_TIME,
+                value: params.constant_heading_time,
+            },
+        ];
+
+        jaiaDispatch({
```
JRoybot, line 210: I expected we would have different actions based on whether the task is a survey task or not.  ... might be follow on work to integrate with undo.

src/web/components/WaypointPanel/TaskParameters/__tests__/TaskParameters.test.tsx
```diff
@@ -31,7 +31,7 @@ test("Render drift parameters", () => {
     expect(driftParams).toHaveAttribute("name", TaskParameterKeys.DRIFT_TIME);
 });
 
-test("Render constant heading parameters", () => {
+test.skip("Render constant heading parameters", () => {
     mockTask.setType(TaskType.CONSTANT_HEADING);
     render(<TaskParameters task={mockTask} isDisabled={false} />);
```
JRoybot, line 36: was this branch scanned for circular dependencies using madge?  possible reason for failure

src/web/context/handlers/waypoint-handlers.ts
```diff
@@ -119,6 +131,38 @@ export function handleToggleBottomDive(mutableState: JaiaContextType, action: Ja
...
+        updatedMapMode = MapModes.SURVEY_PLANNING;
+    }
+
+    if (!action.task.getIsSurveyTask() && currentMapMode !== MapModes.CONSTANT_HEADING_SELECT) {
+        updatedMapMode = MapModes.CONSTANT_HEADING_SELECT;
+        jaiaGlobal.getSelectedWaypoint().isMoveable = false;
+    }
+
+    handleMapModeChange(updatedMapMode);
+    mutableState.mapMode = updatedMapMode;
```
JRoybot, line 161: No need to update openlayers?

src/web/context/handlers/waypoint-handlers.ts
```diff
@@ -119,6 +131,38 @@ export function handleToggleBottomDive(mutableState: JaiaContextType, action: Ja
...
+) {
+    const currentMapMode = jaiaGlobal.getMapMode();
+    let updatedMapMode = MapModes.DEFAULT;
+
+    if (
+        action.task.getIsSurveyTask() &&
+        currentMapMode !== MapModes.SURVEY_CONSTANT_HEADING_SELECT
+    ) {
+        updatedMapMode = MapModes.SURVEY_CONSTANT_HEADING_SELECT;
+    } else if (action.task.getIsSurveyTask()) {
```
JRoybot: already tested this condition, could this be

```
if(action.task.getSurveyTask){
    if( currentMapMode !== MapModes.SURVEY_CONSTANT_HEADING_SELECT) {
        updatedMapMode = MapModes.SURVEY_CONSTANT_HEADING_SELECT
        } else {
         updatedMapMode = MapModes.SURVEY_PLANNING;
         }

```

JRoybot (changes requested):
Built and tested branch all works as expected.  Left a few suggestions on code.  Expected to see use of different actions depending on if the task parameter is a survey task or not... might do as part of the  on-going undo integration task.

src/web/components/WaypointPanel/TaskParameters/__tests__/TaskParameters.test.tsx
michael-jaia: Good thought, just checked with madge, no circular dependencies

src/web/context/handlers/waypoint-handlers.ts
michael-jaia: No because we are just switching the toggle no map changes need to occur.

JRoybot (dismissed):

PR #1397, opened 2025-11-12
task/2.y/integrate-survey-tool-with-undo-button/JAIA-2145

JRoybot:
This task started as "integrate survey tool with undo" however during implementation many limitations of the existing undo/redo functionality were found resulting in rework of existing code.  

- Survey planning can now be undone after the plan is finalized.  
    - We decided to wait until more of the planning data for interim steps is stored in the data model to support undo
- Added basic support to update gridLayer when syncing openLayers from handlers
- Used complete objects from the data model into JaiaContext in place of individual data items to provide more consistent access to the data and eliminate references to private data in the objects that can become obsolete when state is restored by undo
- Replaced circular bi-directional history buffer with a simple stack for undo and removed existing redo functions
    - Follow on with adding separate redo stack and re-implementing redo functions
    - Revisit circular buffer for the stacks to avoid constantly shifting the array used for the buffer (costly array copy)
- Moved history buffer out of JaiaContext and made part or the data model
- Implemented a psuedo Memento pattern for capturing and restoring state snapshots
    - Data model classes are now responsible for capturing and restoring their own snapshots
    - Snapshot interfaces definitions are public, not nested immutable classes
    - Data model classes provide capture and restore functions for their snapshots
    - State variables not in the data model are encapsulated into a snapshot by the history-handler
- Utilized common missionSet snapshots for state history and save/load import/export functions


src/web/context/handlers/history-handlers.ts
```diff
@@ -134,4 +105,15 @@ function updateDataFromSnapshot(snapshot: JaiaHistoryType) {
     jaiaGlobal.setSelectedWaypoint(snapshot.selectedWaypoint);
     jaiaGlobal.setSelectedNode(snapshot.selectedNode);
     jaiaGlobal.setSelectedTaskPacket(snapshot.selectedTaskPacket);
+
+    // Update Survey Plan
```
michael-jaia: Is there we just want to save a copy of the gridPlan object? I also think gridPlanDetails is redundant if we use gridPlan

src/web/data/history/history.ts
```diff
@@ -0,0 +1,8 @@
+import { JaiaContextType } from "../../types/context-types";
+import HistoryBuffer from "./history-buffer";
+import { MAX_HISTORY } from "../../utils/constants";
+
+const emptyState = {} as JaiaContextType;
```
michael-jaia: Does const emptyState: JaiaContextType = {} work?

src/web/data/history/history.ts
```diff
@@ -0,0 +1,8 @@
...
+// The history singleton is being initialized with an empty state
+// It will be reset by the handleInit of the data-modle-handlers
```
michael-jaia: NIT: Fix typo

src/web/openlayers/layers/vector/grid-layer.ts
```diff
@@ -40,7 +40,26 @@ class GridLayer extends JaiaVectorLayer {
         this.layerSource = this.getVectorLayer().getSource();
     }
 
-    override updateFeatures() {}
+    override updateFeatures() {
```
michael-jaia, line 44: Can we use finalizeGrid here? It looks like the same code

src/web/context/handlers/history-handlers.ts
JRoybot: Yes, that is part of step 2, use the top level objects instead of all the different components held by them.  That includes the missionSet, missionManager, gridPlan & jaiaGlobal, perhaps others.  Then this function will be very clean and will not need to be updated if an additional component is added to any of them.  I used gridPlanDetails because it already existed and reduced some of the wordiness in the interim.

src/web/data/history/history.ts
JRoybot: I checked, it needed to be typed the same as what is used to instantiate the generic and without any assignment there is no way the compiler can infer it's type Typescript flags it.

src/web/openlayers/layers/vector/grid-layer.ts
JRoybot: This was a "quick fix" I copied most of it from finalizeGrid, just added the defensive if so sortLanes would not blow up and removed the logic to update the data model.  I planned to revisit this and am very but our focus changed to refactoring the way history buffer was used.  I am open to any suggestions.

src/web/data/history/history.ts
JRoybot: fixed

src/web/context/handlers/selection-handlers.ts
```diff
@@ -27,7 +27,6 @@ export function handleClickedNode(mutableState: JaiaContextType, action: JaiaAct
     jaiaGlobal.setSelectedNode(action.clickedNode);
     const selectedNode = jaiaGlobal.getSelectedNode();
```
JRoybot: delete

src/web/context/handlers/selection-handlers.ts
```diff
@@ -42,8 +41,9 @@ export function handleClickedNode(mutableState: JaiaContextType, action: JaiaAct
  * @returns {JaiaContextType} Updated mutable state object
  */
 export function handleClickedTapToMove(mutableState: JaiaContextType) {
-    mutableState.selectedWaypoint.isMoveable = !mutableState.selectedWaypoint.isMoveable;
-    jaiaGlobal.setSelectedWaypoint(mutableState.selectedWaypoint);
+    const muteableWaypoint = jaiaGlobal.getSelectedWaypoint();
+    muteableWaypoint.isMoveable = !muteableWaypoint.isMoveable;
+    jaiaGlobal.setSelectedWaypoint(muteableWaypoint);
```
michael-jaia: I don't think we need the setter here. Tweak spelling from muteable to mutable

src/web/data/jaia_global/jaia-global.ts
```diff
@@ -7,6 +7,15 @@ import {
 } from "../../types/jaia-system-types";
 import { MapFeatureTypes, MapModes } from "../../types/openlayers-types";
 import { UNASSIGNED_ID } from "../../utils/constants";
+import cloneDeep from "lodash/cloneDeep";
+
+export interface JaigGlobalSnapshot {
```
michael-jaia: NIT: Typo

src/web/data/jaia_global/jaia-global.ts
```diff
@@ -93,6 +102,38 @@ class JaiaGlobal {
...
         this.defaultTaskParameters = defaultTaskParameters;
     }
+
+    /**
+     * Captures a snapshot of JaiaGlobal
+     *
+     * @returns {JaigGlobalSnapshot} snapshot of current Grid Plan
+     */
+    captureSnapshot() {
+        const currentGlobals = {
```
michael-jaia: MINOR: Would you be on board with call this snapshot?

src/web/data/missions_manager/missions-manager.ts
```diff
@@ -126,27 +132,36 @@ class MissionsManager {
...
-        this.botsToMissions.clear();
-        this.missionsToBots.forEach((botID, missionID) => {
-            this.botsToMissions.set(botID, missionID);
-        });
+
+    restoreFromSnapshot(snapshot: MissionsManagerSnapshot) {
+        // Clear current assignments
+        this.clear();
+
+        for (const [k, v] of snapshot.botsToMissions) {
```
michael-jaia: Just out of curiosity, did things start breaking down if we did this.botsToMissions = snapshot.botsToMissions? Just thinking if we need the for loop

JRoybot: Not entirely sure.  It is required for missionSet because the snapshot converts Map to array to support Json conversion.  For the others I was not sure but thought better safe than sorry after all the trouble Maps have been in the past.  I am open to direct copying but, risk is if anything has a reference to the old Map it will break and the bug may not be obvious right away.

src/web/components/MissionsPanel/MissionSetStorage/mission-set-storage.ts
```diff
@@ -30,20 +31,21 @@ export function loadSnapshotFromLocalStorage(saveName: string) {
...
             ]),
         );
     }
 
-    return {
+    const snapshot: MissionSetSnapshot = {
         missions: missions,
         nextMissionID: targetSet.nextMissionID ?? 0,
-        missionIDInEditMode: targetSet.missionIDInEditMode ?? null,
+        missionIDInEditMode: targetSet.missionIDInEditMode ?? UNASSIGNED_ID,
```
JRoybot, line 44: fixing unit test for this change

PR #1398, opened 2025-11-14
bug/2.y/cancel-waypoint-changes

michael-jaia:
### Summary
This pull request fixes the bug where the base waypoint (the state of the waypoint when it is first selected) did not changes on waypoint-to-waypoint selections. It only updated on the first render of the panel which happens on open/close actions not switches between waypoints.

### Solution
Store a copy of the selectedWaypoint object and check that against the latest selectedWaypoint data to detect waypoint-to-waypoint switches. Then on a switch update the baseWaypoint which will be used to revert the state on a cancel event.


src/web/utils/comparisons.ts
```diff
@@ -7,7 +7,7 @@ import { SelectedWaypoint } from "../types/jaia-system-types";
  * @param {SelectedWaypoint} waypointB Waypoint data used in comparison
  * @returns {boolean} True if the waypoints match, false if they do not
  */
-export function compareWaypoints(waypointA: SelectedWaypoint, waypointB: SelectedWaypoint) {
+export function compareSelectedWaypoints(waypointA: SelectedWaypoint, waypointB: SelectedWaypoint) {
```
JRoybot: I would put this function in WaypointPanel.tsx since it is only used there and unlikely will be used anywhere else.

src/web/context/handlers/panel-handlers.ts
```diff
@@ -35,11 +35,11 @@ export function handleClosedDetails(mutableState: JaiaContextType) {
...
-        const originalWaypoint = Object.setPrototypeOf(action.waypoint, Waypoint.prototype);
-        const waypoints = missionSet
-            .getMission(jaiaGlobal.getSelectedWaypoint().missionID)
-            .getWaypoints();
-        waypoints[jaiaGlobal.getSelectedWaypoint().waypointNum - 1] = originalWaypoint;
+        const mission = missionSet.getMission(jaiaGlobal.getSelectedWaypoint().missionID);
+        // Reset waypoint to state when first selected
+        mission
+            .getWaypoints()
+            .splice(jaiaGlobal.getSelectedWaypoint().waypointNum - 1, 1, action.waypoint);
```
JRoybot: can you explain what this does? Maybe a comment is a good idea.

src/web/components/WaypointPanel/WaypointPanel.tsx
```diff
@@ -51,11 +52,21 @@ export default function WaypointPanel() {
...
     const [latInput, setLatInput] = useState(getWaypoint().getLocation().lat.toString());
     const [lonInput, setLonInput] = useState(getWaypoint().getLocation().lon.toString());
+    const [baseWaypoint, setBaseWaypoint] = useState(null);
     const isDisabled = jaiaContext.missionIDInEditMode !== jaiaContext.selectedWaypoint.missionID;
 
     useEffect(() => {
-        originalWaypoint = cloneDeep(getWaypoint());
-    }, []);
+        // Initial assignment on panel's first render
+        if (baseWaypoint === null) {
```
JRoybot: Instead of putting a waypoint in state, would it be simpler to just store a `SelectedWaypoint` typed variable

src/web/components/WaypointPanel/WaypointPanel.tsx
```diff
@@ -208,17 +219,18 @@ export default function WaypointPanel() {
...
      */
     const handleClosePanelClick = (panelAction: PanelActions) => {
-        let waypoint: Waypoint;
-
         if (panelAction === PanelActions.CANCEL) {
-            waypoint = originalWaypoint;
+            jaiaDispatch({
+                type: JaiaActions.CLOSED_WAYPOINT_PANEL,
+                panelAction: panelAction,
+                waypoint: baseWaypoint,
```
JRoybot: Does the panel-handler logic really need to have the base waypoint data when the user selects cancel?  Couldn't the waypoint just be reset in this code before dispatching the close event? and then could handling Done and Cancel be the same,

src/web/context/handlers/panel-handlers.ts
JRoybot: See other related comment on whether the waypoint should be modified here or in the panel code.

JRoybot (changes requested):
Seems to work.  I found the handling between the panel and the action handler awkward, made a few suggestions that might be able to simplify it.

PR #1399, opened 2025-11-14
task/2.y/add-back-3d-depth-map

michael-jaia:
### Summary
This pull request adds back the 3D depth map. The functionality remains the same. When a operator clicks on the 2D depth map, a window appears displaying the depth data in 3D using plotly.

### Testing
* Opening + closing depth map
* Changing plot orientation
* Moving 3D plot window left, right, and center

image


JRoybot (approved):
Built and tested branch , reviewed code.  All looks good.

PR #1402, opened 2025-11-15
task/2.y/default-lanes-to-num-of-bots

michael-jaia:
### Summary
Adds the function `initNumOfLanes` in `SurveyPlanner.tsx` to set the number of lanes to the last user setting, the number of Bots connected or DEFAULT.

### Testing
* Connected 3 Bots in simulation and watched the number of lanes be set correctly as I updated the value and opened/closed the panel

### DIFF
--- a/src/web/components/SurveyPlanner/SurveyPlanner.tsx
+++ b/src/web/components/SurveyPlanner/SurveyPlanner.tsx
@@ -11,9 +11,11 @@ import TaskParameters from "../WaypointPanel/TaskParameters/TaskParameters";
 
 import { gridLayer } from "../../openlayers/layers/vector/survey/grid-layer";
 import Task from "../../data/tasks/task";
+import { bots } from "../../data/bots/bots";
 import { gridPlan, GridPlanDetails, GridPlanningStates } from "../../data/survey_planner/grid-plan";
 import { formatNumericalInput, snakeCaseToTitleCase } from "../../utils/input";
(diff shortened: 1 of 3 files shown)


src/web/components/SurveyPlanner/SurveyPlanner.tsx
```diff
@@ -171,7 +173,29 @@ function RequestEndMissionLocation() {
...
+    /**
+     * Determines the number of lanes to show on first render.
+     * The last user input will be used unless no input has been provided yet.
+     * In that case, we use the number of Bots or DEFAULT_LANES if no Bots connected.
+     *
+     * @returns {number} Number of lanes to show on first render
+     */
+    const initNumOfLanes = () => {
+        const numOfBots = bots.getBots().size;
+        if (gridPlan.getNumOfLanes() === INIT_LANES && bots.getBots().size > 0) {
```
JRoybot, line 185: I think using UNASSIGNED_ID for this would be fine, not sure we need another -1 contant

JRoybot (approved):
Built and tested branch with sim.  All looks good.  Left one minor suggestion.

src/web/components/SurveyPlanner/SurveyPlanner.tsx
michael-jaia: I was on the fence, but I want to keep INIT_LANES for readability because I don't want someone to be asking what does unassigned ID have to do with number of lanes.

PR #1403, opened 2025-11-15
task/add-back-pam-functionality-to-tasks

michael-jaia:
### Summary
This pull request adds back the option for operators to toggle the hydrophone on/off for dive and drift tasks when a JaiaBot PAM is connected to the Hub.

image

### Testing
* Individual dive + drift tasks update accordingly based on toggle interactions
* Survey dive + drift tasks update accordingly based on toggle interactions


src/web/context/handlers/waypoint-handlers.ts
```diff
@@ -119,6 +122,25 @@ export function handleToggleBottomDive(mutableState: JaiaContextType, action: Ja
...
+ *
+ * @param {JaiaContextType} mutableState State object ref for making modifications
+ * @returns {JaiaContextType} Updated mutable state object
+ */
+export function handleToggleHydrophone(mutableState: JaiaContextType, action: JaiaAction) {
+    action.task.setUseHydrophone(!action.task.getUseHydrophone());
+    missionLayer.updateFeatures();
+    // Refresh the grid to show icon updates
+    if (
+        mutableState.mapMode === MapModes.SURVEY_PLANNING &&
```
JRoybot, line 146: We had discussed adding a flag to task to indicate if it is part of a survey or not and to use separate actions for changes to survey tasks and manual mission tasks so they could be tracked/untracked individually.  I think this will be very useful to clean up the undo/redo behaviors of all the task parameter changes.

src/web/data/tasks/task.ts
```diff
@@ -31,7 +31,7 @@ export default class Task {
     MAX_SPEED_CONSTRAINT = 3;
 
     private isBottomDive: boolean;
-    private isEnablePAM: boolean;
+    private useHydrophone: boolean;
```
JRoybot, line 34: see comment in waypoint-hanlders

JRoybot (dismissed):
Built and tested with sim.  All looks good.  Left a comment with suggestion for follow on work.

PR #1404, opened 2025-11-17
task/2.y/update-ol-control-styles

michael-jaia:
### Summary
This pull request adds back the coordinates of the mouse as it moves over the map. It also updates the style for the map scale and attributions.

image


JRoybot (approved):
Reviewed code, built and tested branch with sim.   All looks good.

PR #1405, opened 2025-11-17
feature/display-the-sw-version-in-jdv-log-jaia-2156

edsanville:
It loads the DeviceMetadata object from the log file, and displays the entire message in a dialog in JSON format.  That way, all of the version information is there.

We could also use the DeviceMetadata object to print the entire version string at the top, something like

major.minor.patch (git_branch git_hash)

or similar to that.

As a byproduct, it is now possible to add string series to the plot.  The string value shows up in the data table and on the plot as hovertext.


src/web/jdv/client/src/ui/InformationDialog.tsx
```diff
@@ -0,0 +1,44 @@
...
+    onClose: () => void;
+}
+
+const METADATA_PATH = "jaiabot::metadata/jaiabot.protobuf.DeviceMetadata";
+
+export function InformationDialog(props: Props) {
+    const [dialogText, setDialogText] = React.useState("Loading metadata...");
+
+    useEffect(() => {
+        LogApi.getObjects<DeviceMetadata>(props.logFiles, METADATA_PATH)
```
marshalllele: I think we could map over all of the props.logFiles in this function if we wanted to display all of the files metadata in the same info dialog box.

src/web/jdv/client/src/ui/App.tsx
```diff
@@ -283,6 +312,15 @@ export class App extends React.Component {
...
         if (this.state.chosenLogs !== prevState.chosenLogs) {
             if (this.state.chosenLogs.length > 0) {
+                // Get the metadata for the selected logs
+                LogApi.getObjects<DeviceMetadata>(
+                    this.state.chosenLogs,
+                    "jaiabot::metadata/jaiabot.protobuf.DeviceMetadata",
+                ).then((metadata) => {
+                    console.log("Metadata for selected logs:");
+                    console.log(metadata);
+                });
```
marshalllele: I believe this call to LogApi.getObjects() is redundant as we're making the same call in InformationDialog.tsx when the user clicks the Info button. This seems to be adding extra time to the initial loading of a log file. When I remove it it doesn't seem to be adding any extra time to the metadata loading time when I click the Info button.

src/web/jdv/client/src/ui/App.tsx
```diff
@@ -283,6 +336,15 @@ export class App extends React.Component {
...
             if (this.state.chosenLogs.length > 0) {
+                // Get the metadata for the selected logs
+                // LogApi.getObjects<DeviceMetadata>(
+                //     this.state.chosenLogs,
+                //     "jaiabot::metadata/jaiabot.protobuf.DeviceMetadata",
+                // ).then((metadata) => {
+                //     console.log("Metadata for selected logs:");
+                //     console.log(metadata);
+                // });
+
```
marshalllele: I believe this call to LogApi.getObjects() is redundant as we're making the same call in InformationDialog.tsx when the user clicks the Info button. This seems to be adding extra time to the initial loading of a log file. When I remove it it doesn't seem to be adding any extra time to the metadata loading time when I click the Info button.

marshalllele (dismissed):
Everything looks great! Made some UI and functional changes. Tested in sim.

PR #1407, opened 2025-11-18
task/update-sensor-specs-markdown

marshalllele:
Update page 41 and page 122 in the Markdown docs to reflect updated message and sensor data structure.


ferrom-jaia (approved):
Looks great!

PR #1408, opened 2025-11-19
task/2.y/allow-multiple-lanes-per-bot

michael-jaia:
### Summary
This pull request allows an operator to plan survey missions where Bots can cover multiple lanes. It also caps the number of waypoints a Bot can perform in a single mission at 80.

### Testing
* Single mission does not exceed 80 waypoints
* No Bot covers more than 80 waypoints in a survey
* Grid updates based on the waypoint constraint as the number of Bots and lanes changes


src/web/data/survey_planner/grid-plan.ts
```diff
@@ -143,6 +158,70 @@ export class GridPlan {
...
+    /**
+     * When the number of lanes exceeds the number of Bots, the extra lanes
+     * will be shared among the Bots. Bots can be assigned multiple adjacent lanes.
+     * Each additional lane does not require the Bot to transit to the
+     * mission start + end points.
+     *
+     * @returns {void}
+     */
+    fitLanesToBots() {
+        const lanesPerBot = Math.floor(this.numOfLanes / this.numOfBots);
```
JRoybot, line 186: When running the branch I found it confusing that it allows me to set the number of bots greater than there are in the system and then when the survey plan is finalized the number of lanes is reduced to the number of bots.  I would expect that the UI would warn me that there was not that many bots available and IF the number of lanes was greater than the number of bots we would get that many lanes with bots doing laps until the 80 waypoint limit.

src/web/utils/constants.ts
```diff
@@ -31,3 +30,4 @@ export const KILOMETER_FACTOR = 1000;
 export const MAX_HISTORY = 15;
 export const MISSION_ENDPOINTS = 2;
 export const DEPTH_MAP_3D_NAME = "depth-map-3d";
+export const MAX_WAYPOINTS = 80;
```
JRoybot, line 33: When I tested this limit the final survey plan missions have 82 waypoints including start and end, is that intended or should the total including them be 80?

src/web/data/survey_planner/grid-plan.ts
JRoybot, line 1: I think it is confusing to have this file in a directory with a different name, could they be the same.  As commented on earlier PR, find the survey directory containing grid-layer is unneeded and makes finding the file difficult.

JRoybot (commented):
I reviewed the code and built and tested the branch in the sim.  I didn't expect the behaviors seen in testing, see comments.  The code looks good, although as a whole it is a bit hard to follow all the parts of survey planning with so much of the logic in the grid-layer itself.

michael-jaia:
> I reviewed the code and built and tested the branch in the sim. I didn't expect the behaviors seen in testing, see comments. The code looks good, although as a whole it is a bit hard to follow all the parts of survey planning with so much of the logic in the grid-layer itself.

I agree, that was a mistake on my part including too much logic in the layer itself. It needs to be pulled out in a separate ticket.

src/web/data/survey_planner/grid-plan.ts
michael-jaia: Removed the folder

src/web/utils/constants.ts
michael-jaia: Not intended, thank you for catching this. I updated this in calculateMaxPointsPerLane

src/web/data/survey_planner/grid-plan.ts
michael-jaia: https://github.com/user-attachments/assets/032b6c2a-1255-46ae-b070-36d7e5bd4265

michael-jaia: Jeff I think I'm missing something here, would you be able to share a video of what you are talking about?

JRoybot: I could not reproduce it from a clean state, I think what happened is I might have used undo at some point which, left grid planning in an odd state... not really sure but I do not think anything needs to be done.

JRoybot (approved):

PR #1409, opened 2025-11-20
task/2.y/add-back-task-packet-exclude-feature

michael-jaia:
### Sumamry
This pull request adds back the exclude and include buttons to the task packet panel. Task packets the hold bad sensor data can be removed to prevent them from impacting the contour map. They can be added back by going to the excluded task packet layer, clicking on them, and selecting include.

image

### Testing
1. Exclude a task packet
* Panel stays open
* Contour map updates
* Task packet is faded
* Task packet exists on excluded layer

2. Include a task packet
* Panel stays open
* Contour map updates
* Task packet is not faded
* Task packet removed from excluded layer


src/web/context/handlers/data-model-handers.ts
```diff
@@ -91,6 +91,6 @@ export function handleInit(mutableState: JaiaContextType) {
 export function handlePollDataModel(mutableState: JaiaContextType) {
     mutableState.bots = bots.getBots();
     mutableState.hubs = hubs.getHubs();
-    mutableState.taskPackets = taskPackets.getTaskPackets();
+    // mutableState.taskPackets = taskPackets.getTaskPackets();
```
JRoybot: remove dead code

src/web/openlayers/layers/vector/excluded-task-packets-layer.ts
```diff
@@ -0,0 +1,41 @@
...
+const LAYER_OPACITY = 0.6;
+
+class ExcludedTaskPacketsLayer extends JaiaVectorLayer {
+    constructor() {
+        super(
+            LayerTitles.EXCLUDED_TASK_PACKETS_LAYER,
+            layersZIndexes.get(LayerTitles.EXCLUDED_TASK_PACKETS_LAYER),
+        );
+        this.getVectorLayer().setOpacity(LAYER_OPACITY);
+        this.getVectorLayer().setVisible(false);
```
JRoybot, line 17: Why are we rendering an invisible layer, is there a way to ever see it?

JRoybot (changes requested):
few questions and a nitpick.  Tested in branch

Is this desired behavior?

Once a task packet is excluded I could not find a way to select it and set it back to included unless the panel is still tied to it, once you close the panel or select a different one how do you re-select it.

michael-jaia:
Jeff here is how an excluded task packet can be brought back:

https://github.com/user-attachments/assets/2b3e2719-f6a9-4352-ac78-0b02fc5ffa35

src/web/openlayers/layers/vector/excluded-task-packets-layer.ts
michael-jaia: By default the layer is turned off. the user can toggle it on

JRoybot (dismissed):

PR #1412, opened 2025-11-21
bug/handle-no-start-location-contours

edsanville:
Log a warning and skip dives with no start_location set when generating contours.


marshalllele (approved):
Looks good! Verified on a set of dives we know does not generate a contour map on 2.y.

PR #1413, opened 2025-11-21
task/2.y/add-back-ability-to-move-hub-in-sim

michael-jaia:
### Summary
This pull request adds back the ability to move the Hub in simulation. An operator can do this by manually changing the latitude and longitude or they can click a new point on the map. The location update is not instantaneous because the location needs to be updated on the server.

image

### Testing
1. Manually change the lat/lon inputs
2. Click on map to move Hub


src/web/components/SettingsPanel/MoveHub/MoveHub.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+        if (evt.target.name === CoordinateTypes.LAT) {
+            setLatInput(value);
+            lat = value;
+        } else {
+            setLonInput(value);
+            lon = value;
+        }
+
+        if (isNaN(Number(value))) {
```
ferrom-jaia, line 83: Do you think we should check the validity of the number before setting the lat/lon variable (moving this logic above)

src/web/components/SurveyPlanner/SurveyPlanner.less
```diff
@@ -27,15 +27,15 @@
     margin-bottom: 15px;
 }
 
-.input-grid input {
+.survey .input-grid input {
```
ferrom-jaia, line 30: Did you mean to add this in this pull request?

ferrom-jaia (changes requested):
Do we want this to be centered (The input boxes and the toggle): 
image

Bug: Selecting bot and then toggling move hub location. Results in placing wpts and moving the hub location.

src/web/components/SurveyPlanner/SurveyPlanner.less
michael-jaia: Yes, it was a small tweak I wanted to include

src/web/components/SettingsPanel/MoveHub/MoveHub.tsx
michael-jaia: We are updating the lat/lon in local state. It gets updated on each keystroke. So if an operator wants to use the coordinate -71, we would receive -, -7, -71.  - is not a number, so it would not show up in the input field. We let the operator type whatever, then we check if what they typed is a valid number, if it is we change the location of the Hub, if not we wait.

michael-jaia:
> Do we want this to be centered (The input boxes and the toggle): image
> 
> Bug: Selecting bot and then toggling move hub location. Results in placing wpts and moving the hub location.

I played with that before making the PR and did not love how it looked center (the MUI toggle elements takes up more space than the shape of the toggle so it does not look center). I'm sure there is a way around it, but not a high enough priority to investigate right now.

src/web/components/SettingsPanel/MoveHub/MoveHub.tsx
ferrom-jaia: Nice! Makes sense. I think adding a comment may help here.

michael-jaia: Added more to the docstring.

PR #1414, opened 2025-11-24
task/2.y/add-back-srp

michael-jaia:
### Summary
This pull requests adds back the ability to set the safety return parameters when planning a survey mission. During the planning steps the operator will be asked if they want to the SRP parameters, if they choose no, the mission planning is complete. If they choose yes, the parameters can be customized.

### Testing
image
image


src/web/components/Map/Map.tsx
```diff
@@ -207,11 +207,22 @@ export default function Map() {
             },
         ];
 
-        jaiaDispatch({
-            type: JaiaActions.CHANGE_TASK_PARAMETER,
-            task: task,
-            taskParameterPairs: taskParameterPairs,
-        });
+        if (
+            mapMode === MapModes.SURVEY_PLANNING ||
```
JRoybot, line 214: Would it be simpler to check if the Survey Panel is open?

src/web/context/handlers/waypoint-handlers.ts
```diff
@@ -112,7 +112,10 @@ export function handleChangeTaskParameter(mutableState: JaiaContextType, action:
         action.task.setParameter(taskParameterPair);
     }
 
-    if (jaiaGlobal.getMapMode() === MapModes.SURVEY_CONSTANT_HEADING_SELECT) {
+    if (
+        jaiaGlobal.getMapMode() === MapModes.SURVEY_CONSTANT_HEADING_SELECT ||
```
JRoybot: would visiblePanel be easier to check?

src/web/data/survey_planner/grid-plan.ts
JRoybot, line 260: suggested this on the original PR but worth mentioning again.  I think the code could be cleaner if instead of storing 4 individual types of tasks for grid plans have a dictionary or array that could be indexed by the task type.

JRoybot, line 1: General question on this functionality.  I have found it confusing that the terms Survey Planner and Grid Plan  and variants of them are used in different places.  It seems they mean the same thing but the different phrases imply there is a difference.  If there is no difference I suggest picking one term and using it consistently (ie. gridPlan -> SurveyPlan)

src/web/openlayers/layers/vector/grid-layer.ts
JRoybot, line 1: If you do go with a consistent name for grid planning/survey planning, use that name for this file and it's containing folder.

JRoybot (dismissed):
Left a few questions and suggestions but otherwise OK.  All seems to work so approved.  Address comments how you see fit.

michael-jaia:
> I find it odd that sometimes there is a row of text under yellow line in the Suvey Planner title and sometimes there isn;t. I think it would be helpful to include some text like "Enter Safety Return Parameters" when you get to that sub-panel.

I just played with adding the row of text, but I did not like how it looked for the tasking and SRP. An explicit instruction is needed for the first three pages because they involve map interactions. The remaining pages are just configuration based on the dropdown option selected.

michael-jaia:
> Will SRP every be introduced to non-survey missions?

I don't think so because it is makes the most sense with a succession of dives and if you are going to plan that the survey tool is the most efficient way.

src/web/components/Map/Map.tsx
michael-jaia: Possibly but the Map component cannot access the visible panel unless we put it in JaiaGlobal.

src/web/context/handlers/waypoint-handlers.ts
michael-jaia: Just switched to `if (action.task.getIsSurveyTask()) {`. I feel like that is a little cleaner since it is in the `handleChangeTaskParameter` function

src/web/data/survey_planner/grid-plan.ts
michael-jaia: The idea was that there can be different types of survey patterns (grids, polygons, and any others that might be desired). The foundation of patterns would live in `data/survey_planner` and a dropdown would be included in the `SurveyPlanner` component to select the pattern to be used. If it is confusing, I can make them all the same, and we can figure out a better way to express this idea.

src/web/components/Map/Map.tsx
JRoybot: This is a tsx file, why can't it access it through context?

src/web/data/survey_planner/grid-plan.ts
JRoybot: when do we anticipate the survey plan could be something other than a gird, I've heard this suggested for a few years?  Wouldn't a polygon plan still be a grid, just not necessarily rectangular shaped?

JRoybot (dismissed):

src/web/components/Map/Map.tsx
michael-jaia: Because it is only supposed to render one time to create the element that will hold the map and the click event handler. Introducing JaiaContext will cause this component to render on the polling cycle which causes an error. I forget the exact error but when I tried earlier in the project, an issue with OpenLayers arose.

PR #1415, opened 2025-11-24
task/add-configurable-data-rate-for-imu-jaia-2102

edsanville:
`jaiabot_imu.pu` now has a configurable sample rate.  The C++ driver sends a message to the python driver to configure this rate, (as well as start the flow of data).

Extra thoughts:
* Is it time to remove the BNO055 driver from the tree, since it doesn't get used and is just maintenance now?
* We could potentially reduce our number of used ports drastically:
    * We have a single "jaiabot_udp_gateway" app that listens on a designated port and takes a "UDPMessage"
    * The python sensor drivers contact this port on startup, and `jaiabot_udp_gateway` receives and publishes submessages
    * Example:

```
message UDPMessage {
    optional message IMUData
    optional message PressureData
    optional message SalinityData
    optional message SomeOtherMessageFromOutsideC++
}
```

* The `jaiabot_udp_gateway` app would then publish the messages onto the correct Goby group.

### DIFF
--- a/src/python/adafruit/jaiabot_imu.py
+++ b/src/python/adafruit/jaiabot_imu.py
@@ -18,7 +18,7 @@ parser.add_argument('-t', dest='device_type', choices=['sim', 'bno055', 'bno085'
 parser.add_argument('-p', dest='port', type=int, default=20000, help='Port to publish orientation data')
 parser.add_argument('-l', dest='logging_level', default='WARNING', type=str, help='Logging level (CRITICAL, ERROR, WARNING (default), INFO, DEBUG)')
 parser.add_argument('-i', dest='interactive', action='store_true', help='Menu-based interactive IMU tester')
-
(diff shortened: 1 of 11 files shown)


JRoybot (dismissed):
All looks good to me.  Not very familiar with this part of the code but did not see anything that looked wrong.  Built and tested branch with the simulator.

src/python/adafruit/jaiabot_imu.py
```diff
@@ -121,89 +148,95 @@ def do_interactive_loop():
...
-        print()
+                while True:
+                    current_time = datetime.datetime.now()
+                    sample_time = (current_time - start_time).total_seconds()
+                    if sample_time > sample_duration:
+                        break
+                    
+                    # Send command to take a reading
+                    imuCommand = IMUCommand()
+                    imuCommand.type = IMUCommand.TAKE_READING
```
marshalllele, line 199: I think this TAKE_READING slipped through. It should just be removed, correct?

marshalllele (approved):
Looks good! Tested on the bench as well as on the water with F6B16 and F6B19. Verified that we are able to get both pressure and IMU readings at 100 Hz at the same time without any large gaps in the data.

PR #1417, opened 2025-12-01
bug/jcc-not-being-enabled

ferrom-jaia:
## Description

We were trying to field trial spicers new bots and when we went to the JCC for hub 2 we were brought to the switchboard instead of the JCC.

The reason for the regression is how the jcc.conf gets copied to the /etc/apache2/sites-available directory. This use to happen during the installation of jaiabot-web. Now it happens during the creation of the systemd configuration files. The thought behind the change was it was a systemd configuration since we use apache2. Previously a2ensite was run in the post install script for jaiabot-web, but since initially the jcc.conf is not in the /etc/apache2/sites-available directory it had no affect. The fix is to run the a2ensite after the creation of the jcc.conf file. 

I am also removing from the post install script since we are doing it during the creation of the configuration file.

## Testing

* No tests as of yet

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -178,6 +178,7 @@ filter-template-debian-builds: &filter-template-debian-builds
     branches:
       only:
         - "2.y"
+        - "bug/jcc-not-being-enabled"
 
 # which branches to run the normal source build
 filter-template-normal-builds: &filter-template-normal-builds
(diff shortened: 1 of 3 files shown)


marshalllele (approved):
Looks good! Removed "+ service" from line 820 in systemd.py to get it working. Tested on Newly flashed SD in F14H2.

Was receiving the error "apache2jcc.conf.service does not exist."

PR #1424, opened 2025-12-03
task/create-safe-guard-for-number-of-balls

michael-jaia:
### Summary
This pull request removes dead code from app.cpp in the mission manager app.


src/bin/mission_manager/app.cpp
```diff
@@ -40,8 +40,6 @@ namespace si = boost::units::si;
 namespace zeromq = goby::zeromq;
 namespace middleware = goby::middleware;
 
-#define earthRadiusKm 6371.0
```
michael-jaia, line 43: Are you sure we want to delete this?

ferrom-jaia: Yes

PR #1431, opened 2025-12-08
task/2.y/add-back-rc-overdrive-feature/JAIA-2169

michael-jaia:
### Summary
This pull request adds back the RC overdrive toggle to prevent speed 3 from being available by default. This provides the operator with greater control while remotely operating the vehicle.

image

### Testing
1. Toggle state preserved between single and dual stick selection
2. Toggle on allows speed 3
3. Toggle off does not allow speed 3


src/web/components/__buttons__/DataOffloadButton/DataOffloadButton.tsx
```diff
@@ -50,7 +50,12 @@ export default function DataOffloadButton(props: Props) {
      * After data offload refactor, return DisabledCodes.DOWNLOAD_QUEUE if bot is already in queue
      */
     const getDisabledCode = () => {
-        if (!isCommandAvailable(CommandType.RECOVERED, props.bot.getMissionStatus().missionState)) {
+        const missionState = props.bot.getMissionStatus().missionState;
+
+        if (
+            !isCommandAvailable(CommandType.RECOVERED, missionState) &&
```
JRoybot: This is tricky to understand... assuming it means if recovery commands are not available and retru data offload are also not available, then the disabled code is due to mission state.  

Why isn't that an OR instead of an AND?  Are there situations where one of those command types is not allowed but the other is where we do not want to disable the data off load button?    If it is an AND condition, would it be cleaner to do  This OR That instead of NOT(This) AND NOT(That)?

src/web/components/__buttons__/DataOffloadButton/DataOffloadButton.tsx
```diff
@@ -88,10 +93,15 @@ export default function DataOffloadButton(props: Props) {
     const onDialogClose = (dialogAction: DialogActions) => {
         setIsDialogVisible(false);
 
+        let commandType = CommandType.RECOVERED;
```
JRoybot, line 98: why was this affected by adding a toggle to the RC panel?

src/web/components/ButtonList/ButtonList.tsx
```diff
@@ -119,7 +120,7 @@ export default function ButtonList(props: Props) {
                         handleButtonClick(ButtonTypes.PANEL, ButtonNames.JAIA_ABOUT_PANEL)
                     }
                 >
-                    <img src="/favicon.png" title="About" />
+                    <img src={JaiaOnlyLogo} title="About" />
```
JRoybot, line 123: assume this was just added to this PR because it is trivial.

JRoybot (dismissed):
Built and tested branch  a few questions, nothing to stop a merge.

src/web/components/ButtonList/ButtonList.tsx
michael-jaia: Yes

src/web/components/__buttons__/DataOffloadButton/DataOffloadButton.tsx
michael-jaia: No, I was just using this branch to implement some small changes that came up during testing yesterday

src/web/components/__buttons__/DataOffloadButton/DataOffloadButton.tsx
michael-jaia: Both the `RECOVERED` and `RETRY_DATA_OFFLOAD` commands trigger a data offload, but the `RECOVERED` command cannot be sent from a failed state because of Bot's state machine. if we remove the NOTs wouldn't we have to return `DisabledCodes.NONE` if the expression is true? Or are you thinking  if NOT(expression1 or expression 2)?

JRoybot: Yes my bad
 NOT(A) AND NOT(B) === NOT(A OR B)

JRoybot (dismissed):
Changes look good, verified in browser with simulator.  I do think that conditional could be easier to understand if changed but not a huge deal.

PR #1434, opened 2025-12-09
feature/2.y/make-task-packets-easier-to-click

michael-jaia:
### Summary
This pull request makes it easier to select features on the map. It adds a 15 px radius around each feature by setting the hitTolerance property. This pull request also fixes a bug where useMemo updated one iteration after a state change (since it is triggered by certain properties being modified in state) causing the JCC to crash because it did not have the correct task data to display in the panel

### Testing
1. Clicking through task packets tightly clustered
2. Switching between different task packet types to verify the crash has been resolved


src/web/components/TaskPacketPanel/TaskPacketPanel.tsx
```diff
@@ -101,6 +98,11 @@ export default function TaskPacketPanel(props: Props) {
 
     switch (props.selectedTaskPacket.type) {
         case MapFeatureTypes.DIVE:
+            // Handle transition between state and memo updates
```
JRoybot, line 101: This comment is hard to understand.  What condition leads to the task packet itself not matching the map feature type?  How do they get out of sync?  Is there a bug somewhere else creating this condition?

michael-jaia: Clicking the map feature causes the state update, then the state update causes the task packet update which is managed by useMemo to avoid unnecessary re-rendering

JRoybot: Got it, thanks.

PR #1435, opened 2025-12-09
task/2.y/use-portal-for-mission-set-storage-dialog

michael-jaia:
### Summary
This pull request fixes the bug where the RC panel covered the mission set storage dialog. It uses `createPortal` to render the dialog at the top level. Before it was being rendered inside the mission set panel which uses a different stacking context than the RC panel so increasing the z-index did not have an impact. Now they are rendered in the same stacking context which allows us to use z-index.

### Testing
Opened mission set storage dialog with RC panel open. Dialog renders above the RC panel.


src/web/style/stylesheets/z-index.txt
```diff
@@ -1,2 +1,3 @@
-Jaia Panels: 100
-RC Panel   : 101
+Jaia Panels : 100
+RC Panel    : 101
+Jaia Dialogs: 102
```
JRoybot: How are these used?  I could not identify any references to them, seems most places declare the z index using numeric literal `z-index: 102`

JRoybot (dismissed):
Built and ran branch using simulation, verified dialogs were on top of RC Panel.  Left one questions but otherwise approve.

src/web/style/stylesheets/z-index.txt
michael-jaia: Good call, I should use variables

PR #1436, opened 2025-12-09
task/2.y/limit-max-zoom-on-tile-layers

michael-jaia:
### Summary
This pull request sets the max zoom on the Open Street Maps and ArcGIS layers to prevent 400 errors and no tile available screens.

### Testing
Zoom the map beyond the max and see no requests are being made for new tiles (No console errors and no grey tiles)


JRoybot (approved):
Verified in browser with simulation.  Nice job.

PR #1441, opened 2025-12-11
task/turn-off-auto-apt-update-upgrade

ferrom-jaia:
## Description

Removing auto apt upgrades on image creation and on updating of jaia software.

Without this unattended upgrades would occur locking the user out of upgrading the jaia software.

Example:

```
Waiting for cache lock: Could not get lock /var/lib/dpkg/lock-frontend. It is held by process 4019 (unattended-upgr)
```

## Tests
* Tested on bot 2 fleet 1


marshalllele (approved):
Looks good! Tested on F6H4.

PR #1442, opened 2025-12-12
task/2.y/write-saved-mission-set-translator/JAIA-2170

JRoybot:
This pull request adds translation of legacy mission files into a mission set.

- Cleaned up new file format
- Added a version number to newer file format for future updates
- Lack of version number assumes it is a legacy mission file and attempts to translate it
- Added warnings for old and invalid file formats


src/web/components/MissionsPanel/MissionSetStorage/mission-set-storage.ts
```diff
@@ -168,17 +168,17 @@ export async function loadSnapshotFromFile(): Promise<LoadSnapshotResult> {
                     loadSnapshotResult.snapshot = extractMissionSetSnapshot(parsed.snapshot);
                     loadSnapshotResult.resultType = LoadResultType.CURRENT_FORMAT;
                     resolve(loadSnapshotResult);
+                    return;
```
JRoybot: I thought if the user selected a current version file and it was cleanly parsed the mission storage dialog should close (it used to).  Return here prevents the call to the parent onClose, which closes the storage dialog

michael-jaia (approved):
Nice work!

PR #1443, opened 2025-12-15
task/2.y/add-back-mission-repeats/JAIA-2172

michael-jaia:
### Summary
This pull request adds back the field to change the number of mission repeats. This changes the layout of the accordion to optimize for style and width. By moving the repeats to the top row, the two input boxes can be next to each other and we can reduce the panel width from 408px to 300px.

### Testing
Change the mission repeats and watch the Bot perform the entered number of repeats


JRoybot (approved):
Nice job on this.  Tested with simulation

PR #1451, opened 2025-12-22
task/enable-iridium-subscription-retries

marshalllele:
Enables subscription retry on a 2 1/2 minute interval. Currently, if the Bot tries to send a subscription message over Iridium and is unsuccessful, it will never retry.


ferrom-jaia:
Tested on fleet 36. Need to have a bigger discussion about this going forward.

PR #1457, opened 2026-01-09
task/cleanup-mission-manager-code-jaia-2167

edsanville:
This is a large refactor of the mission_manager code.  It separates the different state classes into their own directory/file structure, and uses a separate `states_fwd.h` function for forward declarations.


ferrom-jaia (approved):
Looks good! Tested in sim. Need to test in the field.

PR #1459, opened 2026-01-12
task/move-undo-snapshots-after-handlers/JAIA-2184

JRoybot:
This task moved the snapshotting of state/context for undo to occur after tracked actions.

The existing history buffer was reworked to be more like a generic stack and it's use was updated to work with the new snapshotting logic.

The defensive code to fix interim state changes related to survey planning was removed as it is no longer needed.

Exercised code in browser with simulator, did not identify any odd behavior

During debugging of issue created with rally points not being undone correctly it was discovered that we needed to clone snapshots when restoring to prevent corruption of the history buffer after undo.  Cloning was added to all restoreFromSnapshot functions.


michael-jaia (changes requested):
Jeff looks good. I did notice the undo button is not working the rally button right now.

src/web/data/rally_points/rally-points.ts
```diff
@@ -72,9 +72,10 @@ export class RallyPoints {
     }
 
     restoreFromSnapshot(snapshot: RallyPointsSnapshot) {
-        this.rallyPoints = snapshot.rallyPoints;
-        this.nextRallyPointID = snapshot.nextRallyPointID;
-        this.selectedRallyPointID = snapshot.selectedRallyPointID;
+        const restored = cloneDeep(snapshot);
+        this.rallyPoints = cloneDeep(restored.rallyPoints);
```
michael-jaia: Jeff do we need a double clone deep here?

michael-jaia (approved):

PR #1462, opened 2026-01-19
bug/edit-color-and-ghost-missions-out-of-sync

michael-jaia:
### Summary
During a training session last week, two minor UI bugs were found: 1) Once you send a mission, the edit toggle does not impact the color of the mission 2) If you send a mission, edit the mission (do not send again), and delete the mission, the ghost mission will show the new changes that did not get sent to the Bot. This pull request also adds back the ability to add waypoints on the depth contour map.

### Solution
1) The first bug was due to the edit mode color being set in an else statement that depended on whether the mission had started 
2) The second bug requires that we save the mission on start rather than deletion.

### Testing
Breaking cases now pass in simulation


marshalllele (approved):
Code looks good and tested fixes in sim.

PR #1475, opened 2026-02-07
bug/fix-imu-envelope

edsanville:
Fix bug: FromString() is a class method on UDPGatewayEnvelope, not an instance method.


ferrom-jaia (approved):
Nice find!

PR #1485, opened 2026-02-16
Remove lora groups that are not used

tsaubergine:
These groups aren't used, so remove them for cleanliness.


ferrom-jaia (approved):
Looks good!

PR #1492, opened 2026-02-20
feat/add-mean-and-std-to-jdv-plots

AdeX11:
### Summary
JDV plot series had no statistical summary, making it hard to assess the distribution of values in a given time window.

### Solution 
Added per-series mean (μ) and standard deviation (σ) overlays to each measurement subplot. A red dashed horizontal line marks the mean, two orange dotted horizontal lines mark ±1 std deviation, and a text annotation in the top-right corner of each subplot displays the numeric μ and σ values. All statistics are computed from the full-resolution data filtered to the current zoom window, so they update dynamically as the user zooms or pans.

### Testing
Using simulation, confirmed that the mean and std lines and annotation display correctly when enabled via the Mean and ±1 σ checkboxes in the toolbar (hidden by default), and that the values update correctly when zooming into a sub-range.

### Additional
 Corrected `layout.grid.rows` and `layout.height` to use `plots.length` instead of `data.length`, which became incorrect after adding three extra Plotly traces per series for the stat lines.


marshalllele (changes requested):
Hey @AdeX11, this looks good! I tested it in sim. I had a few notes though. 

I think we should add a section to the info dialogue box (jaiabot\src\web\jdv\client\src\ui\Plots.tsx) just to explain what exactly each line represents on the plot.

Also, making the mean/std. dev. lines togglable, so the user can decide whether they do or don't want to display the lines. Maybe the ability to toggle only mean, only std. dev., or both?

Lastly, I think making it a bit more clear how the mean and std. dev. correspond to each line - maybe make the color of the std. dev. text match the color of the +/- 1 std. dev. lines? Maybe have a box in the corner of each plot showing the mean/ std. dev., then just label the actual lines as "Mean" and "+/- 1 std. dev."?

marshalllele (dismissed):
Looks good! Mean / Std. Dev. lines are now toggleable in the JDV and naming conventions match well.

ferrom-jaia:
When you hover over a plot, the buttons that appear are squished together.

image

The current JDV has separation between the buttons and the graph:

image

ferrom-jaia:
This is also a lot of console logs

image

ferrom-jaia:
In sim this is the temperature data I have:

image

The y axis goes from 14 to 15.

But when I click on the mean:

image

It gives me zero. 

If it is not showing zeros in the plot then I should not get a zero mean.

src/web/jdv/client/src/ui/Plots.tsx
```diff
@@ -403,33 +638,96 @@ export function Plots(props: PlotsProps) {
...
+                <h2 style={{ flex: 1 }}>Information</h2>
                 <button className="plotButton" onClick={() => setIsPlotInfoDisplayed(false)}>
                     <Icon path={mdiClose} size={1} style={{ verticalAlign: "middle" }}></Icon>
                 </button>
             </div>
+
             <div className="text" style={{ bottom: "10pt" }}>
-                <h4>Plot Downsampling</h4>
+                <h3>Line Legend</h3>
+                <div style={{ display: "flex", alignItems: "center" }}>
```
JRoybot: could these styles be defined in the css?

JRoybot (dismissed):
Reviewed all the code, only a minor suggestion.  Built and tested branch using the simulator and a browser.  Was able to exercise the new features without noticing any errors or odd behavior.

src/web/jdv/client/src/ui/Plots.tsx
ferrom-jaia: Do you mind updating this pull request Jeff if you think this fix is needed?

JRoybot (approved):

PR #1494, opened 2026-02-24
feature/mission-name-taskpacket-query-2201

edsanville:
Completed:

* Added `mission_id` and `mission_name` to the `MissionPlan` message.  This is only included in the first fragment of a multi-fragment `Command`.

* Added `command_from_hub_id`, `mission_id`, and `mission_name` to the `BotStatus` and `TaskPacket` messages.

* Added querying by `mission_name` to the REST API v1.

* When running a mission from JCC, the confirmation dialog includes a text field for the `mission_name` field.  This is ugly at the moment, but we can apply a style to make it nicer.

* There is a pytest script for REST API v1 that tests for start_time querying as well as mission_name querying:
`src/web/rest_api/tests/test_taskpacket_querying.py`.  For the script to work, of course one should have some task packets in the /var/log/jaiabot/db/task_packet.db  that actually have the mission_name field set.

Not completed yet:

* From @tsaubergine 's plan document, the QUERY method is not yet implemented, nor is querying by `command_from_hub_id` or `mission_id`

* Querying task packets by `mission_name` has not been added to API v0, or added to the JCC GUI.


michael-jaia:
### Simulation Testing:

I first ran a beach survey mission called: "FirstBeachNewportSurvey-2026-03-11" with 26 dives per Bot. The test script successfully found 26 task packets for Bot 1.

Next, I ran a second beach survey called "Belmont Beach Survey". At first I did not think the naming worked correctly because when I ran the test script, it said "Not enough unique mission names found in task packets to test multiple mission name filtering". I looked at the script, and I saw multiple mission filtering requires more than 2 missions because the pop on the set of mission names occurs first.

As a result, I replayed the mission set with the name "Hello", and I only saw "Belmont Beach Survey" and "Hello" in the test script output. This made me think we lost "FirstBeachNewportSurvey-2026-03-11" task packets. However, the script puts the mission set names into a set (no strict order) and checks the last two items. As I ran the script multiple times, I saw task packets from all 3 mission sets appear. I also printed out the names of the mission sets in the script to be sure it did not lose any data.

src/bin/hub_manager/hub_manager.cpp
```diff
@@ -755,8 +761,22 @@ void jaiabot::apps::HubManager::handle_task_packet(const jaiabot::protobuf::Task
             std::make_pair(task_packet.bot_id(), task_packet.start_time()));
     }
 
+    // Set the mission_name of the task packet based on the current mission id to name mapping for logging purposes
+    jaiabot::protobuf::TaskPacket task_packet_copy = task_packet;
```
michael-jaia, line 841: Why do we need to make a copy of the task_packet?

michael-jaia (commented):
Looks good, one small question:

src/bin/hub_manager/hub_manager.cpp
edsanville: It's because the task_packet variable is a const parameter.

src/web/rest_api/test/test_taskpacket_querying.py
```diff
@@ -0,0 +1,128 @@
...
+        task_packets=rest_api.TaskPacketsRequest()
+    )
+
+    response = run_request(request)
+    return response.task_packets
+
+
+all_task_packets = get_all_task_packets_for_bot(1).packets
+print(f"Retrieved {len(all_task_packets)} total task packets for bot 1")
+assert len(all_task_packets) > 0, "Expected at least one task packet for bot 1"
```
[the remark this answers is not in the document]
edsanville: We could definitely use a pytest fixture here and start the REST API for the tests.  That would require formalizing these tests and refactoring them all a bit, but definitely worth doing in a future PR.

src/web/components/__buttons__/StartAllMissionsButton/StartAllMissionsButton.tsx
```diff
@@ -109,16 +109,19 @@ export default function StartAllMissionsButton(props: Props) {
...
-    const onDialogClose = async (dialogAction: DialogActions) => {
+    const onDialogClose = async (dialogAction: DialogActions, missionName: string) => {
         setIsDialogVisible(false);
 
         if (dialogAction === DialogActions.CONFIRMED) {
             for (const botID of botReadyStates.get(DisabledCodes.NONE)) {
                 const missionID = missionsManager.getMissionID(botID);
+                const mission_plan = props.missions.get(missionID).packageMissionForHub();
+                mission_plan.mission_name = missionName; // Set the mission name from the dialog
+
```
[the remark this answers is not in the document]
edsanville: We want to match the protobuf message convention, which is snake_case.  Otherwise, life gets hard.

src/python/pyjaia/src/pyjaia/task_packet_database.py
```diff
@@ -128,6 +160,15 @@ def _add_task_packet(self, task_packet: Dict):
...
         if 'mission_name' in task_packet:
             self.db.execute('insert or replace into mission_name (id, mission_name) values (?, ?)', (id, task_packet["mission_name"]))
+        elif 'mission_command_time' in task_packet:
+            # It's possible we know about this task packet's mission_name, but a power cycle occurred.  So we can retrieve it from the
+            # mission_commands table based on the bot_id and the task_packet's mission_command_time.
+            mission_command_time_dccl = dccl_time_round(task_packet['mission_command_time'])
+            rows = self.db.execute('select mission_name from mission_commands where bot_id = ? and mission_command_time = ?', (task_packet["bot_id"], mission_command_time_dccl))
+            row = rows.fetchone()
+            if row is not None:
+                mission_name = row[0]
```
michael-jaia, line 171: Does this add mission name to the task packet?

edsanville: The next line after that will add it to the `mission_name` table, so when that task packet is next pulled from the database it will have the correct mission_name.

```self.db.execute('insert or replace into mission_name (id, mission_name) values (?, ?)', (id, mission_name))```

To test this code path, the `hub_manager` app will need to be restarted after submitting a mission.  This can be done by power cycling the hub.  After the power cycle, the TaskPacket messages won't have the mission_name field set by `hub_manager`, but it will be added by the TaskPacketDatabase when _add_task_packet() is called.

PR #1496, opened 2026-03-03
bug/loading-screen-memory-leak

JRoybot:
Investigated total page size of JCC memory usage increases on page reloads.  Initial report was rapid reloads saw the memory usage climb by approximately 100MB (viewed on browser tab, while being served from local machine in development mode)

Analysis

- memory use increase  seemed to be about 70MB per load when tested but was not exactly the same
- even though flask compress was enabled compression was not being performed
- triggering garbage collection in dev tools immediately reduced app size
- memory use of the app tended to decrease over time if left alone
- application size viewed in dev tools did not grow on reloads
- memory use viewed by hovering on browser tab includes heap and stack space

Conclusion

- javascript was not releasing temporary array used to buffer app download
- javascript garbage collection is low priority and takes time
- will always be a short term memory spike when reloading due to how Chrome buffers reloads
- production image is much smaller resulting in less noticeable spikes
- garbage collection eventually frees memory from stale java code

Changes

- immediately free up chunks array to reduce memory spike (halves memory spike)
- configure mime types to enable  flask compression (reduces bytes transferred)


ferrom-jaia:
The fix was not accumulating chunks. The original code was doing this:
```
javascriptconst chunks = [];
// inside pump():
chunks.push(result.value); 
```
This meant the entire client.js was held in memory as raw byte buffers, plus the blob created from them, plus the parsed JavaScript
The fix was to count the bytes for progress tracking but immediately discard each chunk:
```
javascript// inside pump():
loaded += result.value.byteLength; // just count it, don't store it
```
Then using response.body.tee() to split the stream so one side handles progress counting while the other is passed to new Response(blobStream).blob() — letting the browser handle the buffering internally rather than us building a chunks[] array ourselves.

ferrom-jaia:
> The fix was not accumulating chunks. The original code was doing this:
> 
> ```
> javascriptconst chunks = [];
> // inside pump():
> chunks.push(result.value); 
> ```
> 
> This meant the entire client.js was held in memory as raw byte buffers, plus the blob created from them, plus the parsed JavaScript The fix was to count the bytes for progress tracking but immediately discard each chunk:
> 
> ```
> javascript// inside pump():
> loaded += result.value.byteLength; // just count it, don't store it
> ```
> 
> Then using response.body.tee() to split the stream so one side handles progress counting while the other is passed to new Response(blobStream).blob() — letting the browser handle the buffering internally rather than us building a chunks[] array ourselves.

After merging in, aame issue, but did not show up during the testing until I tried after merging in the code. For now I remove the loading code and just have the svg pulsing to indicate loading.

PR #1497, opened 2026-03-03
bug/no-tap-to-move-on-contour-map

michael-jaia:
### Summary
Previously, when putting a waypoint in tap to move, if you clicked on an OpenLayers feature, the waypoint would not move and potentially another panel would open. Now, tap to move takes precedence over feature clicks.

### Testing
Added a waypoint to the map, toggled tap to move, and clicked OpenLayers features. The waypoint moved as expected. Once tap to move is turned off, features can be selected.


ferrom-jaia (approved):
Looks great! Tested in sim.

PR #1500, opened 2026-03-03
bug/zero-commands-sent-during-rc-dive

michael-jaia:
### Summary
Hub was sending engineering commands that set the motor and rudder to 0 while the Bot attempted to RC dive. The `useEffect` function did not have the `controlType` dependency so it did not update the value in `handleRCInterval`.

### Testing
The web portal does not receive engineering commands from the client while the RC dive panel is open.


ferrom-jaia (approved):
Looks great! Twomey walked me through this fix.

PR #1501, opened 2026-03-03
bug/tsys02-sensors-not-detected

marshalllele:
The systemd.py section which created the python driver for our Celsius fast-response temperature sensor was removed by mistake. This PR adds it back, without the C++ side (which is now being handled inside of the UDP app).

Tested on an internal BIO and PAM .


ferrom-jaia (approved):
Looks good!

PR #1504, opened 2026-03-04
bug/permissions-issue-in-jaiabot-embedded

marshalllele:
Update permissions to make jaiabot-embedded.config and .postinst executable


ferrom-jaia (approved):
Looks good! Works when configuring a new fleet now.

PR #1505, opened 2026-03-04
task/remove-wifi-link-quality-requirement

michael-jaia:
### Summary
Based on today's water testing, this pull request includes the following changes:
1. Sets the WiFi quality threshold to > 0 because we have safe guards in place in the offload script.
2. Stop using `useMemo` for `botCheckedStates`. The purpose was to prevent unnecessary reinitializations of this map, but it seemed to be the cause of a bug where the checked state on the screen did not align with the data in the map.
3. Remove the check for `fullscreenElement` because it appears escaping full screen for a download does not update this value.


ferrom-jaia (dismissed):
Looks good! Static review.

PR #1509, opened 2026-03-05
bug/initialize-first-depth-value-ctd

michael-jaia:
### Summary
This pull request initializes the depth value in the first CTD snapshot to the current depth so we do not need to check the second snapshot for the first sensor reading. Previously, we checked the second reading because the pressure adjusted lags behind raw pressure by one iteration.

### Testing
1. Logged the initial value and conducted multiple dives in simulation, viewing the UNB file each time.


ferrom-jaia (approved):
Looks good! Tested on tablet.

PR #1510, opened 2026-03-06
bug/depth-achieved-reported-incorrectly

marshalllele:
This bug was discovered when a Bot dove with 1.5 m intervals and a max depth of 10m. The bot reached the seafloor at ~6.15 m, began its hold, but almost immediately rose to the surface while still in the hold state (we assume because of seaweed). The Bot was not on the bottom long enough to reach its bottoming_timeout config, so when its hold was complete it attempted to dive down to its next increment of 7.5 m. Since the prop was fouled, it was not able to dive and instead detected the "bottom" while sitting on the surface, entered the hold state at the surface, and set its depth_achieved for that dive to 0.8 m (at the surface), even though it had actually reached a depth of 6.15 m. 

Fix was tested in sim by manually stopping the motor when we hit the seafloor (simulate sucking in seaweed) and adding a natural ascent, causing the simulated bot to accurately represent what happened in the real world.

Simulated data with old code - Depth_achieved is set to our surface depth, not the actual depth the bot reached:
Screenshot 2026-03-06 095010

Simulated data with fix - Depth_achieved is set to the deepest depth our bot reached:
Screenshot 2026-03-06 095519

### DIFF
--- a/src/bin/mission_manager/states/inmission/underway/task/dive/powered_descent.h
+++ b/src/bin/mission_manager/states/inmission/underway/task/dive/powered_descent.h
@@ -194,11 +194,11 @@ struct PoweredDescent
                     context<Dive>().dive_packet().set_depth_achieved_with_units(ev.depth);
                 }
 
-                // Set the max_acceration
+                // Set the max_acceleration
                 context<Dive>().dive_packet().set_max_acceleration_with_units(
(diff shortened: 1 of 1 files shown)


ferrom-jaia (approved):
Looks good to me! Static review.

PR #1511, opened 2026-03-06
bug/repeats-break-when-mission-deleted

michael-jaia:
### Summary
The mission repeats displayed in the Bot details panel went undefined when an active mission was deleted. Now we check to see if a ghost mission exists for that Bot, and if it does, we use that.

### Testing
In simulation, run a bot with repeats greater than 1. After you see 2 of x in the Bot details panel, delete the mission and you will continue to see the total number of repeats .


ferrom-jaia (approved):
Looks good! Static review.

PR #1514, opened 2026-03-11
task/kmz-csv-and-contour-map-endpoints-v1-rest-api-jaia-2211

edsanville:
The task_packets endpoint now has a `format` parameter, which can be one of:

```    
enum Format {
        JSON = 0 [(jaia.ev).rest_api = { presence: GUARANTEED }];
        KMZ = 1 [(jaia.ev).rest_api = { presence: GUARANTEED }];
        CSV = 2 [(jaia.ev).rest_api = { presence: GUARANTEED }];
        GEOJSON_CONTOURS = 3 [(jaia.ev).rest_api = { presence: GUARANTEED }];
}
```


src/web/rest_api/v1/api.py
```diff
@@ -8,6 +9,7 @@
 import common.shared_data
 from common.time import utc_now_microseconds
 from common.api_exception import APIException
+from jaiabot.messages.rest_api_pb2 import APIRequest, APIResponse
```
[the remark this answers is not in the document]
edsanville: I went the other way and made them all `APIResponse`.

src/web/rest_api/app.py
```diff
@@ -123,7 +123,24 @@ def jaia_api_short(version):
         if not check_api_key(jaia_request.api_key, jaia_request.WhichOneof("action")):
             abort(403) # forbidden
 
-        jaia_response.CopyFrom(process_request(version, jaia_request))
+        response = process_request(version, jaia_request)
```
michael-jaia, line 126: Why do we not always return as is? what is the benefit of `jaia_response.CopyFrom(response)`

src/python/pyjaia/src/pyjaia/csv.py
```diff
@@ -0,0 +1,61 @@
...
+        value = input.get(key)
+        if value is None:
+            continue
+
+        if key.endswith('_time'):
+            output[prefix + key] = utime_to_string(int(value))
+        elif isinstance(value, (int, float, str)):
+            output[prefix + key] = value
+        elif isinstance(value, bool):
+            output[prefix + key] = 'YES' if value else 'NO'
```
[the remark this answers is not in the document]
edsanville: Huh... bool is a subclass of int in python.  I did not know that.

src/lib/messages/rest_api.proto
```diff
@@ -18,10 +18,24 @@ message TaskPacketQuery
...
-    repeated string mission_name = 3
+    repeated string mission_name = 4
         [(jaia.field).rest_api = {
             doc: "If set, only return task packets associated with the given mission name(s)."
         }];
+    optional bool included_only = 3
+        [default = true, (jaia.field).rest_api = {
+            presence: GUARANTEED,
+            doc: "Whether to only return included (non-excluded) task packets."
+        }];
```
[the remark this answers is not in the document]
edsanville: This message is not for protobuf serialization anyway, (for now).

src/python/pyjaia/src/pyjaia/jaialog_store/jaia_h5.py
```diff
@@ -20,6 +18,11 @@
 from jaiabot.messages.jaia_dccl_pb2 import *
 from jaiabot.messages.metadata_pb2 import *
 
+# Jaia packages
+from .objects import jaialog_get_object_list
+from . import path_descriptors
```
michael-jaia, line 21: Should these follow the same pattern? Either `.objects` and `.path_descriptors` or `from .`

src/python/pyjaia/src/pyjaia/csv.py
```diff
@@ -0,0 +1,60 @@
...
+
+def task_packets_to_csv(task_packets: Iterable[TaskPacket]):
+    csv_string_io = io.StringIO(newline='') # Use newline='' as recommended for the csv module
+
+    packet_dicts = [flatten_dict(MessageToDict(packet, preserving_proto_field_name=True)) for packet in task_packets]
+    if len(packet_dicts) == 0:
+        return "No task packets."
+
+    # Get all field names
+    union_dict = {}
```
michael-jaia, line 67: I'm not totally following the purpose of the `union_dict` here

src/python/pyjaia/src/pyjaia/jaialog_store/jaia_h5.py
edsanville: Yeah, that sounds reasonable.  I will change them both to "from . import ___", so we get them namespaced.

src/python/pyjaia/src/pyjaia/csv.py
edsanville: It runs through all of the task packets, and gets the union of the field names in each one.  I do this because one task packet might be missing `.field1`, and another might be missing `.field2`, and we want to have a column for every field that's in at least one of the packets.

I used a dict instead of a set, because in modern python dicts are ordered and sets are not, (and I wanted to keep certain field (column) names first before the others, like the time and location).

Excellent question, and I think it deserves a comment in the code, which I will now add.

ferrom-jaia (approved):
Approve! Tested all new endpoints and got the expected result.

PR #1521, opened 2026-03-20
bug/quick fix for disk space on CloudHub - don't log with goby_logger

tsaubergine:
After a while the .goby log files the disk space on CloudHub, causing the machine to hang.

We can consider some more elegant fixes, but this at least keeps the CloudHub up and running.

Possible future fixes:
- Auto log rotation / deletion policy for CloudHub logs
- Ability to run logger only when CloudHub is in active use

## Testing

Tested by installing .deb on chf1.


ferrom-jaia (approved):
Static Review looks good!

PR #1524, opened 2026-03-25
feature/software-update-notification

michael-jaia:
### Summary
This pull request creates a notification dot in the Jaia button when there is no Internet connection and when there is a software upgrade available.

### Implementation
1. Move the metadata into `jaiaGlobal`
2. Extract the polling logic into independent functions
3. Query GitHub for latest release info
4. If we cannot query GitHub show no Internet notification
5. If we can query GitHub, determine if an upgrade is available and show the correct notification (none if no upgrade available)

### Testing
1. Turn off Internet, see yellow dot and message
2. Turn on Internet, see no dot
3. Modify version from metadata, see red dot and upgrade available message

### Notes
1. The upgrade available link will not work in sim


src/web/jcc/polling.ts
```diff
@@ -0,0 +1,328 @@
...
+
+    if (parseInt(gitHubVersion.major) > parseInt(currentVersion.major)) {
+        return true;
+    }
+
+    if (parseInt(gitHubVersion.minor) > parseInt(currentVersion.minor)) {
+        return true;
+    }
+
+    if (parseInt(gitHubVersion.patch) > parseInt(currentVersion.patch[0])) {
```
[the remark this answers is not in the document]
michael-jaia: @copilot  The GitHub version will always be greater than or equal to current version

src/web/components/JaiaAbout/JaiaAbout.tsx
```diff
@@ -1,42 +1,55 @@
...
         if (!version) {
-            return;
+            return "---";
         }
 
         if (version.major && version.minor && version.patch) {
-            setVersion(`${version.major}.${version.minor}.${version.patch}`);
+            return `${version.major}.${version.minor}.${version.patch}`;
+        }
+    };
```
JRoybot, line 42: agree with copilot, what happens if the version exists but is missing a field?

src/web/components/JaiaAbout/JaiaAbout.tsx
JRoybot: This points to documentation for 1.20.4, suggest replacing with https://docs.jaia.tech/ to get current docs

src/web/components/JaiaAbout/JaiaAbout.tsx
```diff
@@ -55,6 +68,13 @@ export default function JaiaAbout() {
...
+            <div className="jaia-about-row">
+                <div className="label">Notifications:</div>
+                <div className="input notification">
+                    <NotificationDot className="jaia-about-panel" />
+                    {getNotificationDotHelperElement()}
+                </div>
+            </div>
             <div className="jaia-about-row">
                 <div className="label">Website:</div>
                 <a href="https://www.jaia.tech" target="_blank" rel="noopener noreferrer">
```
JRoybot: Suggest putting the company information in constants instead of embedded in the HTML

src/web/data/jaia_global/jaia-global.ts
```diff
@@ -7,14 +7,15 @@ import {
...
+import { Metadata, Version } from "../../types/protobuf-types";
 import { UNASSIGNED_ID } from "../../utils/constants";
 
 export interface JaiaGlobalSnapshot {
     selectedNode: SelectedNode;
     selectedWaypoint: SelectedWaypoint;
     selectedTaskPacket: SelectedTaskPacket;
-    mapMode: MapModes;
     defaultTaskParameters: TaskParameters;
+    mapMode: MapModes;
```
JRoybot, line 17: Not sure why this was moved, was original intent was to keep it close to other user UI settings.  Either way, should use the same ordering on line 50, 180, 1901 below.

src/web/utils/constants.ts
```diff
@@ -7,7 +7,8 @@ export const TEXT_OFFSET_RADIUS = 11;
 export const NO_COMMS_STATUS_AGE = 30; // seconds
 export const DATA_MODEL_POLL_TIME = 500; // milliseconds
 export const TASK_PACKET_POLL_TIME = 1000; // milliseconds
-export const METADATA_POLL_TIME = 10000; // milliseconds
+export const METADATA_POLL_TIME = 10_000; // milliseconds
```
JRoybot, line 10: Curious why we need to poll the metadata and github versions, unlikely they change during a single exercise and if they did, does it really matter?  Do we need to tell the user there is a new version available in the middle of an exercise?

JRoybot (commented):
Looks good, left a few picky comments.

src/web/utils/constants.ts
michael-jaia: I think we need the metadata poll for the case when the JCC server is running but the Hub code is not, so when the Hub code starts, we will get the metadata on the JCC side.

Similar with the GitHub poll, if someone starts without Internet, and then connects to Internet, we will want the notification to appear.

src/web/data/jaia_global/jaia-global.ts
michael-jaia: Good catch, I will restore that change

src/web/utils/constants.ts
JRoybot: Got it.  Could cancel the polling timers for those once you get a response.

JRoybot (dismissed):

ferrom-jaia (approved):
Looks good!

Tested in sim with Twomey.

PR #1508, opened 2026-03-05
task/update-web-dependencies/JAIA-2206

JRoybot:
Our NodeJS packages had not been updated in a while and we had several critical updates needed including several major revisions to packages.

- Used `npm outdated` to check status of our node packages, before and after status is listed below.
- Updates were made incrementally starting with major version changes and fixing any impacted code as needed.  
- Cleaned up dependencies by eliminating unused ones and replacing beta and unsupported packages
- Tested by building locally and running JCC anfd JDV in browsers and pushing each build to github for CircleCI to check.

**Notes:**

- Jest and related packages not update to 30 because ts-jest is not updated to work with 30 yet.  We are going to jest 29 and keeping babel-jest and other related packages in sync with jest 29
- Node js 25 is released but it is not an LTS build so we are upgrading from 22 to 24
-  jest-environment-jsdom 29 depends on jsdom ^20.0.0 — it bundles its own jsdom v20 internally
    - The only way to get a newer jsdom in tests is to upgrade to jest 30 / jest-environment-jsdom 30 (which we're deferring)
    - jsdom specified in package list was not used and was removed
-  pushed @types/proj4 to the latest build but still getting 2.5.6 because maintainers have not identified anything newer as stable

 `npm outdated`  output

**Before**
```
Package                                               Current    Wanted   Latest  Location                                                          Depended by
@babel/core                                            7.27.4    7.29.0   7.29.0  node_modules/@babel/core                                          web
@babel/plugin-transform-class-properties               7.27.1    7.28.6   7.28.6  node_modules/@babel/plugin-transform-class-properties             web
@babel/plugin-transform-nullish-coalescing-operator    7.27.1    7.28.6   7.28.6  node_modules/@babel/plugin-transform-nullish-coalescing-operator  web
@babel/plugin-transform-optional-chaining              7.27.1    7.28.6   7.28.6  node_modules/@babel/plugin-transform-optional-chaining            web
@babel/preset-env                                      7.27.2    7.29.0   7.29.0  node_modules/@babel/preset-env                                    web
@babel/preset-react                                    7.27.1    7.28.5   7.28.5  node_modules/@babel/preset-react                                  web
@emotion/styled                                       11.14.0   11.14.1  11.14.1  node_modules/@emotion/styled                                      web
@fortawesome/fontawesome-svg-core                       6.7.2     6.7.2    7.2.0  node_modules/@fortawesome/fontawesome-svg-core                    web
[...]

PR #1528, opened 2026-03-31
task/jcc-reduce-size-of-clientjs/SW-2207

JRoybot:
Looked for ways to further improve JCC load times.  This is a follow on to PR #1508  Investigated how larger modules are included and looked for ways to streamline the load times.  Webpack tree-shaking already does a good job trimming what gets sent in the client bundle for most modules.  Identified the plotly library accounted for more than 50% of the total bundle.  It is only used in JCC for 3d depth plots.

Ruled out "lazy loading" techniques to prevent users having issues if they launch JCC and subsequently lose connectivity to the server.   Implemented asynchronous dynamic loading so the plotly module is loaded immediately after JCC initially renders.

This cuts the initial load time by more than half.  Switching from CommonJS to ESNext also reduced the bundle size as it improves webpack's ability to reduce bundle size by "tree-shakling".  client.js reduced from 3.34 MB to 804 kb

**Before changes**
client.js **3.34 MB** (including plotly)
**After changes**
client.js 804 kb
plotly.js 1.3MB
Total **2.1 MB**

Tested on development platform and used Chrome developer tools to measure sizes and load times.  Tested with webpack in development and production mode and with network speeds throttled to 3G

Should be tested in production environment with real hub before merging.

**Summary of Changes**

`src/web/tsconfig.json`
Changed module from CommonJS to ESNext.   With CommonJS, TypeScript converts all import() calls to require() before webpack sees them, which prevents code splitting. ESNext tells TypeScript to leave import() as native dynamic imports that webpack can split into separate chunks.  This allowed plotly to be dynamically loaded.  This also improves tree-shaking across the entire bundle.

`src/web/tsconfig.test.json (new file)`
Jest runs in a CommonJS environment and can't handle ESNext modules directly. This file extends the main tsconfig but overrides module back to CommonJS so tests continue to work.

`src/web/jest.config.js`
Points ts-jest at tsconfig.test.json so Jest uses the CommonJS module setting.

`src/web/jcc/index.tsx`
Added a dynamic import("plotly.js-dist") after root.render() to eagerly load the Plotly chunk immediately on app startup, before the user navigates to the DepthMap panel. Also updated module.hot.accept() to import.meta.webpackHot?.accept() which is the ESNext equivalent.

`src/web/components/DepthMap3D/depth-map-3D.ts`
Removed the static require("plotly.js-dist") at the top of the file and replaced it with a dynamic import() inside buildDepthMap(). Made buildDepthMap() async to support the await.

`src/web/components/DepthMap3D/DepthMap3D.tsx`
Removed the static require("plotly.js-dist") and changed handleClick to async, using a dynamic import() for the Plotly.purge() call.

`src/web/jdv/client/src/ui/LoadProfile.tsx`
Fixed a TypeScript import X = require() syntax that is not valid in ESNext module mode. Changed to standard import X from syntax.

**Potential Follow on Work**
Add a toast notification after plotly loads telling the user it is OK to move out of wifi range


JRoybot (commented):
Address Copilot comments

edsanville:
I love it!

For reference, here are the sizes of the dev and release `client.js` files (before and after this PR):

Before:
```
-rw-rw-r-- 1 ed ed 11765295 Apr  2 08:46 ./amd64/share/jaiabot/web/jcc/client.js
-rw-rw-r-- 1 ed ed 69504060 Apr  2 08:46 ./web_dev/jcc/client.js
```

After:
```
-rw-rw-r-- 1 ed ed  2404456 Apr  2 08:39 ./amd64/share/jaiabot/web/jcc/client.js
-rw-rw-r-- 1 ed ed 27501923 Apr  2 08:39 ./web_dev/jcc/client.js
```

edsanville (approved):

ferrom-jaia (approved):
Looks great. Tested in sim.

PR #1546, opened 2026-04-17
feature/jcc-string-mission-together/sw-2220

JRoybot:
This PR adds the capability to chain a bunch of mission sets together to create multi survey missions and other complex missions.

The design focused on minimizing impact to existing functionality, minimize the introduction of new terms, and provide the capability to the customers as quickly as possible.

The concept of a Segment was introduced to the MIssionPlan proto message and the Mission class in the JCC data model.  A Segment is used to assign and manage transit speeds and SRPs to a subset of goals in the MissionPlan.  Other minor adjustments were made to the JCC Mission and MissionSet classes to accommodate handling of speed data in Segments.

The general workflow to utilize this feature is; the user will create mission sets with the elements of the desired combined mission set and save each with a meaningful name.  Once all the elements are saved they open the Mission Set Editor Panel to combine them.  The combined mission sets do not need to have the same number of missions in them.  The code will copy/distribute missions as needed.  The user can also specify a desired number of missions.  Currently nothing is displayed on the map until the combined Mission Set is loaded.  Map interaction can be added as a follow on task.  NOTE:  it is useful to place rally points on the map at the beginning and end of individual mission sets when creating the element missions so the user has visual reference of the other element missions after they clear the mission panel to create a new element.

The following design constraints were used in the development of this feature.

- Existing MissionPlans without segments are still supported (support REST API)
- Due to limitations of the existing Mission Panel and editing tools transit speed is set to the maximum value of all the segments of a combined Mission Set.
- Remove existing "hack" to detect lane ends in surveys when using SRPs
- Provide migration tools of existing Mission Sets saved local storage and exported to external file systems
- Combined Missions Set will use the largest number or missions from the elements
- If an element Mission Set has fewer than the max missions, missions repeated to fill slots
- Make all Load and Import functions auto-assign bots to missions

Development of this feature included an initial cut of the JCC UI to collect feedback.  After collecting feedback the UI was modified to better meet expectations and to support desired behaviors when executing the combined missions.  Once the JCC UI to combine mission sets was stable the mission_manager was updated to support MissionPlans with segments.

This functionality was testing in a browser using the simulator.

Several other issues and gaps in the existing code base were were discovered and resolved as part of this PR.

- src/web/custom.d.ts modified to support imports of .less files in Typescript
- implement general migration tools to update Mission Sets and add translator to migrate v2.0 to v2.1,
- made warning dialogs in Import and Load functions more consistent
- fixed bug in Import and Load functions that corrupted mission/bot assignments

Claude Code was used to develop most of this code and was refined through review and testing.

Follow on work

- Currently there is no visual display of component missions sets or combined missions set  when editing them.  A layer could be added to the map to display them.
- The speed sliders are currently at the Mission Set level.  We track the value of them in the Mission Set so we can handle cases with no missions.   Speeds and SRPs should be assignable at the mission level, ideally to a subset of goals.  
- If desired we could modify the Missions Panel to allow more than one Mission Set to be loaded at a time and put functionality to combine them there.
- At some point we probably need to rework the entire mission editing process to allow a more freeform creation of complex mission sets using a pallet of elements like waypoints, rally points and surveys etc

**How to use the new feature**
image

Combining Mission Sets-Page-1 drawio

Combining Mission Sets-Page-3 drawio

image


src/bin/mission_manager/states/inmission.h
```diff
@@ -36,6 +36,13 @@ struct InMission
     InMission(typename StateBase::my_context c) : StateBase(c)
     {
         goby::glog.is_debug1() && goby::glog << "InMission" << std::endl;
+
+        apply_segment_params(goal_index_);
+
+        jaiabot::protobuf::PPKCommand ppk_command;
```
michael-jaia: Do we need PPK changes?

src/web/components/MissionsPanel/MissionSetEditor/SaveAndLoadButton/SaveAndLoadButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { saveSnapshotToLocalStorage } from "../../MissionSetStorage/mission-set-storage";
+import { combineMissionSets } from "../mission-set-editor";
+import { MissionSetSnapshot } from "../../../../data/mission_set/mission-set";
+import { DisabledCodes } from "./save-and-load-messages";
+import { SaveAndLoadDialog } from "./SaveAndLoadDialog";
+
+interface Props {
+    editorName: string;
+    desiredMissionCount: number;
+    rightList: string[];
```
michael-jaia: Should we call this combinedMissionNames so it is not dependent on the layout that could change

src/web/components/MissionsPanel/MissionSetEditor/SaveAndLoadButton/SaveAndLoadButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+        if (dialogAction === DialogActions.CONFIRMED) {
+            const name = props.editorName.trim();
+            const snapshot = combineMissionSets(
+                props.rightList,
+                props.desiredMissionCount,
+                name,
+                props.snapshotCache,
+            );
+            saveSnapshotToLocalStorage(name, snapshot);
+            if (jaiaDispatch) {
```
michael-jaia: We can remove this check, otherwise I think it is a little inconsistent to check it here, but no where else in the code

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
```diff
@@ -0,0 +1,236 @@
...
+ * still receives a contribution (e.g. a single transit mission repeats for all bots).
+ *
+ * @param {Mission[]} missions Source missions from a single mission set
+ * @param {number} slotCount Number of output slots (desiredMissionCount)
+ * @returns {Mission[][]} Array of length slotCount, each element being the source missions for that slot
+ */
+function distributeMissionsToSlots(missions: Mission[], slotCount: number): Mission[][] {
+    // slotCount >= 1 is guaranteed by callers; both getMaxWaypointsPerOutputMission and
+    // combineMissionSets enforce desiredCount >= 1 before reaching here.
+    const result: Mission[][] = Array.from({ length: slotCount }, (): Mission[] => []);
```
michael-jaia: Would it be simpler to say: `const result: Mission[][] = []`?

michael-jaia:
Hi Jeff, I found a way for a Bot to have more than 80 waypoints. In this picture 1 Bot has 148. I did this by creating 2 5-lane surveys with under 80 points for 5 Bots, but when I combine them again and set the number of Bots to 3, I do not get a warning. I thought about trying this because as I was reading the slots code. It is a tricky problem combining missions designed for different numbers of Bots. 


image

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
```diff
@@ -0,0 +1,236 @@
...
+ * Returns the maximum waypoint count across all output missions that would result
+ * from combining the given mission sets with the specified output count.
+ * Used to validate against MAX_WAYPOINTS before saving.
+ *
+ * @param {string[]} names Ordered list of saved mission set names
+ * @param {number} desiredCount Number of output missions
+ * @param {Map<string, MissionSetSnapshot>} snapshotCache Cache of loaded snapshots
+ * @returns {number} Maximum waypoints in any single output mission
+ */
+export function getMaxWaypointsPerOutputMission(
```
michael-jaia, line 33: I could be wrong here, but I feel like this function is doing a lot. Based on the name, I was thinking it would look something like
```
let totalWaypoints = 0;
for (const mission of combinedMissions) {
    totalWaypoints += mission.getWaypoints().length;
}
```

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
```diff
@@ -0,0 +1,236 @@
...
+            0,
+        );
+        maxWaypoints = Math.max(maxWaypoints, slotTotal);
+    }
+
+    return maxWaypoints;
+}
+
+/**
+ * Combines multiple saved mission sets into a single new mission set snapshot.
```
michael-jaia, line 89: I'm wondering if it would be clearer to use a word other than snapshot for stringing missions together? It makes me think about the undo button structure but it has a different meaning here.

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
```diff
@@ -0,0 +1,236 @@
...
+ * carrying the max transit speed from all source missions. Additional segments are
+ * appended only when a source mission contributes bottom_depth_safety_params.
+ *
+ * @param {string[]} names Ordered list of saved mission set names to combine
+ * @param {number} desiredCount Number of missions in the resulting set
+ * @param {string} newName Name for the new combined mission set
+ * @param {Map<string, MissionSetSnapshot>} snapshotCache Cache of loaded snapshots
+ * @returns {MissionSetSnapshot} Snapshot ready to save and/or load
+ */
+export function combineMissionSets(
```
michael-jaia, line 96: I'm wondering if there is a way to break up this 100+ line function into smaller components for modularity and readability

src/web/components/MissionsPanel/MissionSetEditor/MissionSetEditor.less
michael-jaia, line 1: As a convention we should use px values that are multiples of 3 (3px system) for consistency around the stylesheets. For example, if you see something that is 8px, bump it to 9px or drop it to 6px...16 px would go to 15px...

src/web/components/MissionsPanel/MissionSetEditor/MissionSetEditorButton.tsx
```diff
@@ -0,0 +1,26 @@
...
+    return (
+        <div>
+            <Button
+                className="jaia-button"
+                aria-label="mission-set-editor"
+                onClick={() => setIsDialogVisible(true)}
+            >
+                <Icon path={mdiSetMerge} size={MDI_BUTTON_SIZE} title="Mission Set Editor" />
+            </Button>
+            {isDialogVisible && (
```
michael-jaia: Same functionality but we are getting away a little bit from the original structure. For example, in the activate button we have this...where the dialog handles the display value
```
            <ActivateDialog
                isVisible={isDialogVisible}
                disabledCode={getDisabledCode()}
                onClose={onDialogClose}
            />
```

src/web/components/MissionsPanel/MissionSetEditor/MissionSetEditorDialog.tsx
```diff
@@ -0,0 +1,309 @@
...
+import { formatNumericalInput } from "../../../utils/input";
+import SaveAndLoadButton from "./SaveAndLoadButton/SaveAndLoadButton";
+
+import "./MissionSetEditor.less";
+
+interface DialogProps {
+    onClose: () => void;
+}
+
+interface LeftListItemProps {
```
michael-jaia: Similar to the comment above, I'm wondering if we should drop Left and Right too future proof this feature

src/web/components/MissionsPanel/MissionSetEditor/MissionSetEditorDialog.tsx
```diff
@@ -0,0 +1,309 @@
...
+
+    const handleLeftItemClick = (index: number) => {
+        setSelectedLeftIndex((prev) => (prev === index ? null : index));
+    };
+
+    const handleRightItemClick = (index: number) => {
+        setSelectedRightIndex((prev) => (prev === index ? null : index));
+    };
+
+    const handleAdd = () => {
```
michael-jaia, line 92: I'm thinking for simplicity we could remove the insert to a specific point in the combined section and just rely on the arrows to move missions around. When I used the feature, the arrows were clear to me but the insert was not. Do you feel strongly about inserting missions into the combinedList instead of just adding them to the bottom and changing the order with the arrows?

src/web/components/MissionsPanel/MissionSetEditor/MissionSetEditorDialog.tsx
```diff
@@ -0,0 +1,309 @@
...
+        ) {
+            setIsWaypointWarningVisible(true);
+            return;
+        }
+
+        setRightList(projectedList);
+        if (selectedRightIndex !== null) {
+            setSelectedRightIndex(selectedRightIndex);
+        }
+        if (!userHasOverriddenCount && addedMissionCount > desiredMissionCount) {
```
michael-jaia: After using the feature, I'm not really sure how I can override the count. If this is the number of Bots, then I think we should call it like that.

src/web/components/MissionsPanel/MissionSetEditor/MissionSetEditorDialog.tsx
```diff
@@ -0,0 +1,309 @@
...
+            setDesiredMissionCount(addedMissionCount);
+        }
+    };
+
+    const handleMoveUp = () => {
+        if (selectedRightIndex === null || selectedRightIndex === 0) return;
+        const next = [...rightList];
+        [next[selectedRightIndex - 1], next[selectedRightIndex]] = [
```
michael-jaia: This is very concise, but a little tricky to follow. I'm wondering if the longer hand format would be easier to follow.

src/web/components/MissionsPanel/MissionSetEditor/MissionSetEditorDialog.tsx
```diff
@@ -0,0 +1,309 @@
...
+                                    className="jaia-button"
+                                    disabled={!hasRightSelection}
+                                    onClick={handleDelete}
+                                >
+                                    <Icon path={mdiDelete} size={0.8} title="Delete" />
+                                </Button>
+                            </div>
+                        </div>
+                    </div>
+                    {isWaypointWarningVisible && (
```
michael-jaia, line 242: I think we should move this into a `WaypointWarning` function to keep things clean

src/web/components/MissionsPanel/MissionSetEditor/MissionSetEditorDialog.tsx
michael-jaia, line 1: We should have function comments for these functions

src/web/components/MissionsPanel/MissionSetStorage/LoadMissionSetButton/LoadMissionSetButton.tsx
```diff
@@ -49,12 +57,18 @@ export default function LoadMissionSetButton(props: Props) {
...
         if (dialogAction === DialogActions.CONFIRMED) {
-            const missionSetSnapshot = loadSnapshotFromLocalStorage(props.saveName);
+            const loadResult = loadSnapshotFromLocalStorage(props.saveName);
             jaiaDispatch({
                 type: JaiaActions.LOAD_MISSION_SET,
-                missionSetSnapshot: missionSetSnapshot,
+                missionSetSnapshot: loadResult.snapshot,
             });
 
+            if (loadResult.resultType === LoadResultType.OLD_FORMAT) {
```
michael-jaia, line 66: Should this come before we dispatch `JaiaActions.LOAD_MISSION_SET`?

src/web/components/MissionsPanel/MissionsPanel.less
```diff
@@ -3,6 +3,10 @@
     flex-direction: column;
     row-gap: 15px;
     width: 300px;
+
+    .jaia-button-row {
+        column-gap: 8px;
```
michael-jaia: Let's use 6 or 9 px

src/web/data/mission_set/mission-set.ts
```diff
@@ -77,23 +77,33 @@ export class MissionSet {
         this.missionIDInEditMode = missionIDInEditMode;
     }
 
-    getMissionSpeeds() {
-        return this.missionSpeeds;
+    getMissionSpeeds(): Speeds {
+        const first = this.missions.values().next().value;
```
michael-jaia: Could we call this firstMission? It tripped me up a little at first, I think that can be avoided if we say firstMission

src/web/data/mission_set/mission.ts
```diff
@@ -45,12 +47,20 @@ export default class Mission {
...
-    getSpeeds() {
-        return this.speeds;
+    getTransitSpeed(segmentIndex: number = 0): number {
+        return this.segments[segmentIndex]?.speed ?? DEFAULT_SPEED;
     }
 
-    setSpeeds(speeds: Speeds) {
-        this.speeds = { ...speeds };
+    setTransitSpeed(speed: number, segmentIndex: number = 0) {
+        this.segments[segmentIndex] = { ...this.segments[segmentIndex], speed };
```
michael-jaia: Could we just do `this.segments[segmentIndex].speed = speed;`

src/web/data/mission_set/mission.ts
```diff
@@ -61,12 +71,26 @@ export default class Mission {
         this.repeats = repeats;
     }
 
-    getBottomDepthSafetyParams() {
-        return this.bottomDepthSafetyParams;
+    getBottomDepthSafetyParams(segmentIndex: number = 0): BottomDepthSafetyParams | undefined {
```
michael-jaia: I think we should remove `BottomDepthSafetyParams | undefined` for consistency

src/web/data/mission_set/mission.ts
```diff
@@ -61,12 +71,26 @@ export default class Mission {
...
+        return this.segments[segmentIndex]?.bottom_depth_safety_params;
+    }
+
+    setBottomDepthSafetyParams(
+        bottomDepthSafetyParams: BottomDepthSafetyParams,
+        segmentIndex: number = 0,
+    ) {
+        this.segments[segmentIndex] = {
```
michael-jaia: Similar to comment above about setting the speed

src/web/data/mission_set/mission.ts
```diff
@@ -169,6 +190,27 @@ export default class Mission {
...
+            ...(seg.lane_start_goal_indices && {
+                lane_start_goal_indices: [...seg.lane_start_goal_indices],
+            }),
+            ...(seg.bottom_depth_safety_params && {
+                bottom_depth_safety_params: { ...seg.bottom_depth_safety_params },
+            }),
+        }));
+        // Migrate legacy mission-level speeds into the new fields
+        const legacySpeeds = (serializedMission as any).speeds;
+        if (legacySpeeds !== undefined) {
```
michael-jaia, line 201: For the `!== undefined` let's just use `!legacySpeeds` and `!legacySpeeds.transit` and same for the speed checks too

src/web/data/mission_set/mission-set.ts
```diff
@@ -77,23 +77,33 @@ export class MissionSet {
...
+            mission.setStationkeepSpeed(missionSpeeds.stationkeep_outer ?? DEFAULT_SPEED);
         }
     }
 
     addMission(mission: Mission) {
         const missionID = this.getNextMissionID();
         this.missions.set(missionID, mission);
         mission.setMissionID(missionID);
-        mission.setSpeeds(this.missionSpeeds);
+        if (mission.getSegments()[0]?.speed === undefined) {
```
michael-jaia, line 103: Let's do an `if (!mission.getSegments()[0]?.speed){}` here

src/web/custom.d.ts
```diff
@@ -1,4 +1,8 @@
-// Allows ES6 imports of SVG and PNG files to work in TypeScript
+// Allows ES6 imports of SVG, PNG, and LESS files to work in TypeScript
+declare module "*.less" {
```
michael-jaia, line 2: Why did we not need this before?

src/web/data/mission_set/mission.ts
```diff
@@ -5,26 +5,28 @@ import {
...
+    private segments: Segment[];
     private ghostParameters: GhostParameters;
 
     constructor() {
         // missionID assigned by missionSet singleton
         // speeds set by missionSet singleton
         this.waypoints = [];
+        this.stationkeepSpeed = DEFAULT_SPEED;
         this.repeats = 1;
+        this.segments = [{ start_goal_index: 1 }];
```
michael-jaia, line 29: Maybe we should init the speeds here to have less checks in the code

michael-jaia (changes requested):
Jeff, nice work, this was a big one. Structure looks good, I left some comments on more of the finer details.

src/bin/mission_manager/states/inmission.h
JRoybot: This was a mistake resolving merge conflicts from 2.y.  Will fix

src/web/components/MissionsPanel/MissionSetEditor/SaveAndLoadButton/SaveAndLoadButton.tsx
JRoybot: Good Idea, already had to do a lot of edits because Rock wanted the sides switched.

src/web/components/MissionsPanel/MissionSetEditor/SaveAndLoadButton/SaveAndLoadButton.tsx
JRoybot: Yeah, Claude tries to remove all Typescript warnings, we have a ton of false warnings related to jaiaDispatch could be undefined

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
JRoybot: Not equivalent — [] would create an empty outer array with no slots, but Array.from({ length: slotCount }, (): Mission[] => []) creates an outer array pre-populated with slotCount empty inner arrays (e.g. [[], [], []] for slotCount=3). The downstream result[i].push(...) calls depend on those slots existing. The Array.from form needs to stay.

src/web/custom.d.ts
JRoybot: we did, this removes a bunch of pre-existing warnings across our code base.

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
JRoybot: Since they are in the format of MissionSetSnapshot snapshot is the right word, I added missionSet to the names to make it clearer

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
JRoybot: It does need to be like this to simulate the way missions could be combined, and account for missions with different numbers of waypoints.  I simplified the syntax a little but kept the logic the same.

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
JRoybot: done

src/web/components/MissionsPanel/MissionSetEditor/MissionSetEditorDialog.tsx
JRoybot: Will do

src/web/components/MissionsPanel/MissionSetEditor/MissionSetEditorDialog.tsx
JRoybot: replacing left with saved and right with combined everywhere

src/web/components/MissionsPanel/MissionSetEditor/MissionSetEditorDialog.tsx
JRoybot: Not a strong need, but I thought it was a nice feature and consistent with other tools I have used.  I will look into simplifying it.

src/web/components/MissionsPanel/MissionSetStorage/LoadMissionSetButton/LoadMissionSetButton.tsx
JRoybot: No it has already been converted and will load so we want to dispatch the event, then we let the user know we had to translate it and suggest they resave it.

src/web/data/mission_set/mission-set.ts
JRoybot: done

src/web/components/MissionsPanel/MissionsPanel.less
JRoybot: went with 6

src/web/data/mission_set/mission-set.ts
JRoybot: no because it is a number an if it were 0 it would be falsey

src/web/data/mission_set/mission.ts
JRoybot: I am adding the initialization but doesn't really reduce existing checks as they are done on deserialized data

src/web/data/mission_set/mission.ts
JRoybot: agree

src/web/data/mission_set/mission.ts
JRoybot: agree

src/web/data/mission_set/mission.ts
JRoybot: from Claude ...

Unlike getTransitSpeed which always has a meaningful value (DEFAULT_SPEED fallback), bottom_depth_safety_params is a genuinely optional field — many missions simply don't have bottom depth safety params. Callers in mission-set-editor.ts explicitly check for its presence:


const mergedSRP = sourceMissions
    .map((m) => m.getBottomDepthSafetyParams())
    .find((srp) => srp !== undefined);
Removing | undefined would be a lie to the type system — TypeScript would then require callers to treat the result as always-present, which could suppress legitimate null checks and introduce real bugs. The | undefined here is correct by design, not a gap to fill.

src/web/data/mission_set/mission.ts
JRoybot: disagree because number can be 0, leaving as is

src/web/components/MissionsPanel/MissionSetEditor/MissionSetEditor.less
JRoybot: done

src/web/components/MissionsPanel/MissionSetEditor/MissionSetEditorButton.tsx
JRoybot: fixed

src/web/components/MissionsPanel/MissionSetEditor/MissionSetEditorDialog.tsx
JRoybot: can discuss, Matt and M Rock dictated how they wanted this to work.

src/web/data/mission_set/mission.ts
JRoybot: I backed out that change.  it created a bug.   There is a subtle difference between how addMission is used when missions are added manually vs added by the survey planner.  The code is using undefined speed to know when one is added manually.  I looked into trying to get it all to work and it opens up a big can of worms.  I think we should leave this as is for now.  
Matt already discussed follow on changes to how the speed sliders are displayed and used but thought it was outside of the scope of this task.

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
```diff
@@ -68,27 +68,21 @@ export function getMaxWaypointsPerOutputMission(
...
-    // so every get() is guaranteed to hit.
-    const snapshots = names.map((name) => snapshotCache.get(name)!);
-    const missionArrays = snapshots.map((snapshot) =>
-        snapshot.missions.map(([_, mission]: [number, Mission]) => mission),
-    );
-    const distributedSets = missionArrays.map((missions) =>
-        distributeMissionsToSlots(missions, desiredCount),
-    );
+    const distributedSets = names.map((name) => {
+        const missions = snapshotCache.get(name)!.missions.map(([_, m]) => m);
```
michael-jaia: I'm weary of using the !. I would prefer the classic check of:
```
const missions = snapshotCache.get(name);
if (missions) {
    // ...
}
```

JRoybot: OK

src/web/components/MissionsPanel/MissionSetEditor/MissionSetEditorDialog.tsx
JRoybot: ok

src/web/data/mission_set/mission-set.ts
michael-jaia: Ok I'm fine leaving it, but I don't think it will ever be 0 because the min speed we can set in the UI is 0.5

src/web/data/mission_set/mission.ts
michael-jaia: I guess what I am getting at is when we did the refactor, we did not include the return value in the function definition, but in this PR we are. I'm in favor of removing them to be consistent with the rest of the code.

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
michael-jaia, line 1: Jeff if I am reading things correctly, slots is equal to the input the user enters for Number of Bots. Out of curiosity, why did you choose the word slot instead of numOfBots?

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
```diff
@@ -0,0 +1,293 @@
...
+ *
+ * @param {Mission[]} missions Source missions from a single mission set
+ * @param {number} slotCount Number of output slots (desiredMissionCount)
+ * @returns {Mission[][]} Array of length slotCount, each element being the source missions for that slot
+ */
+function distributeMissionsToSlots(missions: Mission[], slotCount: number): Mission[][] {
+    // slotCount >= 1 is guaranteed by callers; both getMaxWaypointsPerOutputMission and
+    // combineMissionSets enforce desiredCount >= 1 before reaching here.
+    const result: Mission[][] = Array.from({ length: slotCount }, (): Mission[] => []);
+    const n = missions.length;
```
michael-jaia: I'm wondering if we should remove n and just use missions.length. I know it is a little longer but I think it would help readability as a developer works through the conditional section

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
```diff
@@ -0,0 +1,293 @@
...
+ * Used to validate against MAX_WAYPOINTS before saving.
+ *
+ * @param {string[]} names Ordered list of saved mission set names
+ * @param {number} desiredCount Number of output missions
+ * @param {Map<string, MissionSetSnapshot>} missionSetSnapshotCache Cache of loaded snapshots
+ * @returns {number} Maximum waypoints in any single output mission
+ */
+export function getMaxWaypointsPerOutputMission(
+    names: string[],
+    desiredCount: number,
```
michael-jaia: I'm also thinking maybe for consistency we should call this numberOfSlots or slots. I think what is tripping me up a little is some of the different words that mean the same thing. For example, I think where I enter Number of Bots on the UI that is equal to `slots` which I think is equal to `desiredCount`

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    return waypointOffset + sourceMission.getWaypoints().length;
+}
+
+/**
+ * Builds the combined output Mission for a single slot by concatenating waypoints from
+ * each source set's contribution. The leading segment carries the uniform transit speed.
+ *
+ * @param {Mission[][][]} distributedSets Per-source-set arrays of per-slot mission assignments
```
michael-jaia: I'm not totally following how we end up with a 3D array. I get we have our slots and each slot is going to have an array of missions to be combined, but I'm not sure where the third dimension comes from.

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
JRoybot: OK

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
JRoybot: Good point, I will do what I can to make them more consistency.  Unfortunately, there has been a lot of churn in this PR (70 commits) and I used Claude to write a lot of this code.

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
JRoybot: Agree it is complicated structure, AI wrote this.  If you want I will try to find a cleaner way.

Think of it as a table of lists:

Rows ([i]) — one per source mission set
Columns ([i][slot]) — one per output bot/mission slot
Cell ([i][slot]) — a list of source missions, usually length 1, but length >1 when chaining

distributedSets is a single array where each element is itself a Mission[][]. You index into it as distributedSets[setIndex][slotIndex][missionIndex], and the inner two dimensions are allocated by distributeMissionsToSlots which returns a Mission[][].

The loop in buildCombinedMission at [line 198-199](vscode-webview://1tkoe12seij4tcv97saslbo14a25fuff4899kt88g3pi5h140ug8/src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts#L198-L199) confirms this — it iterates the outer dimension (for distributedSet of distributedSets) and then indexes the next dimension directly (distributedSet[slot]), getting back a Mission[].

src/web/data/mission_set/mission.ts
JRoybot: OK, will change to implied return types

src/web/components/MissionsPanel/MissionSetEditor/MissionSetEditorDialog.tsx
JRoybot: done

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
JRoybot: slot was used to describe work-in-progress missions, idea was to make it clear they were still not complete.  The UI change came late, used to be desiredMissionCount, change was requested by Rock.  As stated in another response, I did not think it made sense to refer to bots in the code since it does not really work with bots until missions are loaded and auto assigned.

src/web/components/MissionsPanel/MissionSetEditor/SaveAndLoadButton/SaveAndLoadButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+interface Props {
+    editorName: string;
+    combinedMissionNames: string[];
+    missionSetSnapshotCache: Map<string, MissionSetSnapshot>;
+    onClose: () => void;
+}
+
+/**
+ * Button that saves the combined mission set to local storage and loads it into the active mission.
```
michael-jaia, line 20: Jeff for React Component docstrings, we just need the description, we do not include the @params and @returns

src/web/components/MissionsPanel/MissionSetEditor/SaveAndLoadButton/SaveAndLoadDialog.tsx
```diff
@@ -0,0 +1,79 @@
...
+    onClose: (dialogAction: DialogActions) => void;
+}
+
+interface ButtonRowProps {
+    disabledCode: DisabledCodes;
+    onClose: (dialogAction: DialogActions) => void;
+}
+
+/**
+ * Confirmation or alert dialog for the Save and Load action.
```
michael-jaia, line 17: Same as above

src/web/components/MissionsPanel/MissionSetEditor/SaveAndLoadButton/SaveAndLoadDialog.tsx
```diff
@@ -0,0 +1,79 @@
...
+        <div className={`secondary-dialog${isAlert ? " alert" : ""}`}>
+            <h1>{isAlert ? "Alert" : "Confirm"}</h1>
+            <p>{getDialogMessage()}</p>
+            <ButtonRow disabledCode={props.disabledCode} onClose={props.onClose} />
+        </div>
+    );
+}
+
+/**
+ * Renders a Close button for alert dialogs, or Cancel and Save and Load buttons for confirm dialogs.
```
michael-jaia, line 49: Same as above

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
```diff
@@ -0,0 +1,158 @@
...
+ * Used to validate against MAX_WAYPOINTS before saving.
+ * @param {string[]} names Ordered list of saved mission set names
+ * @param {Map<string, MissionSetSnapshot>} missionSetSnapshotCache Cache of loaded snapshots
+ * @returns {number} Maximum waypoints in any single output mission
+ */
+export function getMaxWaypointsPerOutputMission(
+    names: string[],
+    missionSetSnapshotCache: Map<string, MissionSetSnapshot>,
+): number {
+    const slotCount = getMaxMissionCount(names, missionSetSnapshotCache);
```
michael-jaia: I'm in favor of calling this missionCount. I think it would be good to not introduce another word into the mix that means mission,

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * Appends one source mission's waypoints and offset segments to the combined output mission.
+ * @param {Mission} sourceMission Source mission to append
+ * @param {Mission} combined Output mission being built (mutated)
+ * @param {Segment[]} combinedSegments Output segments array to merge SRP data into (mutated)
+ * @param {number} waypointOffset Number of waypoints already added to combined before this call
+ * @returns {number} Updated waypointOffset after appending the source mission's waypoints
+ */
+function applySourceMission(
```
michael-jaia: If we have the source mission, do we need a separate parameter for waypointOffset? That should be the length of the combined mission's waypoint array right?

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
```diff
@@ -0,0 +1,158 @@
...
+    sourceMission: Mission,
+    combined: Mission,
+    combinedSegments: Segment[],
+    waypointOffset: number,
+): number {
+    combined.addWaypoints(cloneDeep(sourceMission.getWaypoints()));
+
+    for (const seg of sourceMission.getSegments()) {
```
michael-jaia, line 65: I could see that the combined mission needs segments, but why does the source mission need segments? Isn't the source  mission just one segment?

JRoybot: Matt decided that going forward all missions would use segments and for simple missions those would contain 1 segment, (although we do support the old format, for the REST API).

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
JRoybot: all the same, was trying to make the function profile explicit, will change

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
michael-jaia: Can we remove `combinedSegments` and call `combinedMission.getSegments()`. I think that would be nice because your parameters match the classic copy parameters of an input and output.

michael-jaia: Can we change this set of `slotCount`s too. Should we use the function `getMaxMissionCount` like we use above?

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
```diff
@@ -0,0 +1,141 @@
...
+export function getMaxWaypointsPerOutputMission(
+    names: string[],
+    missionSetSnapshotCache: Map<string, MissionSetSnapshot>,
+): number {
+    const missionCount = getMaxMissionCount(names, missionSetSnapshotCache);
+    if (names.length === 0 || missionCount < 1) return 0;
+
+    let maxWaypoints = 0;
+    for (let slot = 0; slot < missionCount; slot++) {
+        let slotTotal = 0;
```
michael-jaia: Could we call this waypointTotal?

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
```diff
@@ -0,0 +1,141 @@
...
+ */
+export function getMaxWaypointsPerOutputMission(
+    names: string[],
+    missionSetSnapshotCache: Map<string, MissionSetSnapshot>,
+): number {
+    const missionCount = getMaxMissionCount(names, missionSetSnapshotCache);
+    if (names.length === 0 || missionCount < 1) return 0;
+
+    let maxWaypoints = 0;
+    for (let slot = 0; slot < missionCount; slot++) {
```
michael-jaia: Maybe this should just be `i` instead of `slot`

src/web/components/MissionsPanel/MissionSetEditor/mission-set-editor.ts
```diff
@@ -0,0 +1,141 @@
...
+    let maxStationkeep = DEFAULT_SPEED;
+    for (const missions of missionArrays) {
+        for (const mission of missions) {
+            maxTransit = Math.max(maxTransit, mission.getTransitSpeed());
+            maxStationkeep = Math.max(maxStationkeep, mission.getStationkeepSpeed());
+        }
+    }
+
+    const outputMissions: [number, Mission][] = [];
+    for (let slot = 0; slot < missionCount; slot++) {
```
michael-jaia: Yeah I think `i` would be clearer here since slot just represents the index/counter

michael-jaia:
Jeff the changes look good to me. I would like us to remove the click animation when adding a mission to be combined. Also, we need to add docstrings to the functions in `MissionSetEditorDialog.tsx`. In that component too, I think we will want to move `const savedMissionSets = listSavedMissionSets();` so it is only called once, not every 0.5 seconds.

michael-jaia:
Jeff I made some changes to tighten up the JSX and LESS code. I felt like there were some spots where we had some redundant divs and class names. For the docstrings, I was talking about the functions such as `handleAdd`, `handleMoveUp`, etc

michael-jaia (approved):
Image

Solid code, excited to use on the water

PR #1547, opened 2026-04-17
feature/low-battery-protocol-backend-sw-2221

edsanville:
* Added options to the MissionPlan message to configure the behavior when the battery is low or critically low.
* Added helper functions to recursively collect errors and warnings from a VehicleHealth message, and use those functions to check for battery errors.
* For battery errors, we issue an EvBatteryLevel event to the state machine.
* Added EvStartBatteryProtocol event to trigger the start of the proper battery protocol
* Added new states:
  * in_mission::Battery
    * StationKeep
    * StopAndBroadcast
* All states that should react to battery errors should react to EvBatteryLevel, and call the InMission::start_battery_protocol() function to dispatch to the correct battery protocol state.

Note:  StationKeep is not working yet, and is being debugged.


src/bin/mission_manager/states/inmission.h
```diff
@@ -181,11 +181,94 @@ struct InMission
...
+        jaiabot::protobuf::MissionPlan::BatteryProtocol battery_protocol;
+
+        switch (battery_level)
+        {
+            case BatteryLevel::CRITICAL:
+                if (machine().mission_plan().has_critically_low_battery_protocol())
+                {
+                    battery_protocol = machine().mission_plan().critically_low_battery_protocol();
+                }
+                else
```
ferrom-jaia, line 198: Do we need this else?

Not sure we want to do anything if the user does not set a protocol

src/bin/mission_manager/states/inmission.h
```diff
@@ -181,11 +181,94 @@ struct InMission
...
+                        protobuf::MissionPlan::BatteryProtocol::STOP_AND_BROADCAST);
+                }
+                break;
+
+            case BatteryLevel::VERY_LOW:
+                if (machine().mission_plan().has_very_low_battery_protocol())
+                {
+                    battery_protocol = machine().mission_plan().very_low_battery_protocol();
+                }
+                else
```
ferrom-jaia, line 209: Do we need this else?

Not sure we want to do anything if the user does not set a protocol

src/bin/mission_manager/states/inmission.h
```diff
@@ -181,11 +181,94 @@ struct InMission
...
+                return;
+
+            default:
+                glog.is_debug1() &&
+                    glog << "  Received unknown battery level, not starting battery protocol"
+                         << std::endl;
+                return;
+        }
+
+        if (battery_protocol.action() == protobuf::MissionPlan::BatteryProtocol::NONE)
```
ferrom-jaia, line 227: Should this be the first check?

ferrom-jaia (commented):
Seemed to work in sim. 

Just a little confused with what I saw. 

Without a default action the bot did not go into battery state (Expected)

I set a default action to stop and broadcast in the mission proto and the bot station keeped at very low battery then stopped and broadcasted at critical.

src/bin/mission_manager/states/inmission.h
edsanville: That makes sense.  I will remove this `else` and make the function a no-op in that case.

src/bin/mission_manager/states/inmission.h
edsanville: Removed.

src/bin/mission_manager/states/inmission.h
edsanville: It is the first check after we set the `battery_protocol` variable in the switch statement.

edsanville:
> Seemed to work in sim.
> 
> Just a little confused with what I saw.
> 
> Without a default action the bot did not go into battery state (Expected)
> 
> I set a default action to stop and broadcast in the mission proto and the bot station keeped at very low battery then stopped and broadcasted at critical.

This happened because the `very_low_battery_protocol` and `critically_low_battery_protocol` fields were unset, (which doesn't use the default action in the `BatteryProtocol` message).  In the code, it was defaulting to STATION_KEEP for very_low and STOP_AND_BROADCAST for critical.  

Now, we've replaced those with NONE, so by default it will do nothing.

It's a bit hard to test, actually, without having the GUI in place for specifying the battery protocol in the MissionPlan.  One could use liaison, but building missions in there is a bit cumbersome.

PR #1553, opened 2026-05-11
Add ML pipeline for pre-mission battery drain prediction

AdeX11:
## Summary

This PR adds a machine-learning model that **predicts how much battery a bot will use during a mission**, before the mission starts. The prediction is shown in the mission planning UI so operators can decide whether a mission is feasible on current battery charge.

All prediction code lives inside the **pyjaia** Python package on the hub (`src/python/pyjaia/src/pyjaia/battery_prediction/`). The hub exposes a `/battery-prediction` endpoint; the web client sends mission plan details and gets back a predicted drain percentage.

## What was built

| File | Purpose |
|---|---|
| `extract_features.py` | Reads historical HDF5 log files from real missions and extracts the numbers used for training |
| `train.py` | Trains the model on that historical data and saves it to `battery_model.pkl` |
| `inference.py` | Loads the saved model and answers "how much battery will this mission use?" |
| `dataset.csv` | The labeled training data (62 missions from real fleet deployments) |
| `battery_model.pkl` | The trained model file, bundled inside the package so the hub can load it at startup |

**Training dataset:** [SharePoint — Jaia Battery Prediction Logs](https://jaiatech.sharepoint.us/:f:/s/ProductDevelopment/IgCduK9EBWvpRaRHkfycjwUkASNdRP6554lh7eQhUSsdvTg?e=NkARIn)

## How mission boundaries are detected

Each mission is identified by a **hub command packet** logged in HDF5 — the hub sends one of these every time the operator starts or replans a mission. If a newer hub command arrives, the previous mission is considered complete. If no newer command arrives, the end of the mission is identified by the bot entering recovery/station-keep mode.

## Input features

These are the numbers fed into the model. All of them are computable from the mission plan *before* the mission starts, except `starting_battery_pct` which is read from the bot at mission start.

| Feature | What it captures |
|---|---|
| `bot_type` | Bot hardware type — different bot types have different motor and power profiles |
| `transit_energy_wh` | Physics-based estimate of motor energy for all travel legs, computed from the power curve (speed → watts) |
| `transit_time_s` | Total planned travel time in seconds |
| `turn_density_deg_per_km` | How "twisty" the route is — total turning angle divided by total distance |
| `drift_total_s` | Planned time drifting on the surface (motor off, hotel load only) |
| `station_keep_total_s` | Planned time holding position (motor actively fighting current) |
| `dive_count` | Number of dive tasks — each dive has a fixed energy cost for descent and GPS re-acquisition |
| `mean_dive_depth_m` | Average depth of dives |
| `dive_hold_s` | Total planned time sitting at depth (motor off, hotel load only) |
| `dive_hold_stops` | Total measurement stops across all dives |
| `starting_battery_pct` | How charged the battery is when the mission begins |

## The model

**What "machine learning" means here:** we show the model 62 real missions where we know exactly how much battery was drained. The model learns which combinations of features best explain the variation in drain across those missions. It can then estimate drain for a new mission it has never seen.

**Monotonic constraints:** a core design requirement is that the model must behave sensibly — more dives should never predict *less* drain. We enforce this mathematically using positive-only coefficients, so the model is guaranteed to be monotone in every feature.

**Model type:** Ridge regression with degree-2 polynomial features (interaction-only). "Polynomial" means the model can capture interactions between features (e.g. dive cost varying by bot type), not just simple linear relationships. "Ridge" applies a gentle penalty that keeps coefficients small but rarely zeroes them out, so the model retains a richer view of which factors matter.

**Why Ridge poly2 over a slightly more accurate Lasso variant:** training compared six candidate models. Lasso poly2 had marginally lower error (3.53% vs 3.61% MAE), but its aggressive L1 regularization zeroed out 60 of 66 polynomial terms — dropping `dive_hold_s` and `dive_hold_stops` entirely, even though hold time at depth has a clear physical effect on drain (the bot is still drawing hotel load). Ridge poly2 keeps 24 of 66 terms and preserves those physical signals. The training script's selection criterion picks the model with the most non-zero terms among those within 0.5% MAE of the best — encoding the principle "if accuracy is essentially tied, prefer the model that captures more of the real-world physics."

**Validation method — Leave-One-Out cross-validation:** we train on 61 missions, predict the one left out, then repeat this 62 times (once per mission). This gives an honest estimate of accuracy on unseen missions, avoiding the trap of testing on data the model already saw.

| Metric | Value |
|---|---|
| Mean Absolute Error (LOO) | **3.61%** |
| Median Absolute Error (LOO) | 2.80% |
| Best single-mission prediction | within ~0.5% |
| Worst single-mission prediction | ~18% error |
| Missions within ±3% of actual | 58% |
| Missions within ±5% of actual | 79% |

A 3.6% mean error means: if the model predicts a mission will use 25% battery, it is typically within ±3.6 percentage points of the real outcome.

## Feature importance

Ridge poly2 produced 24 non-zero terms out of 66 candidates. The most influential, ranked by coefficient magnitude (on standardized features):

| Term | Coefficient | Plain meaning |
|---|---|---|
| `transit_energy_wh × transit_time_s` | +4.65 | Long-duration high-energy transits drain more than each factor alone would suggest |
| `transit_energy_wh × starting_battery_pct` | +4.01 | High-energy transits drain more when starting at higher charge — a correlation in the training data (operators tend to plan longer missions on full batteries), not a direct physical effect |
| `transit_energy_wh` | +3.24 | Baseline: more motor energy planned → more drain |
| `dive_count` | +1.54 | More dives → more drain, independent of depth or bot type |
| `bot_type × dive_count` | +1.28 | Dive cost differs between bot types |
| `dive_hold_stops` | +1.25 | More measurement stops mid-dive → more drain |
| `bot_type × dive_hold_stops` | +1.14 | Stop cost varies by bot type |
| `dive_count × dive_hold_stops` | +1.13 | Compound effect: many dives with many stops scale super-linearly |
| `bot_type × turn_density_deg_per_km` | +0.84 | Turn penalty varies by bot type |
| `turn_density_deg_per_km × dive_count` | +0.67 | Twisty routes amplify the cost of having many dive sites |

Plus 14 smaller terms involving `dive_hold_s`, `transit_time_s` interactions, and others.

All 24 coefficients are positive, confirming the monotonicity guarantee is respected.

## Package changes

- `pyproject.toml`: added `scikit-learn` and `pandas` dependencies; `.pkl` and `.csv` bundled as package data; CLI entry points `jaia-battery-extract` and `jaia-battery-train`

## Test plan

- [ ] `jaia-battery-extract --logs-dir <hdf5 dir> --output dataset.csv` — verify expected columns and ~62 rows
- [ ] `jaia-battery-train --dataset dataset.csv --model-output battery_model.pkl` — confirm Ridge poly2 selected, LOO MAE ~3.6%
- [ ] POST to `/battery-prediction` with valid feature JSON — verify a float drain percentage is returned
- [ ] `from pyjaia.battery_prediction import predict_drain` works after `pip install -e ./pyjaia`
- [ ] Model sanity check: increasing `dive_count` or `transit_energy_wh` never decreases the predicted drain


michael-jaia:
Ade, looks good, is it possible to get a break down on what variables have the biggest impact on battery drain? I'm curious if we reduced the variables from 14 to 5 (for example) how would the LOO MAE and RMSE change?

marshalllele (dismissed):
Hey Ade, this looks like a pretty good start for this effort. I read through the code and tested against my local logs and everything tracks with your results as well.

For a next effort, I think it would be interesting to include log files from sea trials, testing, or trainings going forward. That way with each release, the results get better and better. 

`Best model: Ridge (LOO MAE = 3.26% RMSE = 3.49%)`
`

AdeX11:
> Ade, looks good, is it possible to get a break down on what variables have the biggest impact on battery drain? I'm curious if we reduced the variables from 14 to 5 (for example) how would the LOO MAE and RMSE change?

updated the PR to include some info on feature importance. Some features arent importnat right now because our dataset isnt as diverse, but they could become important later. I was trying to build in such a way that you could run with new data and it will still work well.

AdeX11:
> Hey Ade, this looks like a pretty good start for this effort. I read through the code and tested against my local logs and everything tracks with your results as well.
> 
> For a next effort, I think it would be interesting to include log files from sea trials, testing, or trainings going forward. That way with each release, the results get better and better.
> 
I updated the PR to show the sharepoint link to the dataset I used

AdeX11:
Superseded by #1557 (`feature/battery-prediction-physics-features`), which restructures the model to use calibrated energy features measured directly from log data instead of plan-derived durations. The new branch is a strict superset of what was here, plus the calibration pipeline and a physics-adherence test suite. Closing this one to consolidate review on #1557.

PR #1554, opened 2026-05-12
feature/mission-battery-feasibility

AdeX11:
Adds the operator-facing side of the battery prediction feature. When a mission is planned in JCC and a bot is assigned, the client computes the plan features, calls `POST /battery-prediction` on the hub, and uses the predicted final battery percentage to color the mission accordion in the side panel.

The hub model and inference path are in #1557 (`feature/battery-prediction-physics-features`).

## UI behavior

- **Green accordion** — predicted final battery is above the 20% safety threshold; mission is feasible
- **Red accordion** — predicted final would be below 20%; the operator gets a visual warning before sending the bot
- **Grey accordion** — no bot assigned yet, or the prediction is still in flight (avoids a green-flash before the first response arrives)
- Predictions are debounced 500 ms after any mission or bot change so we're not hammering the hub on every keystroke
- The mission stats block shows the predicted "Battery After" alongside the existing fields

## What the client sends

`utils/battery_prediction.ts` walks the mission plan and computes the same seven features the hub model expects (matching `pyjaia.battery_prediction.predict_drain`):

```ts
{
  bot_type,                 // HYDRO=1, ECHO=2, BIO=3
  transit_energy_wh,        // calibrated curve(planned_speed) × distance / speed
  transit_time_s,           // planned distance / speed, summed across all legs
  turn_density_deg_per_km,  // total turn angle / total distance
  hotel_energy_wh,          // 62.8·dive_hold_s + 0.9·drift_s + 7.2·sk_s, all /3600
  dive_energy_wh,           // dive_count × (0.71 + 0.025 × mean_dive_depth_m)
  starting_battery_pct,     // current bot battery
}
```

The wattage constants mirror `pyjaia/battery_prediction/calibration.json` so the client doesn't need an extra HTTP round-trip on every prediction. If those constants change (re-calibration), both the JSON and the TS file get updated together — they're six numbers in two places.

A few details worth knowing:

- **Planned speed is clamped at 2.0 m/s** before the lookup. Logs show the bot can't actually go slower (commanded 1.0 m/s achieves ~1.7 m/s), so a sub-2.0 plan would otherwise inflate the predicted transit time 2× and over-predict drain.
- **Return-to-start legs between repeats are included** in `transit_energy_wh`. If a mission repeats, the bot has to come back to the first waypoint each time.
- **The bot's current GPS position** is the start of the first transit leg, not the first waypoint — so an off-station bot pays for the leg to get there.
- **Constant-heading task displacements** are tracked in a running position so the energy and time accumulate correctly.

## Changes

- `utils/battery_prediction.ts` — plan walker + `/battery-prediction` POST
- `utils/constants.ts` — `BOTTOM_DIVE_DEPTH_PRIOR_M = 10` (used as the assumed depth when a dive task has no explicit `max_depth`)
- `components/MissionsPanel/MissionsList/MissionsList.tsx` — accordion color coding, debounced prediction call, battery-after stat in the body
- `openlayers/layers/vector/excluded-task-packets-layer.ts` — typo fix unrelated to battery prediction

## Dependency

This branch needs #1557 merged first — that's where the `/battery-prediction` endpoint and the trained model live. Without it the client will get a 404. (The previous PR #1553 was closed in favor of #1557, which is a strict superset.)

## Test plan

- [ ] Assign a fully-charged bot to a short mission — accordion turns green
- [ ] Assign a low-battery bot, or stretch the mission long enough that predicted final < 20% — accordion turns red
- [ ] Unassign the bot, or open a fresh mission before any prediction has returned — accordion is grey, not green
- [ ] Change a waypoint or repeat count — the prediction updates ~500 ms later, no flicker during typing
- [ ] Verify the "Battery After" stat reflects the predicted final battery
- [ ] Add a mission with repeats > 1 and confirm the predicted drain accounts for return legs (compare to a single-pass plan with same waypoints)


joeymussalli:
## Summary

### Building on Ade's battery drain prediction work, the recent changes add:

- A server side guard so no prediction is produced for bot types the model wasn't trained on. A message appears in the Battery After field when that triggers to avoid user confusion.
- Test coverage for the feature.

### A few changes made to the original implementation:

- Predicted final battery % could display as negative. It is now clamped to 0% for display only. 
- The frontend kept its own hardcoded copy of the server's calibration constants and they differed slightly from the server version. The server now stores the only version.
- Predicted drain % was fully unbounded. It is now capped at 1000% for display to prevent unrealistic drain values from being shown to the user.
- For missions with a constant heading task, the frontend was estimating a new bot position after that task and using it to calculate the return trip distance for repeated missions. The training data did not match this logic, throwing off predictions. The frontend calculations were changed to match the model's training process.
- Clicking Start Mission while not in control could bypass the last fetched insufficient battery warning because the check read a stale state.

## Status

I think this is in good shape to merge. Would like one more person to stress test it in the simulator before we merge.

src/python/pyjaia/src/pyjaia/battery_prediction/battery_model.pkl
JRoybot, line 1: what is in this pickle file?  does it need to be in the repo?

src/python/pyjaia/src/pyjaia/battery_prediction/calibrate.py
```diff
@@ -0,0 +1,374 @@
...
+
+def _haversine(lat1, lon1, lat2, lon2):
+    to_rad = np.pi / 180
+    dlat = (lat2 - lat1) * to_rad
+    dlon = (lon2 - lon1) * to_rad
+    a = (np.sin(dlat / 2) ** 2
+         + np.cos(lat1 * to_rad) * np.cos(lat2 * to_rad) * np.sin(dlon / 2) ** 2)
+    return 2 * EARTH_R * np.arcsin(np.sqrt(np.clip(a, 0, 1)))
+
+# Mission state enums from src/lib/messages/mission.proto
```
JRoybot: why is this duplicated instead of imported?

src/python/pyjaia/src/pyjaia/battery_prediction/dataset.csv
JRoybot, line 1: what is this file used for?

src/web/components/__buttons__/StartAllMissionsButton/StartAllMissionsButton.tsx
```diff
@@ -47,27 +48,42 @@ export default function StartAllMissionsButton(props: Props) {
...
-
-            // Download queue
-            else if (isCritiallyLowBattery(bot.getBatteryPercent())) {
-                updatedBotReadyStates.get(DisabledCodes.LOW_BATTERY).push(botID);
-            } else {
-                updatedBotReadyStates.get(DisabledCodes.NONE).push(botID);
-            }
-        }
+        await Promise.all(
+            Array.from(props.bots.entries()).map(async ([botID, bot]) => {
```
JRoybot: this is a pretty big arrow function to be in a map statement, especially since the outer block is also an arrow function... maybe break it out?

src/web/components/__buttons__/StartMissionButton/StartMissionButton.tsx
```diff
@@ -79,6 +83,13 @@ export default function StartMissionButton(props: Props) {
             return DisabledCodes.LOW_BATTERY;
         }
 
+        // A null prediction (unsupported bot type, no waypoints, request failure, etc.)
+        // intentionally does not block starting the mission -- the battery check is
+        // advisory, not a safety gate, so we fail open when we have no prediction to check.
+        if (prediction !== null && prediction.predicted_final_pct < MIN_BATTERY_PERCENT) {
```
JRoybot: whats the distinction between LOW_BATTERY & INSUFFICIENT_BATTERY?

src/web/components/__buttons__/StartAllMissionsButton/StartAllMissionsButton.tsx
```diff
@@ -78,7 +94,7 @@ export default function StartAllMissionsButton(props: Props) {
      *
      * @returns {void}
      */
-    const handleClick = () => {
+    const handleClick = async () => {
```
JRoybot: Can these aysnc calls fail to return?  Should there be an exception handler around them?

joeymussalli:
That's a good find. I think it's worth changing as to not confuse the user.

src/web/components/MissionsPanel/MissionsList/MissionsList.tsx
```diff
@@ -118,81 +148,198 @@ export default function MissionsList() {
...
+        }
+
+        if (bot.getBatteryPercent() < MIN_BATTERY_PERCENT) {
+            setDisabledCode(DisabledCodes.LOW_BATTERY);
+            setPrediction(null);
+            return;
+        }
+
+        const timer = setTimeout(async () => {
+            const result = mission ? await fetchBatteryPrediction(mission, bot) : null;
```
JRoybot: Rather than have the code to fetch these predictions in the 2 buttons and in this list.  Can we do this periodically for all the missions and store it in a singleton in the data model and add that to JaiaContext like other data needed across the application, then access the data via the context in the components.  That way they will all stay and sync and we are not fetching in reponse to button clicks.

JRoybot (commented):
I have only done a partial review, but based on my first read and some free testing I think we would be better off if the battery predictions were distributed through the data model and context like the rest of our shared data.  That is a pretty big restructure.  We can discuss if needed.

src/web/components/MissionsPanel/MissionsList/MissionsList.tsx
joeymussalli: Something we can consider for sure. Because of the nature of this feature -- where users expect to get an updated value shortly or immediately after editing a mission -- I would lean towards refreshing the predictions upon certain actions as we do now in addition to periodic updates. I think more than a few seconds in between predictions would feel clunky.

src/python/pyjaia/src/pyjaia/battery_prediction/dataset.csv
joeymussalli: It houses all the data the model is trained on. The numbers represent the different model inputs labeled at the top of the file.

joeymussalli: It is helpful to have if we wish to retrain the model in the future with additional data

src/python/pyjaia/src/pyjaia/battery_prediction/battery_model.pkl
joeymussalli: Yeah that's the prediction model itself. It stores the trained weights

src/python/pyjaia/src/pyjaia/battery_prediction/calibrate.py
joeymussalli: You're right the function is duplicated several times. I believe the scripts were made to run on their own but we can consolidate if you think that would be better

src/web/components/__buttons__/StartAllMissionsButton/StartAllMissionsButton.tsx
joeymussalli: I agree, will change

src/web/components/__buttons__/StartMissionButton/StartMissionButton.tsx
joeymussalli: LOW_BATTERY means the bot is already on low battery. INSUFFICIENT_BATTERY means the bot will be too low on battery by the end of the mission

src/web/components/__buttons__/StartAllMissionsButton/StartAllMissionsButton.tsx
joeymussalli: The calls are built to never throw so no handlers needed.

src/python/pyjaia/src/pyjaia/battery_prediction/calibrate.py
JRoybot: I was actually referring to the enum.  Aren't python versions of the protobuf messages created by the proto compiler?  Even if they are manually maintained they should only be declared in one place.

src/web/components/MissionsPanel/MissionsList/MissionsList.tsx
JRoybot: I'll continue to review and make some suggestions.  I would be surprised if the values change much in a few seconds, especially if the bots are drifting or station keeping waiting for a mission, and it probably would be light load to update them once a second.  It would be pretty straightforward to do both a periodic and an event driven update if needed.  Just need to follow the action->reducer->handler model to update context and useContext instead of useEffect

src/web/components/MissionsPanel/MissionsList/MissionsList.tsx
```diff
@@ -118,81 +148,198 @@ export default function MissionsList() {
...
+ * @param {Function} props.onAccordionChange Called when the accordion is expanded or collapsed
+ * @param {Function} props.onRepeatsChange Called when the repeats value is changed
+ * @param {Function} props.onDuplicateClick Called when the duplicate button is clicked
+ * @param {Function} props.onToggleEditClick Called when the edit toggle is clicked
+ * @returns {JSX.Element} The mission accordion
+ */
+function MissionAccordion(props: MissionAccordionProps) {
+    const jaiaContext = useContext(JaiaContext);
+    const [disabledCode, setDisabledCode] = useState<DisabledCodes>(DisabledCodes.NO_MISSION);
+    const [prediction, setPrediction] = useState<BatteryPrediction | null>(null);
```
JRoybot: this looks like a candidate for context

src/web/components/MissionsPanel/MissionsList/MissionsList.less
```diff
@@ -43,3 +43,30 @@
...
+    color: @cc-accordian-color;
+    padding-right: 12px;
+}
+
+.mission-stats-value {
+    font-weight: 500;
+}
+
+.mission-accordion-summary--unassigned {
+    background-color: @cc-accordian-color !important;
```
JRoybot, line 63: Just a matter of taste but I am not a big fan of the big green and red summary bars.  Might be cleaner if they were not so tall or use a colored box, icon or circle for the status instead of the entire bar.  Not a bid deal, just worth mentioning

src/web/data/battery_predictions/battery-prediction-calculator.ts
```diff
@@ -0,0 +1,376 @@
...
+    const turnDensityDegPerKm = totalDistanceKm > 0 ? totalTurnAngleDeg / totalDistanceKm : 0;
+    return { turnDensityDegPerKm };
+}
+
+/**
+ * Extracts mission features from a Mission and Bot, sends them to the hub's
+ * battery prediction endpoint, and returns the predicted drain and final battery percentage
+ *
+ * @param {Mission} mission Used to derive waypoint distances, dive counts, and depth
+ * @param {Bot} bot Used to provide the starting battery percentage and bot type
```
JRoybot, line 204: Why do we need to pass in the bot?  We can only do a prediction if the mission has a bot assigned, in that case we can get it from the mission manager.  If we change the structure so the predictions are done in one place and stored in the data-model/context they we can just loop through the missions and get the prediction for each mission.  In that model it would make more sense to just check for bot assignment in the prediction code.

src/web/utils/battery_prediction.ts
```diff
@@ -0,0 +1,376 @@
...
+/**
+ * Calculates the great-circle distance in meters between two geographic coordinates
+ *
+ * @param {number} lat1 Latitude of the first point in degrees
+ * @param {number} lon1 Longitude of the first point in degrees
+ * @param {number} lat2 Latitude of the second point in degrees
+ * @param {number} lon2 Longitude of the second point in degrees
+ * @returns {number} Distance between the two points in meters
+ */
+function haversineMeters(lat1: number, lon1: number, lat2: number, lon2: number): number {
```
JRoybot: we definitely should put this function in utilities rather than duplicate it

src/web/server/app.py
```diff
@@ -550,6 +554,57 @@ def get_ctd_profiles():
         mimetype="application/zip",
     )
 
+@app.route('/battery-calibration', methods=['GET'])
```
JRoybot: I'm not that familiar with the structure of the server code.  Would these functions fit better in src/web/server/jaia_portal.py

JRoybot (commented):
A few more follow up comments.... I am still reviewing this

src/python/pyjaia/src/pyjaia/battery_prediction/calibrate.py
joeymussalli: Oh I see, yeah I'll change that

src/web/server/app.py
joeymussalli: From my understanding jaia_portal.py is related to a hub connection, which the model doesn't deal with.

src/web/components/__buttons__/StartAllMissionsButton/StartAllMissionsButton.tsx
```diff
@@ -53,35 +54,8 @@ export default function StartAllMissionsButton(props: Props) {
...
-                        prediction !== null &&
-                        prediction.predicted_final_pct < MIN_BATTERY_PERCENT
-                    ) {
-                        updatedBotReadyStates.get(DisabledCodes.INSUFFICIENT_BATTERY).push(botID);
-                    } else {
-                        updatedBotReadyStates.get(DisabledCodes.NONE).push(botID);
-                    }
-                }
+                const disabledCode = await getBotReadyState(botID, bot, props.missions);
+                updatedBotReadyStates.get(disabledCode).push(botID);
```
JRoybot: Maybe I am missing something.  This looks like it is only updating the bot read status locally and none of these updates are being saved in the JaiaContext.  What is going to keep the dialog in sync with the rest of the components, particularly the MissionsList?

src/web/components/__buttons__/StartAllMissionsButton/StartAllMissionsButton.tsx
```diff
@@ -205,3 +179,45 @@ function isMissionUnassigned(botID: number) {
 function isCritiallyLowBattery(batteryPercent: number) {
     return batteryPercent < MIN_BATTERY_PERCENT;
 }
+
```
JRoybot: Why is this different than the similar function in StartMissionButton?  Shouldn't they be the same.  Could we combine some of the logic so it is not redundant?

src/web/context/handlers/data-model-handers.ts
```diff
@@ -71,6 +72,8 @@ export function handleInit(mutableState: JaiaContextType) {
         missionsManager: missionsManager,
         taskPackets: taskPackets,
         exclusionZoneSet: exclusionZoneSet,
+        batteryPredictions: batteryPredictions,
+        predictionsVersion: 0,
```
JRoybot: This is used to track each prediction and debounce updates?  Not stored in the data model anywhere, just in context?  I think versions is a little misleading it is basically a counter.

src/web/context/JaiaContext.tsx
```diff
@@ -67,6 +74,21 @@ export function JaiaContextProvider({ children }: JaiaContextProviderProps) {
         return () => clearInterval(intervalID);
     }, []);
 
+    /**
```
JRoybot: I am uncomfortable with adding more business logic in the reducer. ... Will give this more thought to try to develop a cleaner way to handle the predictions updates.  I do not think there is a real need to debounce these updates.  The only asynchronous updates are from the 2 start buttons.  I would like to see a process where the periodic updates always run as scheduled and IF NEEDED the buttons produce an action/event to trigger an asynchronous update and the values are always read from context so the dialogs and the missions list can not get out of sync.   I am also not convinced that data that is less then a second old in the dialogs is a critical issue, do we have any analysis to support that?

src/web/data/battery_predictions/__tests__/battery-predictions.test.ts
```diff
@@ -0,0 +1,30 @@
+import { BatteryPredictions } from "../battery-predictions";
+
+test("getStatus returns undefined for a mission with no computed status yet", () => {
+    const predictions = new BatteryPredictions();
```
JRoybot, line 4: Not sure I understand this test.  We will never replace the singleton map in the data model with a new one.  It should be created in the constructor and used in context and always be the same refernce.

src/web/data/battery_predictions/battery-prediction-calculator.ts
JRoybot, line 1: Now that there is a src/web/data/battery_predictions directory these functions should be moved to that directory to be consistent with how other parts of the data model are structured.

src/web/jcc/index.tsx
```diff
@@ -1,27 +1,34 @@
...
 import {
+    BATTERY_PREDICTION_POLL_TIME,
     DATA_MODEL_POLL_TIME,
     GITHUB_POLL_TIME,
     INTERNET_POLL_TIME,
     METADATA_POLL_TIME,
     TASK_PACKET_POLL_TIME,
 } from "../utils/constants";
 import { pollGitHub, pollInternet, pollMetadata, pollStatus, pollTaskPackets } from "./polling";
+import { refreshBatteryPredictions } from "../context/handlers/battery-prediction-handlers";
```
JRoybot: I would prefer to see this integrated with the rest of the polling functions for consistency.  IF we do need the asynchronous updates we can add a specific even handler for that.

src/web/context/handlers/battery-prediction-handlers.ts
JRoybot, line 1: The handlers and the reducer are intended for handling actions/events.  This does not follow that pattern at all.  As this stands now it should be part of src/web/jcc/polling.ts.  IF aysncronous updates are needed for the start button dialogs we should trigger that update with an event.

JRoybot (commented):
Reviewed the updates.  Nice step in the right direction but I do not think it is in line with the patterns used in the rest of the code that work with context.  A big question in my mind is how critical the asynchronous updates for button dialogs are really needed.  How much do the predictions change in less than a second when a bot is drifting or position keeping waiting for a mission?  IF they are really needed we could add an extra event-driven update to the data mode.  We can discuss this more before you make more updates.

JRoybot (approved):
Updates look good, nice job.  Reviewed code and tested in browser with simulator.  Found and fixed one small bug on Start button visibility

PR #1556, opened 2026-05-14
Bug/hub2hub not initialized properly failure/sw 2367

tsaubergine:
## Overview

Fixes bug when two hubs have a different INTERVEHICLE_API version: https://jaia-innovation.atlassian.net/browse/SW-2367

## Bug

 previously this wasn't detected for the Hub2HubData message, potentially causing crashes if the message versions were incompatible.

## Fix

This PR adds code for the hub_manager to detect the mismatch for the Hub2HubData message much like the Command message.

Upon subscription to Hub2HubData from some other hub:

1. If this hub is older then the other hub, the error is:
ERROR__VERSION__MISMATCH_INTERVEHICLE__UPGRADE_HUB
2. If this hub is newer than the other hub, the error is: ERROR__VERSION__MISMATCH_INTERVEHICLE__UPGRADE_OTHER_HUB
3. Stop sending Hub2Hub data.
4. Stop processing Hub2Hub incoming.

## Testing

Manually switched INTERVEHICLE_API using 1 hub on v21 and 1 hub and 1 bot on v22. Then used 1 hub and 1 bot on v21 and 1 hub on v22. Checked that all bots and hubs gave expected error messages.

**Note**: HubStatus errors are not viewable in JCC. This should be updated so that they are.

### DIFF
--- a/cmake/JaiaVersions.cmake
+++ b/cmake/JaiaVersions.cmake
@@ -81,4 +81,4 @@ set(PROJECT_SOVERSION "1")
 
 # increment when DCCL messages change. See also src/lib/messages/CMakeLists.txt
 # start at 1 as 0 would be used prior to introducing this version (goby::middleware::Group::broadcast_group == 0)
-set(PROJECT_INTERVEHICLE_API_VERSION 21)
+set(PROJECT_INTERVEHICLE_API_VERSION 22)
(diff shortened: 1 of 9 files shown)


ferrom-jaia (approved):
Tested in sim and also Toby went through this with me and showed his test results as well. Will be tested on hardware during beta testing.

PR #1558, opened 2026-05-22
bug/cannot-tap-depth-contour-map-in-exclusion-zone/JAIA-2406

michael-jaia:
### Summary
This pull request allows an operator to tap the depth map in a exclusion zone to see the 3-D plot. We could not previously do this because our click handler only responded to the top level feature (not features layered below). Now, with the Boolean return value of `handleMapFeatureClick` we can control if the handler stops after the top most feature or continues to handle the features below.

### Testing
1. Created an exclusion zone on top of a contour map. Clicked on the exclusion zone above the contour map and the 3-D plot appeared.
2. Clicked on a task packet, waypoint, and rally point on the depth map, no 3-D plot occurred as expected

### Notes
I needed to skip a test because we changed the mechanics of `map.forEachFeatureAtPixel`. Some more time is needed to investigate how to adapt the test to the new structure. The test verifies Bot selection behavior, I tested that manually in this case.

### DIFF
--- a/src/web/components/Map/Map.tsx
+++ b/src/web/components/Map/Map.tsx
@@ -106,38 +106,14 @@ export default function Map() {
             return;
         }
 
-        const feature = map.forEachFeatureAtPixel(event.pixel, (feature: Feature) => feature, {
-            hitTolerance: MAP_FEATURE_HIT_TOLERANCE,
-        });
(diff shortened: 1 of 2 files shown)


ferrom-jaia (approved):
Looks great! Tested in sim

PR #1560, opened 2026-05-26
bug/jcc-edit-toggle-mode-settings-panel/SW-2382

JRoybot:
Bug: Exclusion zone panel incorrectly reset mission edit mode and broke OL/React boundary

Summary

- Removed a post-action enforcement block from JaiaContext.tsx (the reducer) that reset missionIDInEditMode whenever the exclusion zones panel was opened. This was the root cause: it was added it to suppress waypoint placement while the zone panel was visible, but placing policy logic in the reducer broke mission editing — switching panels would silently discard the active mission edit state.
- Restored the OL/React boundary in the exclusion zone drawing flow. The previous implementation stored a React dispatch reference on the ExclusionZoneLayer OL singleton. The drawend handler is now wired in Map.tsx via a proper React closure, keeping OL modules free of React dependencies.
- Fixed waypoint suppression correctly: EXCLUSION_ZONE_DRAWING map mode is now set when the exclusion zones panel opens (consistent with how SURVEY_PLANNING works for the survey panel). handleMapClick returns early in that mode, preventing waypoint placement without touching missionIDInEditMode. Mode is cleared when the panel closes or the user navigates away.

Test plan
 Open exclusion zones panel — verify map clicks do not add waypoints to mission in edit mode
 Click "New Zone" and draw a polygon — verify zone is created and button returns to unpressed state
 Close exclusion zones panel — verify mission edit mode resumes and map clicks add waypoints normally
 Undo after zone creation restores prior state correctly
 Opening/closing zone vertex panel while exclusion zones panel is active does not leave map mode stuck


src/web/components/Map/Map.tsx
```diff
@@ -58,8 +59,34 @@ export default function Map() {
         map.on("click", (event: MapBrowserEvent<PointerEvent>) => {
             handleMapClick(event);
         });
-        exclusionZoneLayer.setDispatch(jaiaDispatch);
         styleControlButtons();
+
```
michael-jaia: Jeff can we place this logic in its own function and then call it here? Just thinking of ways we can keep useEffect less congested

src/web/openlayers/maps/map.ts
```diff
@@ -55,6 +55,14 @@ export const map = new Map({
  * @returns {void}
  */
 export function handleMapModeChange(mapMode: MapModes) {
+    // Always deactivate zone Draw first — re-activated by setExclusionZoneDrawActive(true) if needed.
```
michael-jaia: Could we include this in a section like: `if (mapMode !== MapModes.EXCLUSION_ZONE_DRAWING) {` to match the pattern of the function?

src/web/openlayers/maps/map.ts
```diff
@@ -92,6 +95,23 @@ export function handleMapModeChange(mapMode: MapModes) {
     jaiaGlobal.setMapMode(mapMode);
 }
 
+/**
```
michael-jaia, line 99: Include @params and @returns lines

michael-jaia (changes requested):
Jeff, nice cleanup, left a couple of comments

michael-jaia (approved):
This is just the start, we know we have more to tighten up before next release.

PR #1564, opened 2026-05-27
task/jcc-server-revert-bot-status-calculation

ferrom-jaia:
## Description:

Revert using the current_utime function that included sim warp speed corrections to the previous now_utime function. Interesting results were discovered that made us uncomfortable with moving forward with the change until more testing is completed. Weird status ages were reported. We still utilize the corrected sim method for the active link status ages because the last received times for those are set in the hub manager and need to be corrected when running in sim.

## Tests

* Tested at warp 1 with xbee and wifi comms
* Tested at warp 3 with xbee and wifi comms
* Tested at warp 5 with xbee and wifi comms


marshalllele (approved):
This looks good. Removes any possibility of status age calculation being muddied by active link calculation errors by moving the two pieces to separate functions. Tested in sim.

PR #1568, opened 2026-05-29
task/3.y/init-3.y/SW-2373

tsaubergine:
## Background

Target branch `3.y` is same as release 2.7.0. This branch does the minimum of updates and fixes required to start the rest of the 3.y epic.

### Ubuntu

- Use latest LTS Ubuntu 26.04 "resolute"
- Python 3.14
  - Updated PyPI installed dependencies to latest versions
- Boost 1.90
- Stay on C++23 (nothing critical in C++26; some parts are still unimplemented in Clang 21 - this lets us still test build 3.y on noble if we want).

### GobySoft

- Goby 3.5.1
- DCCL 5.0.0

### Arduino

- `arduino-cli` 1.5.0

### rpicam

- Use latest Debian 13 "trixie"

### Python PyPI

- Adafruit libraries, Flask-Compress, h5py-stubs to latest version in `src/python/requirements.txt`
- Fixed `src/python/update_requirements.sh`, removed duplicate script `src/python/sync_requirements.sh`.

## Packaging

- Use debhelper version 13
- CDBS is gone, remove it for vanilla debhelper

## Other required changes

- `apt-key` is dead, use separate gpg keyrings.
- fix old bug now exposed with new CMake for chaining arduino dependencies (ArduinoCompile.cmake)
- fix `scripts/docker-create-push-for-circleci.sh` script to not have hardcoded Ubuntu version
- goby::zeromq::SimpleThread replaces goby::middleware::SimpleThread in Goby 3.5.1
- default temperature_stdev and salinity_stdev of 0 now create assertion error in std distributions. Changed to default to 1 so when these are unused they don't cause the simulator to crash.
- Fixes to changes in io_context API (Boost ASIO 1.90)
- Fixes to changes in Boost Process API (use v1 API for now)
- Switched to use similar library `pyturf` instead of `turfpy` as this is being updated and `turfpy` has more dependencies (turfpy was requiring a manual install of older scipy which is challenging). This is only used in the `src/python/pyjaia/src/pyjaia/drift_interpolation.py` script
- Fixed duplicate install of JDV to share/jaiabot when all web resources are already installed to share/jaiabot/web.
- `flash-image` package is now `piboot-try` and flash-image machine name changed ("Raspberry Pi" instead of "Raspberry Pi 4B")
- `ntp` package gone, it's now `ntpsec`. Changed to this, which will get updated further with https://jaia-innovation.atlassian.net/browse/SW-2203?atlOrigin=eyJpIjoiMWU5Yjg4ZTFiMzdkNDA5ZmIyZjZmMGI2Y2FjZWJjMmYiLCJwIjoiaiJ9

## Remove old workarounds that are no longer needed

- Remove `optimize=-lto` flag that was working around a bug in Ubuntu 24.04 LLVM
- Remove splines=line in Doxygen generated output now that bug in ortho is fixed.

## Minor 

- Doxygen QUIET=YES to reduce build noise
 
## Testing

- Local simulation works
- [Build pipeline through bundle](https://app.circleci.com/pipelines/github/jaiarobotics/jaiabot/10561/workflows/3cec137c-9aa0-45ad-8beb-7694ba27966a) works (still need AWS Sync updates)

This is good enough for now, let's merge this so we can knock off the rest of the 3.y devel tasks using the 3.y after this merge.


ferrom-jaia (approved):
Looks good to me! Thanks Toby

PR #1585, opened 2026-06-08
Add tooltip to rally point button

JaiaJames:
Adds a "Create Rally Point" tooltip to the rally point button in the map toolbar.


michael-jaia:
James, I am going to close the PR, and it would be good for you to recreate it by making a branch off of 2.y (task/add-button-tool-tips) and then pushing up those changes (you do not need to work off of your fork in this case). It makes it easier for the review to be able to look at your branch in an IDE.

PR #1601, opened 2026-06-15
task/change-exclusion-zone-buffer-to-10-meters

joeymussalli:
This pull request updates the default safety margin for exclusion zone routing from 5 meters to 10 meters. The goal is to ensure that the safety margin used in pathfinding matches the value defined in the bot configuration. 

**Exclusion zone routing improvements:**

* Increased the `DEFAULT_SAFETY_MARGIN_METERS` from 5 to 10 in `exclusion-zone-router.ts` to match the `hibernation_radius` setting in the bot configuration.

**Code consistency and usage:**

* Updated the call to `routeAroundExclusionZones` in `Map.tsx` to use `undefined` for the safety margin argument so that the new default value is applied.

**Testing:**
* Verified in the simulator.

### DIFF
--- a/src/web/components/Map/Map.tsx
+++ b/src/web/components/Map/Map.tsx
@@ -494,7 +494,11 @@ export default function Map() {
             // lat/lon values match those produced by detectMissionReroutes,
             // which also uses the first clean waypoint as origin.
             const firstCleanLoc = waypoints.filter((wp) => !wp.getIsBypass())[0]?.getLocation();
-            const result = routeAroundExclusionZones(miniPlan, 5, firstCleanLoc ?? fromLocation);
+            const result = routeAroundExclusionZones(
+                miniPlan,
+                undefined,
+                firstCleanLoc ?? fromLocation,
+            );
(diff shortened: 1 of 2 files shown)


src/web/components/Map/Map.tsx
```diff
@@ -494,7 +494,11 @@ export default function Map() {
             // lat/lon values match those produced by detectMissionReroutes,
             // which also uses the first clean waypoint as origin.
             const firstCleanLoc = waypoints.filter((wp) => !wp.getIsBypass())[0]?.getLocation();
-            const result = routeAroundExclusionZones(miniPlan, 5, firstCleanLoc ?? fromLocation);
+            const result = routeAroundExclusionZones(
+                miniPlan,
+                undefined,
```
michael-jaia: I think we should pass DEFAULT_SAFETY_MARGIN_METERS here for clarity

michael-jaia (changes requested):

michael-jaia:
We are going to remove the buffer zone in a future PR

PR #1609, opened 2026-06-26
bug/align-segment-speeds-with-slider

michael-jaia:
### Summary
This pull request enforces the following rule: _when the operator sends a command, whatever the speed in the speed slider says is the speed that should be sent to all of the segments_

### Testing
**Test 1**
* Save a mission set with a speed of 1
* Save a mission set with a speed of 2
* Combine them
* Set the speed slider to 3, hit play

_Transit speeds in each segment should be set to 3_

**Test 2**
* Create a survey with SRP and speed 2
* Create a survey with SRP and speed 3
* Combine them
* Set the speed slider to 1, hit play

_Transit speeds in each segment should be set to 1 (SRP speed parameters are not changed)_

* Change the mission speed slider to 3, hit play again

_Transit speeds in each segment should be set to 3 (SRP speed parameters are not changed)_


ferrom-jaia (approved):
Looks good!

Tested in sim.

PR #1613, opened 2026-06-29
task/sensor-calibration

marshalllele:
Adds our sensor calibration scripts to the repo so users/calibrators do not need to copy the scripts onto the vehicle.


ferrom-jaia (approved):
Looks good, we have been using these without them being in the repo.

PR #1617, opened 2026-07-07
feature/sensor-data-filtered-algorithm

joeymussalli:
## Summary
This PR introduces a live Hampel filter to detect and flag outliers in sensor data (e.g., salinity, temperature, pH) during missions.

Sensor data occasionally produces spikes that do not reflect the actual environment. The filter uses a sliding window to calculate the median and Median Absolute Deviation (MAD), flagging outliers without allowing the spikes to skew the underlying baseline. The filter only *reports* outliers; raw sensor values are never altered. 

### Algorithm (`src/lib/utils/hampel_filter.h`):

- Implemented the `HampelFilter` class using a sliding window.
- Added configurable parameters `window_size` (default: `7`) and the `num_mads` threshold (default: `3.0`).
- Added logic to calculate the median and scaled MAD (`1.4826 * MAD`) to approximate standard deviation.

### Unit Testing (`src/test/utils/test.cpp`):

- Added Boost unit tests to verify the filter's behavior across multiple cases. These cases include:
  - Sliding window initiation phase. 
  - No false positives. 
  - Outlier detection.
  - Reaction to persistent level shifts.
  - Handling of data windows with zero spread.

### Documentation (`src/doc/markdown/page015_sensor_data_filtering.md`):

- Added a documentation page in `docs.jaia.tech` that contains the math of the algorithm, configuration options, and edge cases.

### DIFF
--- a/src/doc/markdown/page015_sensor_data_filtering.md
+++ b/src/doc/markdown/page015_sensor_data_filtering.md
@@ -1,9 +1,8 @@
 # Sensor Data Filtering (Hampel Filter)
 
-Sensor streams occasionally produce spurious readings ("spikes") that do not
-reflect the real environment. The **Hampel filter** flags these outliers live
-during a mission, so that users can rely on a cleaned data stream while the raw
-value is still preserved.
+Sensor streams occasionally produce spikes that do not reflect the real environment. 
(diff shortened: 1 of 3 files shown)


ferrom-jaia (approved):
Looks good! 

Ran the unit tests and all passed.

PR #1618, opened 2026-07-07
feature/hub-add-system-time/SW-2454

brownc-jaia:
Added system time (UTC) to hub quick look data using HubStatus time.

Testing:
(On the JCC)
1. Open the Hub interface by selecting the _Hub 1_ button
2. Select _Quick Look_
3. Verify "System Time (UTC)" is present and contains the current time in UTC

### DIFF
--- a/src/web/data/hubs/hub.ts
+++ b/src/web/data/hubs/hub.ts
@@ -103,6 +103,20 @@ export default class Hub {
         this.statusAge = statusAge;
     }
 
+    getSystemTime() {
+        const date = new Date();
+
+        const options = {
+            timeZone: "UTC",
+            hour12: true,
+            hour: "numeric",
+            minute: "2-digit",
+            second: "2-digit",
+        };
+
+        return date.toLocaleTimeString("en-US", options);
+    }
+
     /**
      * Determines if the hub has lost comms with the client
      *
--- a/src/web/data/hubs/hub.ts
+++ b/src/web/data/hubs/hub.ts
@@ -21,6 +21,7 @@ export default class Hub {
     private linuxHardwareStatus: LinuxHardwareStatus;
     private botOffload: BotOffloadData;
     private statusAge: number;
+    private systemTime: string;
 
     constructor() {
         // Init base sensors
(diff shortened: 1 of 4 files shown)


src/web/data/hubs/hub.ts
```diff
@@ -103,6 +103,17 @@ export default class Hub {
         this.statusAge = statusAge;
     }
 
+    getSystemTime() {
+        const date = new Date();
```
ferrom-jaia: This is probably useful to get as well, but this is getting the time from the tablet or computer that is being used. 

The hub should have a time field in the hub status that we can use to convert to UTC time. 

We can talk through this.

ferrom-jaia (changes requested):
First pull request at Jaia!!

Thanks Cam!

src/web/utils/conversions.ts
```diff
@@ -35,6 +35,24 @@ export function timestampToISOString(tMicroseconds: number) {
     return new Date(tMicroseconds / 1000).toISOString();
 }
 
+/**
+ * Converts a Unix timestamp (microseconds) to an locale time string
+ *
+ * @param {number} tMicroseconds Unix timestamp in microseconds since Unix epoch
+ * @returns {string} Time string in UTC time format
+ */
+export function timestampToLocaleTimeString(tMicroseconds: number) {
```
ferrom-jaia, line 44: What do you think about this as a name:

timestampToUTCTimeString 

Since it is always going to be UTC time.

ferrom-jaia (commented):
Looks good! Just one comment.

Tested in sim.

Used this website for comparison: https://www.timeanddate.com/worldclock/timezone/utc

src/web/utils/conversions.ts
brownc-jaia: That's what I was thinking initially but turns out `date.toUTCString()` and `date.toLocaleTimeString()` are two different functions returning different values:
`date.toUTCString()`: Day, DD Mon YYYY HH:mm:ss GMT
`date.toLocaleTimeString()`: HH:mm:ss

ferrom-jaia (approved):

PR #1619, opened 2026-07-07
feature/python-lib-zeromq-connection

ferrom-jaia:
## Description

Using connection to zeromq instead of using udp connection for rpm data by using a minimal native Python client for the goby3 ZeroMQ interprocess transport (goby::zeromq::InterProcessPortal).

## Test

Tested in sim. Used liaison to confirm rpm information is being published.

### DIFF
--- a/config/gen/bot.py
+++ b/config/gen/bot.py
@@ -279,8 +279,6 @@ elif common.app == 'jaiabot_health':
                                      fleet_id=fleet_index,
                                      app_block=app_common,
                                      interprocess_block = interprocess_common,
-                                     bind_port=common.udp.motor_cpp_udp_port(),
-                                     remote_port=common.udp.motor_py_udp_port(bot_index),
                                      # do not power off or restart the simulator computer unless we're a VirtualFleet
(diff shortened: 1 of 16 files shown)


edsanville (dismissed):
This is a great idea, and the code looks good!

We could use this method to replace:
* jaiabot_udp_gateway
* jaiabot_web_portal

The python sensor apps and the JCC/JED web server could communicate directly with goby using this method.  The only drawback would be that the web server couldn't be located on another machine... but I'm guessing that's not a use case we care about.

tsaubergine (commented):
I'm not a big fan as this adds a lot of autogen code. I would rather finish the real Python3 bindings for Goby3 (like I did for the Julia language). Happy to do this soon if it's a priority.

ferrom-jaia:
> I'm not a big fan as this adds a lot of autogen code. I would rather finish the real Python3 bindings for Goby3 (like I did for the Julia language). Happy to do this soon if it's a priority.

I forgot about that project. Yes, that would be the better way. We can discuss this the next time we meet.

PR #1621, opened 2026-07-09
feature/jcc-sliders-instead-of-input-only/SW-2323

joeymussalli:
## Summary
The Mission Repeats field was a normal text field and the number fields in the Survey Planner and Waypoint task panels had no increment/decrement controls on a tablet. Initially I tried to restore the browser's standard `type="number"` arrow buttons, but those are mouse exclusive and aren't rendered on a tablet. 

The current version of the PR instead uses a `JaiaNumberInput` component with its own up/down buttons that work the same on laptop and tablet. These buttons are applied to the three fields mentioned above. 

## Changes Made

`src/web/components/JaiaNumberInput/` (new files)
New component `JaiaNumberInput` renders a `type="number"` input with a up/down button pair aligned to the right side of the field. It also exports `StepperButtons` on its own so the buttons can be reused elsewhere. The `.less` formats the buttons and lets them keep the surrounding panel's text color.

`src/web/components/MissionsPanel/MissionsList/MissionsList.tsx`
Mission Repeats is now `type="number"` with a minimum value of 1. 

`src/web/components/MissionsPanel/MissionsList/MissionsList.less`
Hides the browser's standard arrows on the Repeats field so desktop doesn't show the browser arrows and the custom buttons.

`src/web/components/SurveyPlanner/SurveyPlanner.tsx` / `.less`
The four inputs (lanes, bots, lane spacing, point spacing) now use `JaiaNumberInput` so the arrows are displayed. Also altered formatting so the values remain next to the buttons.

`src/web/components/WaypointPanel/TaskParameters/TaskParameters.tsx` / `.less`
All task related number inputs (depths, times, heading, speed, etc.) now use `JaiaNumberInput`.

## Testing
Verified in sim on both desktop and the tablet mode found in the device toolbar.

### DIFF
--- a/src/web/components/JaiaNumberInput/JaiaNumberInput.less
+++ b/src/web/components/JaiaNumberInput/JaiaNumberInput.less
@@ -1,0 +1,41 @@
+.jaia-number-input {
+    position: relative;
+    display: flex;
+}
+
+.jaia-number-input > input {
+    width: 100%;
+}
+
+.jaia-stepper {
+    display: flex;
+    flex-direction: column;
+    width: 16px;
+}
+
+.jaia-number-input .jaia-stepper {
+    position: absolute;
+    top: 0;
+    right: 2px;
+    bottom: 0;
+}
+
+.jaia-stepper-button {
+    all: unset;
+    flex: 1;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    cursor: pointer;
+    color: inherit;
+}
+
+.jaia-stepper-button:disabled {
+    opacity: 0.4;
+    cursor: default;
+}
+
+.jaia-stepper-button svg {
+    width: 14px;
+    height: 14px;
+}
--- a/src/web/components/JaiaNumberInput/JaiaNumberInput.less
+++ b/src/web/components/JaiaNumberInput/JaiaNumberInput.less
@@ -17,13 +17,13 @@
 .jaia-stepper {
     display: flex;
     flex-direction: column;
-    width: 16px;
+    width: 15px;
 }
 
 .jaia-number-input .jaia-stepper {
     position: absolute;
     top: 0;
-    right: 2px;
+    right: 3px;
     bottom: 0;
 }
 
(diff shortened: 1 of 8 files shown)


JRoybot (dismissed):
The changes look good.  Was a bit confused by the title but after reading the Jira ticket see why.  Adding Jira suffix to PR title for consistency.

ferrom-jaia:
Hey Joey,

Was this tested on a tablet?

I was using the web developer tools (F12) and I put it into tablet mode. I did not see the buttons for increase and decreasing the repeats.

joeymussalli:
No it wasn't -- I assumed the buttons would carry over. Turns out the numerical field format I was using doesn't get rendered on a tablet. To get an equivalent effect on the tablet I'll need to make a new component. Will get started on that now.

src/web/components/JaiaNumberInput/JaiaNumberInput.less
```diff
@@ -0,0 +1,48 @@
...
+.jaia-number-input input[type="number"]::-webkit-outer-spin-button {
+    appearance: none;
+    -webkit-appearance: none;
+    margin: 0;
+}
+
+.jaia-stepper {
+    display: flex;
+    flex-direction: column;
+    width: 16px;
```
JRoybot: Twomey has commented on a few of my reviews that styles should use 3 px increments, for consistency.  Maybe these could be 15 & 3 instead of 16. 2, & 14 ?

src/web/components/WaypointPanel/TaskParameters/TaskParameters.less
```diff
@@ -6,8 +6,7 @@
 }
 
 .task-parameters input[type="number"] {
-    appearance: auto;
-    -webkit-appearance: auto;
+    padding-right: 20px;
```
JRoybot: multiple of 3 ?

JRoybot (changes requested):
Reviewed changes, tested in browser set to andoid and ios tablets.  Left a few minor comments re style.

JRoybot (approved):

PR #1623, opened 2026-07-15
feature/jcc-add-task-packet-filtering

joeymussalli:
image

## Summary

This re-adds the ability to filter which task packets show up on the JCC map. There's now a filter icon under Settings → Map Layers → Measurements that opens a panel where you can pick a date range, optionally narrow by mission name, and hit Search to get a list of matching missions. Packets without a mission name get grouped under "Unnamed". From there you can check/uncheck missions to show or hide them, and drag a time slider to narrow things down further.

The depth contours regenerate on search, mission selection, and slider release to match whatever is filtered. New packets that match the filter still show up live, and there's a Clear Filter button to get back to the normal view. Reloading the page clears the filter. 

## Changes

New files:
- `data/task_packets/task-packet-filter.ts` - keeps track of the current filter (date range, selected missions, slider window), builds the mission list for the results panel, and is what the map layers check packets against
- `components/TaskPacketFilterPanel/` - the panel itself

Frontend changes:
- `LayerSwitcherMenu.tsx` and `SettingsPanel.tsx` - the filter icon and rendering the panel next to settings
- The dive, drift, and excluded-task-packets layers now check packets against the filter before drawing them
- `contour-layer.ts` - when a filter is active, generates the contour map using the filtered packets instead of the default window
- `handler-utils.ts` - added `syncTaskPacketMarkerLayers()` which redraws only the task packet markers with no network calls, so dragging the slider is instant
- `jcc/polling.ts` - when a filter is active, polls for the selected date range instead of the default one so the map stays live
- `context-types.ts`, `jaia-actions.ts`, `panel-handlers.ts`, `action-configs.ts`, `data-model-handers.ts`, `selection-handlers.ts` - open/close state for the panel, closes when you leave settings
- `protobuf-types.ts` - added the existing `mission_name` field (proto field 72) to the `TaskPacket` interface
- `utils/style.ts` - theme for the panel

Backend:
- `server/jaia_portal.py`, `server/app.py` - new `POST /jaia/v0/depth-contours` endpoint. It takes the already filtered packets and runs them through the existing `taskPacketsToColorMap`. 
- `utils/jaia-api.ts` - added the API call that sends the filtered packets to the new contour endpoint and returns the generated contours

## Testing

Tested manually using the simulator with named and unnamed missions: 

- Date range narrows the mission name options, and "Unnamed" shows up for nameless packets
- Search lists the missions with everything checked by default
- Checking/unchecking missions adds/removes their packets, deselecting all hides everything
- Dragging the slider filters markers instantly, and releasing it regenerates the contour
- As a mission runs, new packets keep showing up and the slider's upper bound follows along
- Clear Filter goes back to the normal view, and reloading clears the filter

## Design notes

This is purely a client-side view filter. It never touches the server's per-packet `included` flag, so nothing is destructive and it doesn't affect other users. The only backend change is the read-only contour endpoint.

For the contours, the user just sends the packets it already filtered to the existing generator, so the contour always matches the markers exactly. Contours only recompute on search/selection/slider release rather than every drag frame to minimize the compute needed from the Pi. That cadence is no heavier than the current behavior of regenerating on every new packet.


ferrom-jaia:
Nice first pass Joey!

image

What do you think about adding another accordion called Task Packet Filter above Map Layers? We used to have it here previously and I am thinking it will help conserve space for tablet users.

image

When dragging the time slider the data and time are off the screen.

I also there is a bug with selecting the dates:

image

I have the end data set to the 31st but it shows the 17th.

Another bug I noticed is sometimes you can get into a bad state where no taskpackets show:

https://github.com/user-attachments/assets/37d47b71-2540-4a38-a737-d7072f402e2c

joeymussalli:
image

Filter panel now lives in an accordion above map layers in the settings panel. The bugs found by @ferrom-jaia should be ironed out.

src/web/components/TaskPacketFilterPanel/TaskPacketFilterPanel.less
```diff
@@ -0,0 +1,84 @@
...
+.task-packet-filter-dates {
+    display: flex;
+    gap: 12px;
+}
+
+.task-packet-filter-dates label {
+    display: flex;
+    flex-direction: column;
+    flex: 1;
+    gap: 4px;
```
JRoybot: can we make all the px values multiples of 3

src/web/components/SettingsPanel/SettingsPanel.tsx
```diff
@@ -109,6 +110,21 @@ export default function SettingsPanel() {
...
+                        slotProps={{ transition: { unmountOnExit: true } }}
+                    >
+                        <AccordionSummary
+                            expandIcon={<ExpandMoreIcon />}
+                            className="accordion-summary"
+                        >
+                            <Typography>Task Packet Filter</Typography>
+                        </AccordionSummary>
+                        <AccordionDetails>
+                            <TaskPacketFilterPanel />
```
JRoybot: Since this is no longer a panel, I think it it should just be named TaskPacketFilter.  That will be more consistent with the other accordion details

src/web/components/TaskPacketFilterPanel/TaskPacketFilterPanel.less
JRoybot, line 1: See comment in SettingsPanel.tsx regarding name change

src/web/components/TaskPacketFilterPanel/TaskPacketFilterPanel.tsx
JRoybot, line 1: See comment in SettingsPanel.tsx regarding name change

src/web/components/TaskPacketFilterPanel/TaskPacketFilterPanel.tsx
```diff
@@ -0,0 +1,503 @@
...
+        end: getHTMLDateString(new Date()),
+    };
+}
+
+/**
+ * Filter panel that sits beside the settings panel and lets the operator filter which task packets are shown on the map.
+ */
+export default function TaskPacketFilterPanel() {
+    // Search setup state
+    const [startDateStr, setStartDateStr] = useState(() => {
```
JRoybot: Prefer to see these arrow functions defined as const instead of in the useState call.  Mostly for readability and consistency with other components

src/web/components/TaskPacketFilterPanel/TaskPacketFilterPanel.tsx
```diff
@@ -0,0 +1,503 @@
...
+        taskPacketFilter.getSliderLowerUtime(),
+        taskPacketFilter.getSliderUpperUtime(),
+    ]);
+    const [sliderValue, setSliderValue] = useState<[number, number]>(() => [
+        taskPacketFilter.getSliderLowerUtime(),
+        taskPacketFilter.getSliderUpperUtime(),
+    ]);
+
+    // Latest values for use inside the live interval.
+    const missionsRef = useRef(missions);
```
JRoybot: Want to make sure I understand why the useRef are needed in addition to the useState hooks above.  Do these change independent of rendering?  Please explain their use.

src/web/components/TaskPacketFilterPanel/TaskPacketFilterPanel.tsx
```diff
@@ -0,0 +1,503 @@
...
+    const lastVersionRef = useRef(-1);
+    // On reopen the filter is already committed, so skip the first commit run.
+    const skipInitialCommitRef = useRef(taskPacketFilter.isActive() && selectedKeys.size > 0);
+    // Distinguishes the first fetch (mount/reopen) from a later date-range change.
+    const isInitialFetchRef = useRef(true);
+    missionsRef.current = missions;
+    selectedKeysRef.current = selectedKeys;
+
+    /**
+     * Builds the "yyyy-mm-dd hh:mm" query strings for the selected date range.
```
JRoybot: should include @returns for consistency

src/web/components/TaskPacketFilterPanel/TaskPacketFilterPanel.tsx
```diff
@@ -0,0 +1,503 @@
...
+     */
+    const buildQueryStrings = () => {
```
JRoybot: Can these const helper functions be moved out of panel function so they are not built on each render?  If there are enough of them and they are pure typescript (no React), could be put in a helper .ts file.

src/web/components/TaskPacketFilterPanel/TaskPacketFilterPanel.tsx
```diff
@@ -0,0 +1,503 @@
...
+            console.error(error);
+            setMissions([]);
+        } finally {
+            setIsLoading(false);
+        }
+    };
+
+    // Refresh the mission list on date change, debounced. The first run (mount/reopen)
+    // only refreshes the list; later changes re-apply the window (see fetchMissions).
+    useEffect(() => {
```
JRoybot: prefer to see all hooks at the top of the file.  This does not break any hook rules because it is top level, just looking for more consistency and readability.

Also prefer to see the helper arrow functions declared as const functions

src/web/components/TaskPacketFilterPanel/TaskPacketFilterPanel.tsx
JRoybot, line 1: commented in a few places below, in general we like to put non-React helper functions in a .ts file to keep the tsx code cleaner.  It also provides easier approach to unit testing those functions.  Any functionality that is non-trivial should be unit tested so our CI will catch anything that breaks down the road.  (note, most of the AI coding agents are very good at writing unit tests for non react code.)

src/web/components/TaskPacketFilterPanel/TaskPacketFilterPanel.tsx
```diff
@@ -0,0 +1,503 @@
...
+                    setSliderBounds(bounds);
+                    if (sliderValue[0] === 0 && sliderValue[1] === 0) {
+                        setSliderValue(bounds);
+                    }
+                }
+                return;
+            }
+
+            // Date range changed: re-apply it to the map, keeping the current selection.
+            taskPackets.setIncludedTaskPackets(included);
```
JRoybot: should not be modifying the data model directly, should go through the reducer so it triggers updates to other components that may reference it.  General design pattern in React is changes to the data model should be done by handlers triggered by actions sent to the reducer.  Example src/web/context/handlers/waypoint-handlers.ts

src/web/components/TaskPacketFilterPanel/TaskPacketFilterPanel.tsx
```diff
@@ -0,0 +1,503 @@
...
+            taskPacketFilter.setSearchWindow(
+                new Date(`${startDateStr}T00:00:00`),
+                new Date(`${endDateStr}T23:59:59`),
+            );
+            const bounds = computeBounds(summaries, selectedKeysRef.current);
+            setSliderBounds(bounds);
+            setSliderValue(bounds);
+            taskPacketFilter.setSliderWindow(0, 0);
+            taskPacketFilter.setAutoFollowUpper(true);
+            syncTaskLayers();
```
JRoybot: should not be initiating OL changes in a React panel directly, this should only be done in an action handler

src/web/components/TaskPacketFilterPanel/TaskPacketFilterPanel.tsx
JRoybot, line 1: Made a few comments below but this is really a component wide comment.

This panel is directly modifying data in the data model (taskpackets etc) and triggering rendering of OpenLayers.  This is not consistent with the general patterns used elsewhere.  Modifying the data model should be done in a handler called by the reducer, triggered by an action.  This keeps all the components in sync and reduces unneeded rendering.  It also makes the code easier to debug since all the code uses similar patterns and all the core data changes are in well defined scope.

Below documents the approach, Twomey may be able to point you to more or documentation

https://jaia-innovation.atlassian.net/wiki/spaces/PD/pages/1715798017/Jaia+Command+Control+Documentation

src/web/context/handlers/handler-utils.ts
```diff
@@ -97,3 +97,17 @@ export function syncTaskLayers() {
...
+
+/**
+ * Repaints only the per-packet task layers (dive, drift, excluded) from the current
+ * client-side data. Unlike syncTaskLayers() this does NOT refresh the contour layer,
+ * so it makes no network request. Used by the task packet filter's time slider so
+ * dragging is instant.
+ *
+ * @returns {void}
+ */
+export function syncTaskPacketMarkerLayers() {
```
JRoybot, line 109: handler functions should only be called via the reducer not directly by any React panels.  And as stated above, React panels should not trigger Open Layer rendering directly.

src/web/data/task_packets/task-packet-filter.ts
```diff
@@ -0,0 +1,162 @@
...
+}
+
+/**
+ * Groups task packets into mission summaries sorted by start time. Packets with no
+ * mission_name are grouped under UNNAMED so they aren't missed.
+ *
+ * @param {TaskPacket[]} taskPackets Packets to summarize
+ * @returns {MissionSummary[]} Summaries sorted by startTime ascending
+ */
+export function buildMissionSummaries(taskPackets: TaskPacket[]): MissionSummary[] {
```
JRoybot: unit test of this function would be good to verify the edge cases and help CI catch any future bugs

src/web/components/TaskPacketFilterPanel/TaskPacketFilterPanel.tsx
```diff
@@ -0,0 +1,503 @@
...
+        taskPacketFilter.clear();
+        setSelectedKeys(new Set());
+        setHasSearched(false);
+        setNameFilter([]);
+        setMissions([]);
+        setSliderBounds([0, 0]);
+        setSliderValue([0, 0]);
+        setStartDateStr(defaultDateRange.start);
+        setEndDateStr(defaultDateRange.end);
+        pollTaskPackets();
```
JRoybot: Calling the polling function from a React Panel should not be done, needs to either be done periodically or triggered by sending an action to the reducer.

JRoybot (changes requested):
Read through the code.  Left a lot of comments related to the general structure of this functionality. We should maintain the design patterns used in the rest of the code to keep things consistent, maintainable and easier to debug/test.  We do not want any React components directly modifying data used across the app.  The data model is shared to React components via JaiaContext and updates are done by handlers triggered by actions being sent to the jaiaReducer.  

Realize this requires significant rework of the functionality but is important for consistency and maintainability.

Below documents the approach, Twomey may be able to point you to more or documentation

https://jaia-innovation.atlassian.net/wiki/spaces/PD/pages/1715798017/Jaia+Command+Control+Documentation

src/web/components/TaskPacketFilterPanel/TaskPacketFilterPanel.tsx
joeymussalli: They let the live timer read the newest values without restarting itself on every render. They don't trigger re-renders. I'm going to move the logic into a handler after getting your comments, so most of them will go away.

src/web/components/TaskPacketFilterPanel/TaskPacketFilterPanel.tsx
joeymussalli: Yeah that sounds good -- I'll make a .ts file.

src/web/components/TaskPacketFilterPanel/TaskPacketFilterPanel.tsx
JRoybot: is there a performance issue or other reason we cannot leave the poll all task packets and then filter them when rendering?

src/web/components/TaskPacketFilterPanel/TaskPacketFilterPanel.tsx
joeymussalli: Claude to the rescue!

src/web/components/TaskPacketFilterPanel/TaskPacketFilterPanel.tsx
JRoybot: great

src/web/components/TaskPacketFilter/TaskPacketFilter.tsx
```diff
@@ -1,15 +1,26 @@
-import { useEffect, useRef, useState } from "react";
+import { useContext, useEffect, useRef, useState } from "react";
 
-import { syncTaskLayers, syncTaskPacketMarkerLayers } from "../../context/handlers/handler-utils";
-import { pollTaskPackets } from "../../jcc/polling";
+import { JaiaDispatchContext } from "../../context/JaiaContext";
+import { JaiaActions } from "../../context/jaia-actions";
 import { taskPackets } from "../../data/task_packets/task-packets";
```
JRoybot: should replace these references with JaiaContext

src/web/components/TaskPacketFilterPanel/TaskPacketFilterPanel.tsx
joeymussalli: The mission name and time slider filters actually already filter the polled packets at render. Only the date range filter has to poll the server because the default poll doesn't return older missions (meaning they would be missing from the filter settings if the date range were to be shifted earlier).

src/web/components/TaskPacketFilter/task-packet-filter.ts
JRoybot, line 1: not a big deal but if you look through the other components that have helpers in ts files they just use the same name as the tsx code but are in dashed lower case instead of camel case.  Small detail but keeps a consistent feel.

src/web/components/TaskPacketFilter/TaskPacketFilter.tsx
```diff
@@ -0,0 +1,433 @@
...
+import Slider from "@mui/material/Slider";
+import Button from "@mui/material/Button";
+import CircularProgress from "@mui/material/CircularProgress";
+
+import "./TaskPacketFilter.less";
+
+const SEARCH_DEBOUNCE_TIME = 400; // milliseconds
+
+/**
+ * Filter panel that sits beside the settings panel and lets the operator filter which task packets are shown on the map.
```
JRoybot: Update comment to match new layout

src/web/components/TaskPacketFilterPanel/TaskPacketFilterPanel.tsx
JRoybot: would still like to see the arrow functions declared as const to make things easier to read.  Some of the useEffect statements span 10s of lines.

src/web/data/task_packets/task-packet-filter.ts
```diff
@@ -0,0 +1,162 @@
+import { TaskPacket } from "../../types/protobuf-types";
+
+// Grouping key for task packets that have no mission_name.
+export const UNNAMED_MISSION_KEY = "__UNNAMED__";
+
+export interface MissionSummary {
+    key: string;
```
JRoybot, line 7: Just a question... Why use mission name as keys instead of missionIds?  Most places we manage mission information we use the IDs which are always present, unlike names.

src/web/data/task_packets/task-packet-filter.ts
```diff
@@ -0,0 +1,162 @@
...
+}
+
+/**
+ * Returns the grouping name for a task packet.
+ *
+ * @param {TaskPacket} taskPacket Packet to key
+ * @returns {string} Mission name or UNNAMED_MISSION_KEY
+ */
+export function missionKeyOf(taskPacket: TaskPacket): string {
+    return taskPacket.mission_name ? taskPacket.mission_name : UNNAMED_MISSION_KEY;
```
JRoybot: the only other reference to UNNAMED_MISSION_KEY is in the unit test.  could this be simplified by using IDs instead of names?

src/web/jcc/polling.ts
```diff
@@ -112,6 +126,49 @@ export async function pollTaskPackets() {
...
+ * Fetches task packets for the active filter window (or the server default when no filter
+ * is active).
+ *
+ * @returns {Promise<{ result: { included: TaskPacket[]; excluded: TaskPacket[] } }>} Response
+ */
+export async function fetchTaskPacketsForWindow() {
+    const startDate = taskPacketFilter.getStartDate();
+    const endDate = taskPacketFilter.getEndDate();
+    if (taskPacketFilter.isActive() && startDate && endDate) {
+        return jaiaAPI.getTaskPackets(
```
JRoybot: Why are we using jaiaAPI here when the rest of this module is using the REST API ?  might want to check with Matt or Twomey.  I thought were we migrating away from jaiaAPI

src/web/openlayers/layers/vector/contour-layer.ts
```diff
@@ -2,17 +2,50 @@ import JaiaVectorLayer from "./jaia-vector-layer";
...
     constructor() {
         super(LayerTitles.CONTOUR_LAYER, layersZIndexes.get(LayerTitles.CONTOUR_LAYER));
     }
 
     override updateFeatures() {
-        jaiaAPI
-            .getDepthContours()
+        const isFiltered = taskPacketFilter.isActive();
+        const includedTaskPackets = isFiltered
+            ? taskPacketFilter.filter(taskPackets.getIncludedTaskPackets())
```
JRoybot, line 23: I see this pattern `taskPacketFilter.filter(taskPackets.getIncludedTaskPackets())` in all of the layers displaying task packets in open layers.  However `src/web/components/DataOffloadPanel/DataOffloadPanel.tsx` & `src/web/components/DepthMap3D/depth-map-3D.ts` call `taskPackets.getIncludedTaskPackets()` without applying the filter.  Is that intentional?  Shouldn't the filter affect all consumers of task packets?  That said if the filter should be applied universally, the logic would be simpler if the filter call was embedded in `getIncludedTaskPackets()` instead of all the individual consumers.

src/web/openlayers/layers/vector/excluded-task-packets-layer.ts
```diff
@@ -25,7 +26,7 @@ class ExcludedTaskPacketsLayer extends JaiaVectorLayer {
     override updateFeatures() {
         let source = this.getVectorLayer().getSource();
         source.clear();
-        for (const taskPacket of taskPackets.getExcludedTaskPackets()) {
+        for (const taskPacket of taskPacketFilter.filter(taskPackets.getExcludedTaskPackets())) {
```
JRoybot, line 29: same comments about the use `getIncludedTaskPackets` apply here

JRoybot (commented):
Nive Job incorporating JaiaContext into this!  I left some comments and questions.  I am going to follow up after building and testing the branch but wanted to get these to you as soon as I could.

src/web/data/task_packets/task-packet-filter.ts
JRoybot: Also, I think we need to be careful to not mix up our naming conventions.  Mission Sets have names assigned to them, Missions do not.  I think in the context of this entire function the word mission is used when it is actually referring to Mission Sets.

If we are actually filtering on Mission Sets, we should change the labels in the UI and in the code to reflect that.
If not, how are names being used to select individual missions??

joeymussalli: The decision to use mission name came from the ticket. I assume it is because mission name is easier to interpret than a mission ID.

joeymussalli:
> Trying to exercise the Mission Name filter and am not sure how it is intended to work. As indicated in other comment, I do not know how an individual mission would get a name. I assigned a name to the Mission Set and saved it so I know it exists in the data model. However the filter only shows "Unnamed" in the drop down list. Please explain how this feature works.

From my understanding whatever text is in the highlighted field from image 1 then becomes the mission name for any task packets produced by completed missions. 

Screenshot 2026-07-22 110107

image

src/web/data/task_packets/task-packet-filter.ts
ferrom-jaia: We are filtering on mission names because that is what the user inputs for the mission set. Each bot then gets that mission set name for its mission.

The task packets belong to a mission set though, so I am on board with mission set name.

ferrom-jaia: Customers will not know what a mission id is since this gets created in the background

src/web/data/task_packets/task-packet-filter.ts
ferrom-jaia: We can always revisit this but for the normal customer I do not think they will know how to interpret the mission id. It will also push them to create mission set names going forward. (Or maybe we add a way to remind them)

src/web/jcc/polling.ts
ferrom-jaia: The jaiaAPI is just an interface to the REST API version 0. 

We do have a V1 but I am not aware that we are using it currently for the JCC. There is a 3.y task to transition the JCC from using V0 to V1. 

Customers that interact with our system not through the JCC use V1.

src/web/openlayers/layers/vector/contour-layer.ts
joeymussalli: Yes that's intentional, though now that you point it out I will change the 3D map to include only filtered packets. When the user offloads the data they should be given all the loaded task packets, not just the filtered ones. 

In regard to putting the filter call inside `getIncludedTaskPackets()` -- doing so would only give the filter the filtered task packets, which would not suit the desired behavior. No task packet data should be filtered before entering the filter itself.

joeymussalli:
> It seems that if I apply a mission name filter and then clear it, the drop down list for the mission name filter changes from Unnamed to "no options" I had run 2 mission sets both with names at that point.

You would need to run search again before any mission set names appear in that dropdown.

joeymussalli:
> It seems the task packets shown on the map update immediately when the date filters are changed or when the time bar is trimmed. This makes it confusing what the "Run Search" is actually doing. (compounded because I am not sure what is being used as "mission name".

This is valid -- I originally had the filter not update upon date range changes and that seems to be more intuitive than the current logic. Will revert.

src/web/data/task_packets/task-packet-filter.ts
JRoybot: comment was aimed at simplifying the logic, not to change what is presented to user.  not a big deal

src/web/jcc/polling.ts
JRoybot: All of the other polling in this file does a direct fetch using a URL.  This stood out because it added another module to import.

src/web/openlayers/layers/vector/contour-layer.ts
JRoybot: agreed

src/web/jcc/polling.ts
ferrom-jaia: Interesting, I guess it really is just a design decision then. Do we use jaiaAPI anymore? Is it better to move the translation that interacts with the REST API in a separate file or just put the handling here. I will defer to you on this Jeff.

src/web/data/task_packets/task-packet-filter.ts
ferrom-jaia: Oh gotcha, I see what you mean, it might be confusing on the server side of things since we have mission id and mission name

src/web/data/task_packets/task-packet-filter.ts
JRoybot: Understood.  My suggestion was more aimed at using a string that could be undefined vs a number that will always be there when creating the Set.  Code can always get the name if it has ID.  No big deal.

src/web/jcc/polling.ts
JRoybot: There are lots of places in JCC that sill use `jaiaAPI`.  Twomey decided to not use it here, would have to defer to him the reasons.

That said, I would prefer to see this use the same method all the other polling in the file uses.  The new code replaced this 
`const taskPacketRes = await fetch(TASK_PACKET_URL);`

couldn't we just build the URL with the parameters used in the call to `jaiaAPI`?

I wouldn't put the logic anywhere else, just suggesting keeping it consistent with the rest of the file.

ferrom-jaia: Yes, we can absolutely do that.

joeymussalli:
> Clear filter should eliminate all filtering it seems the date filters are always applied and when Clear Filter is used the dates get reset but the reset dates do not include all of the available task packets. (see image above, the task packets from 6/3 get filtered out after pressing "Clear Filter"

Clearing the filter currently reverts the map view to what was the standard before this PR of the last 14 hours. If you think it's more intuitive I can leave the date range unchanged when clear filter is press.

joeymussalli:
> the text at the top of the accordion "Show only the task packets from specific mission sets on the map." is misleading, especially since mission set name is optional. I think it would be clearer to simply state "Filter task packets displayed on the map"

Agree. I will change this.

joeymussalli:
> I think the use of the Run Search button is still confusing. The only buttons are Run Search and Clear Filter. I think it would be much clear if it they were "Apply Filter" and "Clear Filter" and that filtering should only be done when the user hits "Apply Filter". I also think they should both be at the bottom of the accordion. I am not sure why we need to run a search to use the mission set name filter. If the list of mission set names in the task packet data is in jaiaContext the menu should be updated when the context changes and the panel is re-rendered.

I'm going to remove the run search/apply filter button entirely. My thought process in including it was to decrease the load on the hub but in reality it probably has a negligible effect. I agree with you that it's current behavior/purpose is confusing.

joeymussalli:
> I do not think the numbers on the labels of the filters are needed and are a bit confusing. It implies things should be done in order but they work no matter what order you use them

I agree. Will remove the numbers.

joeymussalli:
> > > Clear filter should eliminate all filtering it seems the date filters are always applied and when Clear Filter is used the dates get reset but the reset dates do not include all of the available task packets. (see image above, the task packets from 6/3 get filtered out after pressing "Clear Filter"
> > 
> > 
> > Clearing the filter currently reverts the map view to what was the standard before this PR of the last 14 hours. If you think it's more intuitive I can leave the date range unchanged when clear filter is press.
> 
> I was not aware of the 14 hour rule, is that on the server side or client side? Ideally it would be obvious to the user either through a message or resetting the filter to match (a bit of a hack if it is being done on the server side).
> 
> As a user it seemed very odd that clicking "Clear Filter" when a name filter had pulled in old task packets can filter out task packets that are currently displayed.

It's server side. It probably is more intuitive if the date range wasn't affected by Clear Filter. I will change this.

JRoybot (approved):
Great job on this ! 
I think the  code is well structured and consistent with the design of JCC.  
UI very clean and simple to use.

ferrom-jaia:
@claude review

PR #1626, opened 2026-07-16
3D Map now fills entire window

brownc-jaia:
## Description:

Modified the JCC's 3D Depth Map to fill its entire window

## Tests:
(On the JCC)

1. Create a mission that generates a depth map
2. Double-Click on the depth map
3. Verify that the 3D Depth Map fills the entire window that is displayed rather than a small portion


ferrom-jaia (approved):
Looks good!

Tested in sim.

PR #1635, opened 2026-07-28
Feature/create util for downsampling/sw 2483

brownc-jaia:
## Summary
Created a new util header called `src/lib/utils/downsample.h` that provides the ability to downsample data to any byte size.

## Usage
There are two ways for the user to downsample their data:

1. Generic - User provides their data and the maximum bytes to downsample to and the util determines the data points to be used.
2. Specific - User provides their data, the maximum bytes, AND the (X,Y) values the function should use for downsampling. An example of usage is available within the header.


michael-jaia:
@claude review

PR #1647, opened 2026-08-03
bug/fix-jdv-blank-screen-on-undefined-constant-heading-speed

joeymussalli:
## Summary

Fixes a JDV white screen when hovering over a log file graph. The issue was caused by undefined constant heading speed values which made the constant heading distance NaN. There was no error boundary so the throw escaped `componentDidUpdate` and unmounted the whole React setup. Constant heading speed now defaults to 2 m/s.

## Test Procedure

- [x] Tested on the hub/log file setup that surfaced this issue (Fleet 6 Hub 1, 2026-07-28 log). Hovering between 16:51:47–16:57:15 UTC blanked the screen before the fix and works after.

### DIFF
--- a/src/web/shared/MissionFeatures.ts
+++ b/src/web/shared/MissionFeatures.ts
@@ -9,6 +9,9 @@ import { transformTranslate, point } from "@turf/turf";
 import { PortalBotStatus } from "./PortalStatus";
 import { getMapCoordinate } from "./Utilities";
 
+const CONSTANT_HEADING_DEFAULT_SPEED = 2; // meters per second
+const CONSTANT_HEADING_DEFAULT_TIME = 0; // seconds
+
 export function createMissionFeatures(
     map: Map,
     bot: PortalBotStatus,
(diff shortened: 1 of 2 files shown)


michael-jaia:
@claude review

michael-jaia:
@copilot Can you add this unit test please: 🟡 This PR fixes a real crash in createMissionFeatures (undefined constant_heading_speed/constant_heading_time producing NaN distance, which threw and blanked the JDV screen), but adds no unit test covering that scenario or the new default-speed/default-time/isFinite guard behavior. The function is pure and easily testable — a regression test would pin down the previously-crashing case (missing constant heading params) and the new default (speed=2, time=0) so it can't silently regress.

Extended reasoning...
createMissionFeatures in src/web/shared/MissionFeatures.ts is a pure function — it takes a map, bot status, mission plan, and a few flags, and returns an array of OpenLayers Feature objects. This PR changes its core NaN-guarding logic for constant-heading goals: previously the code assumed task.constant_heading.constant_heading_speed and constant_heading_time were always defined, multiplied them directly, and used the result as the transformTranslate distance. When either field was undefined, the multiplication produced NaN, which transformTranslate/turf likely threw on, and since this ran inside componentDidUpdate with no error boundary, the throw unmounted the whole React tree — the JDV white-screen bug described in the PR.\n\nThe fix defaults the speed to 2 and the time to 0 via ??, and adds an explicit isFinite(distance) \&\& isFinite(heading) guard before doing the transform, falling back to the goal's own location as the segment start point otherwise. This is a sound fix, but it is exactly the kind of pure, deterministic, input->output logic that is trivial to unit test, and none was added. There is no existing test file anywhere under src/web/shared (verified via search), so this change goes in completely uncovered.\n\nTest infrastructure for exactly this kind of feature-generation logic already exists in the repo: src/web/jest.config.js sets up ts-jest with a jsdom environment, and src/web/openlayers/layers/__tests__/bot-layer.test.ts / hub-layer.test.ts already build OpenLayers Features from status objects and assert on the resulting getFeatures() output — the same shape of test that would apply here. So 'where possible' is clearly satisfied; this isn't a case where OL/DOM mocking makes testing impractical.\n\nConcrete regression scenario a test should cover: construct a MissionPlan with one goal whose task.type === TaskType.CONSTANT_HEADING and task.constant_heading = {constant_heading: 90} (i.e. constant_heading_speed and constant_heading_time both undefined — the exact shape that crashed before this PR). Call createMissionFeatures(map, bot, plan, ...) and assert it returns without throwing, and that no constant-heading LineString segment is added (since distance defaults to 2 * 0 = 0, which is finite, so the current code would actually still enter the if branch and draw a zero-length segment — worth asserting explicitly since it's a subtle side effect of the 0-second default). A second test with a fully-defined constant_heading_speed/constant_heading_time should assert the previous (correct) end-coordinate behavior is unchanged.\n\nThis doesn't break anything at runtime — the crash fix itself is correct and was manually verified against a real log file per the PR's test procedure — so it's a nit rather than a blocking issue. But per this repo's REVIEW.md, 'Code comes with unit tests where possible' is listed under 'Always check', which is a repo-specific instruction that must be surfaced rather than treated as optional style, and per that same document all nits must be resolved before approval.

michael-jaia:
@claude review

PR #1676, opened 2026-08-14
3.y / Serve the fleet's bot and hub names from a DNS server on each hub

tsaubergine:
Every hub runs `dnsmasq` (as the generated `jaiabot_dns` service) resolving the names of all the bots and hubs in its fleet, so that neither operators nor tooling have to look an address up from the `jaia ip` scheme to reach a node:

```
jaia@hub1-fleet2:~$ ssh bot3
jaia@hub1-fleet2:~$ rsync bot3:/var/log/jaiabot/ .
jaia@hub1-fleet2:~$ ping hub0
```

## Names

Each node answers to four names, e.g. bot 3 of fleet 2 (`10.23.2.103`): `bot3`, `bot3.jaia`, `bot3-fleet2` (the machine's actual hostname) and `bot3-fleet2.jaia`. In addition `hub` resolves to whichever hub answered the query — the name the JCC's Apache virtual host already uses, so `http://hub/` reaches the JCC of the hub you are on — and `cloudhub` is an alias for `hub30`. Reverse lookups are served too (`10.23.2.103` → `bot3.jaia`).

`.jaia`, not `.local`, which RFC 6762 reserves for multicast DNS. `local=/jaia/` keeps fleet names off the internet and answers an unknown one `NXDOMAIN` rather than forwarding it. Set `jaia_network_dns_domain` in `/etc/jaiabot/network.env` for a different domain.

Records cover the whole of the addressing scheme — every bot and hub id it allows on the fleet WLAN, plus the CloudHub on the CloudHub VPN — rather than the nodes in `inventory.yml`. Adding a bot therefore needs no change on the hub, and a name cannot resolve to something other than what `jaia ip` reports. The tradeoff: a typo'd `bot99` resolves to an unreachable address instead of failing fast, where an inventory-driven server would say `NXDOMAIN`.

Anything that is not a fleet name is forwarded to the upstream nameservers the hub itself uses, so a hub's DNS can be the only one a client needs. The hub's own `/etc/resolv.conf` points at `127.0.0.1` with those upstreams kept as fallbacks, so it still resolves public names if `dnsmasq` is stopped.

## Where the package is installed

Both `debian/control` and the rootfs package list, as with the other embedded dependencies: first boot installs `jaiabot-embedded` with `dpkg -i`, which does not pull dependencies (hence the package list), while the dependency is what carries `dnsmasq-base` to fleets that upgrade rather than re-image. `dnsmasq-base` rather than `dnsmasq` because the latter enables a daemon of its own on install — port 53 on bots included — where the base package is the binary alone and the unit is generated like every other Jaia service.

## Scope

Only a physical hub of a runtime fleet serves DNS. Simulated fleets often share one host and so one port 53, and the CloudHub is reachable from the public internet, where an open resolver is an amplification liability. Bots and laptops keep resolving through their own nameservers by default — a hub that is powered off should not take name resolution with it — with the one-liner to opt in documented instead.

## `jaia admin bounds` (from review)

Rather than hardcoding the id ranges, the DNS generator asks for them. The new action reports what `ip.h` defines:

```
$ jaia admin bounds
bot id: [0 150]
hub id: [0 30]
fleet id: [0 255]
desktop id: [1 9]
gateway id: [0 0]
rpicam id: [0 49]
cloudhub id: 30

$ jaia admin bounds --bot_id --max
150
```

`--format json` gives the same as an object. As with `jaia ip`, the action execs a standalone binary (`jaia_bounds`) that links neither goby nor protobuf, so scripts that run at boot can call it directly.

The CloudHub is part of that: it is always hub 30, so `ip.h` now says so in its own right (`constexpr int cloudhub_id`, with a `static_assert` that it is a valid hub id) instead of `parse_host_code` inferring it from `hub_id_max`, and `bounds` reports it as a single id rather than a range. The three Python copies of `cloudhub_id = 30` (`config/gen/systemd.py`, `config/gen/hub.py`, `src/sh/fleet/jaia-create-cloudhub.py`) now read it from `jaia_bounds` as well, so the number exists in exactly one place.

## Also in this PR

`config/gen/systemd.py` had `service = 'jaia_firm_' + servi` (introduced in 556b63b, also present on `3.y`), a `NameError` that aborts `systemd.py` and so the whole postinst for any bot or hub — `gps-i2c-pty.py` reaches it — before any firmware unit or the new DNS unit can be enabled. Fixed here since the feature cannot be installed without it; happy to split it out.

## Testing

- Ran real `dnsmasq` 2.90 against the generated config: `bot3`, `bot3.jaia`, `hub1`, `hub`, `hub1-fleet2`, `bot150` and `cloudhub` (AAAA) all resolve, reverse gives `bot3.jaia.`, and an unknown `.jaia` name gives `NXDOMAIN` without going upstream.
- `systemd.py` emits `jaiabot_dns.service` for a runtime hub and not for a bot, a simulation hub, or hub 30 — the same split before and after the CloudHub id moved to `jaia_bounds`.
- The generator is idempotent across re-runs, and reconfiguring a hub to a bot restores the upstream `resolv.conf`.
- The generated hosts file is byte-identical across all three revisions of how the ranges are obtained (184 records for a fleet, in 0.5 s).

https://claude.ai/code/session_0179Kv1vJb4eN1nDYPKoKeon


src/sh/system/jaia-update-dns.sh
```diff
@@ -0,0 +1,236 @@
+#!/bin/bash
+
+# Writes the fleet DNS configuration served by the jaiabot_dns service, and
+# points this machine's resolver at it.
+
+set -u -e
+
+# same as the node id ranges in jaiabot/src/lib/utils/ip.h
+bot_id_max=150
```
tsaubergine: I'd like to not hardcode these numbers. Can we add a `jaia admin bounds` action that gives the min/max bot and hub ids from ip.h?

`jaia admin bounds` with no arguments gives something like:
```
bot id: [0 150]
hub id: [0 30]
fleet id: [0 250]
```

`jaia admin bounds --format json` gives something like:
```
{ "bot_id": { "min": 0, max: "150" }, ...}
```
`jaia admin bounds --bot_id --max` gives

```
150
```

tsaubergine: Done in 4dd88b5 — the script now asks rather than hardcodes:

```bash
bot_id_min=$(jaia_bounds --bot_id --min)
bot_id_max=$(jaia_bounds --bot_id --max)
hub_id_min=$(jaia_bounds --hub_id --min)
cloudhub_id=$(jaia_bounds --hub_id --max)
```

All three output forms are as you specified. `jaia admin bounds` with no arguments:

```
bot id: [0 150]
hub id: [0 30]
fleet id: [0 255]
desktop id: [1 9]
gateway id: [0 0]
rpicam id: [0 49]
```

`--format json` gives `{ "bot_id": { "min": 0, "max": 150 }, ... }`, and a single id with a single end (`--bot_id --max`) gives a bare `150`. Selecting narrows in both directions: `--min`/`--max` alone gives that end for every id, and `--hub_id --format json --min` gives just `{ "hub_id": { "min": 0 } }`.

Two things worth flagging:

- It reports every node type in `ip.h`, not just bot and hub, since they were all right there. Say the word if you'd rather it only showed the three in your example.
- Your example has `fleet id: [0 250]` but `ip.h` has `fleet_id_max = 255`, so that is what it prints. The `jaiabot-embedded` debconf template's `fleet_id` choices do stop at 250, so one of the two is wrong — I have not touched either, since which one gives is your call.

Implemented as a standalone `jaia_bounds` binary that `jaia admin bounds` execs, the same way `jaia ip` wraps `jaia_ip`: `jaia-update-dns.sh` runs at boot and in the postinst, so it calls `jaia_bounds` directly and skips loading goby/protobuf. Docs are in `page005_jaia_tool.md`.

---

src/sh/system/jaia-update-dns.sh
```diff
@@ -0,0 +1,236 @@
...
+        write_resolv_conf "$(upstream_nameservers)"
+    fi
+}
+
+bot_id_min=$(jaia_bounds --bot_id --min)
+bot_id_max=$(jaia_bounds --bot_id --max)
+hub_id_min=$(jaia_bounds --hub_id --min)
+# the highest hub id is the CloudHub, which is reachable over the CloudHub VPN
+# rather than the fleet WLAN
+cloudhub_id=$(jaia_bounds --hub_id --max)
```
tsaubergine: Cloudhub is hub 30 (always). Update this to reflect that and add it to jaia admin bounds as a special case.

tsaubergine: Done in 0a602fd. The CloudHub id now exists in its own right in `ip.h` rather than being inferred from the top of the hub range:

```cpp
// The CloudHub is always this hub id, on every fleet
constexpr int cloudhub_id = 30;
static_assert(cloudhub_id >= hub_id_min && cloudhub_id <= hub_id_max,
              "cloudhub_id must be a valid hub id");
```

`parse_host_code` uses it for the `ch` code too, which previously read `hub_id_max` — same value today, but it would have followed the hub range if that ever grew.

`jaia admin bounds` reports it as a special case, one id rather than a range:

```
$ jaia admin bounds
...
rpicam id: [0 49]
cloudhub id: 30

$ jaia admin bounds --cloudhub_id
30

$ jaia admin bounds --cloudhub_id --format json
{
    "cloudhub_id": 30
}
```

Since it is already a single number, `--cloudhub_id` alone gives the bare `30` without needing `--min`/`--max` (either of those gives 30 as well).

The DNS generator now says what it means — it walks the whole hub range and skips the CloudHub, instead of stopping one short of the top:

```bash
cloudhub_id=$(jaia_bounds --cloudhub_id)
...
for id in $(seq ${hub_id_min} ${hub_id_max}); do
    if [ "${id}" = "${cloudhub_id}" ]; then
        continue
```

Generated hosts file is byte-identical to before the change (184 records, `hub30`/`cloudhub` still the AAAA on the CloudHub VPN).

One thing I did not touch: `cloudhub_id = 30` is also hardcoded in `config/gen/systemd.py`, `config/gen/hub.py` and `src/sh/fleet/jaia-create-cloudhub.py`. They can all read it from `jaia_bounds --cloudhub_id` now (`common/comms.py` already shells out to `jaia_ip`, so there is precedent), but that is outside this PR's diff and touches config generation for every bot and hub, so I would rather do it deliberately than fold it in here. Happy to add it to this PR or a follow-up, whichever you prefer.

---

tsaubergine (commented):
Cloudhub is hub 30 (always). Update this to reflect that and add it to jaia admin bounds as a special case.

src/sh/system/jaia-update-dns.sh
tsaubergine: Followed up on the Python in e993f42 — all three copies of `cloudhub_id = 30` now come from `jaia_bounds`, so `ip.h` is the only place the number exists.

They ask in three different ways, because of what each one can import:

- **`config/gen/hub.py`** already imports the `common` package, so the subprocess call lives there in a new `common/bounds.py`, next to the existing `jaia_ip` wrapper in `common/comms.py`:
  ```python
  cloudhub_id=common.bounds.cloudhub_id()
  ```
- **`config/gen/systemd.py`** cannot import `common` — its `__init__.py` calls `check_args()` for an app name in `argv[1]` and requires `jaia_log_dir`/`jaia_mode` in the environment, none of which a maintainer script has. It gets a small local helper that runs the binary under `--jaiabot_bin_dir`, so the from-source deploy path uses the build's `jaia_bounds` rather than whatever is in `/usr/bin`.
- **`src/sh/fleet/jaia-create-cloudhub.py`** is installed to `/usr/bin`, away from the `common` package, so it calls `jaia_bounds` from `PATH` inline.

Verified: `common.bounds.cloudhub_id()` returns `30` as an `int` through the real package import, and `systemd.py` still generates 20 units including `jaiabot_dns` for hub 1, and 18 without it for hub 30 — the same split it had with the literal.

`config/gen/common/bounds.py` ships automatically (the whole `config` tree is installed by the top-level `install(DIRECTORY ...)`), so there is no packaging change.

Two related numbers I left alone, since they are not the CloudHub id and not from `ip.h`: `common/comms.py`'s `number_of_hubs_max`/`number_of_bots_max`, which mirror `src/lib/comms/comms.h` (its own `hub_id_max`, for modem ids), and `hub.py`'s `iridium_jaia_tech_hub_id`. Say the word if you want either folded into `bounds` as well.

---

PR #1679, opened 2026-08-14
3.y / Generate web protobuf TypeScript types from the .proto files / SW-2439

tsaubergine:
Jira: [SW-2439](https://jaia-innovation.atlassian.net/browse/SW-2439)

# Toby
## Goal

I want to make the .protos the sole source of truth for data types passed throughout the Jaia system. This is already true for the C++, C (nanopb/MCU), and Python (inc. REST API) code. This will reduce the possibility of data-mismatch errors and reduce the amount of work required to change a .proto message contents.

However, the Typescript code uses hand-edited versions of the .protos. This PR is intended to remove this manual pathway in favor of autogenerated Typescript code directly from the .protos using ts-proto.

## Testing Performed

Rebuild JCC, ran basic simulation with 4 bots and 1 hub. All appears to work. We can do more testing with the full 3.y release.

# Claude

## Summary

`src/web/types/protobuf-types.ts` was hand-edited whenever a `.proto` changed. There was also a near-duplicate at `src/web/shared/JAIAProtobuf.ts` used by `shared/` and `jdv/`; the two had drifted from the protos and from each other. Both are replaced by [ts-proto](https://www.npmjs.com/package/ts-proto) output, generated at build time and not checked in.

## Where these types sit in the data path

The generated types are **types only** — they never encode anything. The browser talks to the Flask server (`src/web/server/app.py`) in plain HTTP JSON, and `jaia_portal.py` is what crosses into protobuf:

```
JCC/JED (TS)  --HTTP JSON-->  Flask (jaia_portal.py)  --binary protobuf over UDP-->  jaiabot_web_portal
                                 ParseDict / MessageToDict          ClientToPortalMessage
                                                                    PortalToClientMessage
```

So the types describe the JSON representation at the browser↔Flask boundary, which is `google.protobuf.json_format` with `preserving_proto_field_name=True`. That is what the ts-proto options are chosen to mirror:

| option | why |
| --- | --- |
| `snakeToCamel=false` | `preserving_proto_field_name=True` |
| `stringEnums=true` | `MessageToDict` emits enum names |
| `unrecognizedEnum=false` | `MessageToDict` never emits `UNRECOGNIZED` |
| `forceLong=string` | proto3 JSON maps 64-bit ints to strings |
| `useOptionals=all` | unset fields are simply absent from the dict |
| `onlyTypes=true` | interfaces and enums only, no encode/decode runtime |

## Generation

`src/web/gen_protobuf_types.sh` runs `protoc` with the ts-proto plugin over `jaiabot/messages/portal.proto` and `rest_api.proto`, which transitively cover every message the web apps consume — 21 files mirroring the proto package layout. **The output is not checked in.** It is generated in three places:

- **CMake** generates into `build/<arch>/intermediate/web/shared/proto`; `jcc_jed` and `jdv` depend on it, so the built bundles always match the `.proto` files they were built alongside. Include paths come from `GOBY_INCLUDE_DIR` / `DCCL_INCLUDE_DIR` / `Protobuf_INCLUDE_DIRS`, so cross builds resolve through the sysroot.
- **`src/web/run.sh`** generates after installing npm dependencies, for the dev watch loop.
- **`npm run gen:proto`** for working in `src/web` directly — needed once on a fresh clone before running `tsc` or `jest` by hand. Documented in `src/web/README.md`.

No C++ build is required for any of this; the web targets do not depend on the C++ graph.

The generated tree currently sits under `shared/` because `shared/` is symlinked into `jdv/client/src/`, and TypeScript compiles the symlinked copy as its own set of files — so a relative import out of `shared/` resolves inside `jdv/client/src/` rather than `src/web/`. Putting them anywhere else needs a path alias. That placement is under discussion (see review thread) since it deepens JCC's coupling to `shared/`.

## Migration

Consumers import the generated modules directly, so nested types now carry their parent's name:

| was | becomes |
| --- | --- |
| `TaskType` | `MissionTask_TaskType` |
| `Goal`, `MissionStart`, `MovementType` | `MissionPlan_*` |
| `DiveParameters`, `DriftParameters`, `ConstantHeadingParameters`, `StationKeepParameters` | `MissionTask_*` |
| `CommandType` | `Command_CommandType` |
| `HubCommandType` | `CommandForHub_HubCommandType` |
| `BotType`, `ActiveLink` | `BotStatus_*` |
| `Metadata`, `Version` | `DeviceMetadata`, `DeviceMetadata_Version` |
| `PIDSettings` | `PIDControl_PIDSettings` |
| `BotOffloadData` | `HubStatus_BotOffloadData` |
| `ContactStatus` | `ContactUpdate` |

## Drift the generated types exposed

Real bugs this fixes:

- `Error` was declared with goby coroner's 3 values, but `BotStatus.error` carries `jaiabot.protobuf.Error`, which has 115. `HealthRow` was rendering that field against the wrong enum.
- `Metadata` was missing 11 fields (`raspi_firmware_version`, `fleet_id`, `hub_id`, `bot_id`, …) and `Version` was missing `deb_repository` / `deb_release_branch`.
- `Attitude`, `Speed` and `EulerAngles` were each declared twice in one file, so TypeScript silently merged them — `Attitude` was a union of goby's gpsd `Attitude` and jaiabot's.
- `uint64` fields were typed `number` but arrive as strings; timestamps now get an explicit `Number()` where the app wants a number.

Consistency fixes, not behavior changes:

- `MissionPlan.bottomDepthSafetyParams` used the lowerCamelCase JSON name while the rest of the file used proto field names. `ParseDict` accepts both spellings, so this was **not** a functional bug — but it disagreed with what `MessageToDict` emits on the read path. The generated types use `bottom_depth_safety_params` throughout.
- `BottomDepthSafetyParams` was populated with stringified numbers; `ParseDict` accepts numeric strings for numeric fields, so again not a functional bug, but the fields are `double`/`int32` and are now passed as numbers.

One test assertion changed substantively: `HealthRow.test.tsx` used `Error.ERROR__PROCESS_DIED`, which does not exist in the jaiabot enum, replaced with `ERROR__FAILED__JAIABOT_HEALTH`.

## Verification

- `tsc --noEmit` clean for both `src/web` and `jdv/client` (the 21 remaining errors are pre-existing ones inside `node_modules`, confirmed against a baseline worktree at the merge base).
- `npm test`: 166 passed, 11 skipped, 0 failed, including a new `mission-package.test.ts` covering the `bottom_depth_safety_params` wire key.
- Fresh-clone path: with no generated types in the tree, a real `cmake` configure + `cmake --build . --target jcc_jed jdv` generates them and builds both bundles.

https://claude.ai/code/session_01EoV7RFvFVG5hrXkJ4AgFGQ

[SW-2439]: https://jaia-innovation.atlassian.net/browse/SW-2439?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/web/components/JaiaAbout/JaiaAbout.tsx
```diff
@@ -99,7 +99,7 @@ export default function JaiaAbout() {
                 <div className="input">{COMPANY_ADDRESS}</div>
             </div>
             <div className="jaia-about-row">
-                <div className="label">Software Version:</div>
+                <div className="label">Software DeviceMetadata_Version:</div>
```
tsaubergine: This should still read "Version" as we don't want to change the user-facing text in this pull request. Check for other places in this new PR that this mistake might have been made.

tsaubergine (commented):
Is it possible to regenerate the protos as part of the CMake build and/or as part the src/web/run.sh script?

tsaubergine:
# Claude 

Pushed b4f1c05 addressing the review, plus a correction to my own PR description.

**Correction first:** I claimed `MissionPlan.bottomDepthSafetyParams` meant bottom depth safety params "were never reaching the hub." That is wrong. `json_format.ParseDict` accepts both the proto field name and its lowerCamelCase JSON name, so the old key worked. Verified against the generated Python bindings:

```
ParseDict({'bottomDepthSafetyParams': {...}}, MissionPlan())  -> accepted
ParseDict({'bottom_depth_safety_params': {...}}, MissionPlan()) -> accepted
ParseDict({'nonsenseKey': 1}, MissionPlan())                   -> ParseError
```

Same for `BottomDepthSafetyParams` being populated with stringified numbers — `ParseDict` accepts numeric strings for numeric fields. Both are consistency fixes (the read path via `MessageToDict(preserving_proto_field_name=True)` emits `bottom_depth_safety_params`), not bug fixes. The PR description now says so. The genuine bugs — the wrong `Error` enum, the missing `DeviceMetadata` fields, and the silently merged duplicate `Attitude`/`Speed`/`EulerAngles` interfaces — stand as described.

**"Software Version:" label** — fixed. The blanket identifier rename reached a JSX text node, which my earlier scan for mangled strings missed because it only looked inside quotes and comment lines. I re-scanned the whole diff for JSX text nodes and quoted strings; that label was the only user-facing hit. One comment also read worse (`Goal / Waypoint map style function` → `MissionPlan_Goal / ...`) and is reverted. The `@param {X}` / `@returns {X}` annotations stay renamed, since those are type references rather than prose.

**Regenerating from CMake and run.sh** — both added.

- `src/web/CMakeLists.txt` regenerates into `build/<arch>/intermediate/web/shared/proto` at build time, with `jcc_jed` and `jdv` depending on it, so the built bundles always match the `.proto` files. Include paths come from `GOBY_INCLUDE_DIR` / `DCCL_INCLUDE_DIR` / `Protobuf_INCLUDE_DIRS` rather than a hardcoded `/usr/include`, so cross builds resolve through the sysroot. The configure step drops the copied-in `shared/proto` and its stamp, so the intermediate tree only ever holds freshly generated types.
- `src/web/run.sh` regenerates right after `install_dependencies.sh`.

`gen_protobuf_types.sh` now takes the output directory as an argument for this. Verified with a real `cmake` configure and `cmake --build --target jcc_jed jdv`: both bundles build, and the CMake-generated types are byte-identical to the committed copy.

That leaves the committed copy serving IDEs, `tsc` and jest, with three independent regeneration points keeping it honest. Happy to drop the committed copy entirely if you would rather the generated code not live in the tree, but that would cost IDE type-checking and a bare `npm test` outside CMake.

**Copilot's two points** were both worth taking:

- The CI check used `git diff --exit-code`, which ignores untracked files — a newly generated file would have passed. Now uses `git status --porcelain`. Confirmed the old form missed a new untracked file in that directory and the new one catches it.
- Added `data/mission_set/__tests__/mission-package.test.ts` asserting the packaged plan carries `bottom_depth_safety_params` with numeric values and no camel-case key, and omits it when unset.

Unrelated to this branch: the `copilot-pull-request-reviewer` check failed in its own "Install build tools" setup step (`scripts/build/setup-tools-build.sh`), which this PR does not touch.

---

src/web/shared/proto/dccl/option_extensions.ts
JRoybot, line 1: Since this is and the message descriptions are generated files do they need to be checked in?

src/web/gen_protobuf_types.sh
JRoybot, line 1: Does this belong here?  Would it be better placed somewhere in scripts

src/web/shared/proto/goby/middleware/protobuf/coroner.ts
JRoybot, line 1: Will these generated files get checked into the repo or will there only be local copies from the build process?

JRoybot (commented):
This is a great improvement!  

Code all looks good.  Left a few basic questions.  

During the JCC refactor Twomey removed all uses of src/web/shared from JCC and only left it there so we did not touch JDV.  Should check with him to see if he has any strong opinions on the location of the message definitions.

I was unable to build the branch.  I ran setup-tools-build.sh but got FAILED: lib/libjaiabot_messages.so.1 lib/libjaiabot_messages.so 

What is required to build this ?  Can it be built on 24.04?

tsaubergine:
> This is a great improvement!
> 
> Code all looks good. Left a few basic questions.
> 
> During the JCC refactor Twomey removed all uses of src/web/shared from JCC and only left it there so we did not touch JDV. Should check with him to see if he has any strong opinions on the location of the message definitions.
> 
> I was unable to build the branch. I ran setup-tools-build.sh but got FAILED: lib/libjaiabot_messages.so.1 lib/libjaiabot_messages.so
> 
> What is required to build this ? Can it be built on 24.04?

Yes it builds on 24.04 even though the target is 26.04 primarily.

`setup-tools-build.sh` should work directly but the new workflow is:

On new machine:
```
cd jaiabot
source init.sh
```

After that use `jaia dev build` (see `jaia dev --help`)

tsaubergine:
# Claude 

Thanks — answers to all four, and 8b41edf acts on the generated-files one.

**Do the generated files need to be checked in?** No, and they no longer are. Now that both CMake and `run.sh` regenerate them, the committed copy was only serving IDEs and a bare `tsc`/`jest` run in `src/web` — not worth 3.1k lines of generated code in every diff. They are gitignored, and `src/web/README.md` documents `npm run gen:proto` for working in `src/web` outside CMake. The CircleCI freshness check went with them (nothing left to compare against), as did the prettier pass over the output, which only existed to keep the committed copy stable under the pre-commit hook.

Verified the fresh-clone path: with no `src/web/shared/proto` in the tree at all, `cmake` configure + `cmake --build . --target jcc_jed jdv` generates the types into `build/<arch>/intermediate/web/shared/proto` and both bundles build.

**Should the script live in `scripts/`?** It reads on the proto files but is really part of the web build — it needs `src/web/node_modules` for the ts-proto plugin, and it is wired up as `npm run gen:proto`, which resolves relative to `src/web`. `install_dependencies.sh` and `run.sh` are already there for the same reason, so `src/web/` seemed the consistent home. Happy to move it if you would rather all shell scripts live under `scripts/`; it would just need the npm script to point up a couple of levels.

**Location of the message definitions / JCC vs `src/web/shared`.** Good catch, and I would rather not decide this without Twomey. Where it stands:

- JCC does still reach into `shared/` on 3.y today — `shared/PortalStatus` (22 sites), `shared/Utilities` (5), `shared/IPPrefix` (1) — so the coupling was reduced rather than removed. This PR adds 112 more import sites into `shared/proto`, which is a real deepening of exactly what the refactor was walking back.
- The reason they are under `shared/` is the `jdv/client/src/shared -> ../../../shared` symlink. TypeScript compiles the symlinked copy as its own set of files, so a relative import out of `shared/` resolves inside `jdv/client/src/`, not `src/web/`. That means anything `shared/` imports has to live under `shared/` too, or the jdv build breaks. I hit this concretely: the first cut put them in `src/web/types/proto` and jdv failed with `Cannot find module '../types/proto/...'`.
- The clean alternative is a path alias: generated types at `src/web/proto`, imported everywhere as `@proto/jaiabot/messages/mission`. JCC's coupling to `shared/` stays at today's 28 sites and the symlink problem disappears. The cost is declaring the alias in four places — `src/web/tsconfig.json`, `jdv/client/tsconfig.json`, both webpack configs, and jest's `moduleNameMapper`.

I am happy to do the alias version if that is the direction — it is a mechanical change on top of this. @tsaubergine, worth pulling Twomey in before I move it?

**Building.** tsaubergine already covered the general answer (`source init.sh`, then `jaia dev build`; 24.04 works). Worth adding for this PR specifically: you do not need the C++ build at all to check it. The web targets do not depend on the C++ graph — that is what #1682 relied on — so `cmake -DCMAKE_BUILD_TYPE=Debug .. && cmake --build . --target jcc_jed jdv` is enough, and `cd src/web && npm install && npm run gen:proto` is enough to just look at the generated types. For what it is worth, `libjaiabot_messages.so` does build from this branch on 24.04 here, so that failure looks environmental rather than something this branch introduced — if you still hit it, the full error would help.

---

tsaubergine:
# Claude 

You did not miss a step — neither failure comes from this PR. The branch changes nothing outside `src/web/`, and no Python at all:

```
$ git diff --name-only origin/3.y...HEAD | grep -E '\.py$|src/python|src/web/server'
(nothing)
```

Both failures are in the Python venv half of `run.sh`, which runs before the one line this PR adds to that script.

**The `geojson` one is a real pre-existing bug on 3.y.** `pyjaia/src/pyjaia/drift_interpolation.py` imports `geojson`, but it is declared neither in `src/python/requirements.txt` nor in `pyjaia/pyproject.toml`'s `dependencies`. It only works today because `build_venv.sh` creates the venv with `--system-site-packages`, so it picks up a system-wide `geojson` where one happens to exist. A fresh 26.04 WSL has none, so the import fails. On 3.y:

```
$ git show origin/3.y:src/python/requirements.txt | grep -c geojson
0
$ git show origin/3.y:src/python/pyjaia/src/pyjaia/drift_interpolation.py | grep -c geojson
2
```

To unblock right now:

```
build/web_dev/python/venv/bin/pip install geojson
```

The proper fix is adding `geojson` to `pyjaia`'s `dependencies` (it is that package's import, not the venv's). Happy to put that up as its own small PR if you want — it is unrelated to this one, and `python3.14-venv` looks like a separate base-image gap.

**Worth knowing for reviewing this PR specifically:** you do not need `run.sh` at all. It brings up the Flask server and the Python venv, none of which this PR touches. `jaia dev build` on its own builds the web bundles including the generated types, and for just the TypeScript side:

```
cd src/web && npm install && npm run gen:proto
npx tsc --noEmit && npm test
```

Also, from your paste — the run got to `Building the client apps` and `31 modules`, which means webpack was already compiling against freshly generated types. The generation step worked on your machine; it was only the Python server that fell over afterwards.

---

tsaubergine:
# Claude 

Great — and thanks for running it all the way through to JCC in the browser on a fresh 26.04 WSL. That is a useful second confirmation on top of tsaubergine's 4-bot sim.

On pushing the `pyproject.toml` / `requirements.txt` edits: my suggestion is a separate PR against 3.y rather than onto this branch, though it is @tsaubergine's call as PR author.

Reasons to split it:

- It is genuinely unrelated. This branch touches nothing outside `src/web/`, so folding a Python dependency fix in muddies an otherwise self-contained diff and gives reviewers two unrelated things to reason about.
- It unblocks people independently of this PR. Anyone doing a fresh `run.sh` setup hits the `geojson` failure today on 3.y, whether or not this lands — so the fix is more useful merged on its own schedule, and does not have to wait on the `shared/proto` location question that is still parked for Twomey.
- It keeps the fix findable. A dependency fix buried in a commit titled "generate web protobuf TypeScript types" is hard to locate later.

One note on where to put it: `geojson` is imported by `pyjaia/src/pyjaia/drift_interpolation.py`, so `pyjaia`'s `[project] dependencies` in `pyproject.toml` is the right home for it — `requirements.txt` covers the venv's other installs. If it works with it in only one of the two, `pyproject.toml` is the one that should carry it. Worth double-checking whether the other packages you added are pyjaia imports too, or genuinely venv-level.

Either way, happy to open that PR myself if you would rather not carry the change — just say which packages you ended up adding.

---

JRoybot (approved):
Verified build on 26.04 with minor modifications to fix pre-existing Python venv issues.  Location of generated files should be decided by @michael-jaia

tsaubergine:
# Claude 

@JRoybot — following up on my "put it in a separate PR" suggestion: tsaubergine asked for it here instead, so 0708f41 fixes it on this branch. Rather than adding the missing packages to `requirements.txt` / `pyproject.toml`, it installs them from the authoritative list.

**The gap.** `build_venv.sh` creates the venv with `--system-site-packages`, so several `pyjaia` imports resolve to system packages rather than to anything in `requirements.txt`. Those system packages are already declared — `python3-geojson` is right there in `jaiabot-python`'s `Depends` in `debian/control` — but nothing installed them for a source-tree developer. `setup-tools-build.sh` covers `Build-Depends` via `apt-get build-dep`, which is a different set. Hence a machine that builds fine but dies in `run.sh`.

New `scripts/build/install-runtime-deps.sh` reads the `Depends` of the named binary packages out of `debian/control` and installs whichever are missing:

```
$ scripts/build/install-runtime-deps.sh jaiabot-python jaiabot-web
🟢 Installing 33 missing runtime dependencies of jaiabot-python jaiabot-web: apache2 gdal-bin ... python3-geojson ...
```

`run.sh` calls it before building the venv. Already-installed packages are skipped, so the usual case needs neither sudo nor network, and a failure to install warns rather than aborting.

**The same gap elsewhere**, since tsaubergine asked me to look:

- `scripts/build/setup-tools-runtime.sh` had a hand-maintained copy of the same list, and it had drifted. It was missing 30 packages — `python3-geojson` among them, so it would not have saved you either — plus `python3-shapely`, `python3-simplejson`, `python3-flask-compress`, `python3-protobuf`, `python3-scipy`, `python3-numpy`, `python3-pandas`. It also carried `python3-dateutil` and `python3-pyqt5`, which no jaiabot package depends on. It now reads `debian/control` for `jaiabot-python`, `jaiabot-web` and `jaiabot-embedded`, keeping an explicit list only for genuinely dev-only tools (opencpn, i2c-tools, screen, socat, parallel, the goby `-dev` packages, and those two Python packages). I verified every package from the old list is still covered.
- `scripts/sim-docker/Dockerfile` has no gap — it installs the real `.deb`s, so apt resolves `Depends` itself. That is why the sim container works where a source-tree run did not.
- `.docker/*/Dockerfile` are build images and correctly carry build deps only.
- `scripts/hardware/imu-bno085/create-python-virt-env.sh` is a bare `python3 -m venv venv` with no requirements install at all — separate and much smaller, flagging it rather than touching it here.

Worth saying: declaring `geojson` in `pyjaia`'s `pyproject.toml` is still defensible on its own merits, and would make `pyjaia` installable outside this tree. It is just orthogonal to the system-package path, so I did not fold it in. If you would still like to push your `pyproject.toml` / `requirements.txt` edits, they should not conflict.

---

michael-jaia:
@JRoybot @tsaubergine If it is a light lift, can we move the output to `src/web/types/`? I imagine Claude can knock that one out. The shared directory had good intentions early in JCC/JDV development, but it spiraled out of control and after the refactor, it only needs to be used for JDV until that is reworked.

tsaubergine:
# Claude

@michael-jaia Yes, it's a light lift — I prototyped the move and verified it end to end. One wrinkle worth a decision before I push.

**The wrinkle.** `src/web/types/` was my original plan, and it failed on the `jdv/client/src/shared -> ../../../shared` symlink. Nine files under `shared/` import the generated types, and TypeScript compiles the symlinked copy as its own set of files, so a relative import out of `shared/` resolves inside `jdv/client/src/` rather than `src/web/`. Moving the types to `types/proto` and making those nine use `../types/proto/...` gives exactly this, 173 errors' worth:

```
jdv/client/src/shared/PortalStatus.ts: Cannot find module '../types/proto/jaiabot/messages/hub'
```

So the move needs a way for jdv to reach them. I tried two:

**1. A second symlink** (`jdv/client/src/types/proto -> ../../../../types/proto`), mirroring the existing `shared` one. Works — tsc clean, 238 tests, both bundles build. But it's the same trick you're trying to get away from, so I didn't stop there.

**2. A `@proto/*` path alias**, which is what I'd suggest. Types live at `src/web/types/proto`, everything imports `@proto/jaiabot/messages/mission`, and no symlink is involved. Verified the same way: `tsc --noEmit` clean for both `src/web` and `jdv/client`, 238 tests pass, and a full `cmake --build --target jcc_jed jdv` produces both bundles. Worth confirming that last one specifically — `ts-loader` does not read tsconfig `paths`, so the webpack aliases are doing real work, not just satisfying the type checker.

The diff is 103 files, +148/−139 — almost entirely one-line import rewrites. The non-mechanical part is five config files:

| file | addition |
| --- | --- |
| `src/web/tsconfig.json` | `baseUrl` + `paths: {"@proto/*": ["types/proto/*"]}` |
| `src/web/jdv/client/tsconfig.json` | same, pointing at `../../types/proto/*` |
| `src/web/webpack.config.js` | `resolve.alias` `@proto` |
| `src/web/jdv/client/webpack.config.js` | same |
| `src/web/jest.config.js` | `moduleNameMapper` `^@proto/(.*)$` |

A side benefit: import paths stop being depth-dependent, so `@proto/...` reads the same from `components/`, `data/` and `shared/` alike, and files can move without rewriting them.

Two things I'd like a steer on before pushing:

- **Alias or symlink?** I lean alias for the reason above, but it is five config files instead of one symlink, and that is your call rather than mine.
- **`types/proto` or `types/`?** The generator writes a tree mirroring the proto packages (`jaiabot/messages/`, `goby/middleware/protobuf/`, `dccl/`, `google/protobuf/`), so it wants its own subdirectory rather than being spread through `types/` next to the hand-written `.ts` files. `src/web/types/proto` keeps it gitignored as one unit.

Say the word and I'll push it — the change is ready, not just sketched.

---

tsaubergine:
# Claude

@michael-jaia Done — pushed as d10a182. Generated types now live at `src/web/types/proto`, reached everywhere as `@proto/...`, and `src/web/shared/proto` is gone.

Went with the alias rather than a second symlink, for the reason you gave: `shared/` should be shrinking, not growing another dependency. The alias is declared in five places, which is the whole cost of the move:

| file | addition |
| --- | --- |
| `src/web/tsconfig.json` | `baseUrl` + `paths: {"@proto/*": ["types/proto/*"]}` |
| `src/web/jdv/client/tsconfig.json` | same, `../../types/proto/*` |
| `src/web/webpack.config.js` | `resolve.alias` `@proto` |
| `src/web/jdv/client/webpack.config.js` | same |
| `src/web/jest.config.js` | `moduleNameMapper` `^@proto/(.*)$` |

Everything else is one-line import rewrites: 103 files, +149/−140.

Both webpack aliases are load-bearing — `ts-loader` does not read tsconfig `paths`, so a green `tsc` alone would not have told us the bundles still build. Verified accordingly: `tsc --noEmit` clean for `src/web` and `jdv/client`, `npm test` 238 passed / 11 skipped, and a full `cmake --build . --target jcc_jed jdv` producing both bundles.

One consequence worth flagging: when jdv is reworked and stops sharing those nine files, nothing about this needs revisiting. The alias is independent of the symlink, so `shared/` can be dismantled whenever you like without touching a single proto import.

---

PR #1680, opened 2026-08-17
3.y / Replace NTP with Chrony for time synchronization / SW-2203

tsaubergine:
Jira: [SW-2203](https://jaia-innovation.atlassian.net/browse/SW-2203)

## Summary

This PR replaces the NTP daemon (ntpsec) with Chrony for time synchronization across the system. The changes include:

- **Time synchronization daemon**: Replaced ntpsec with chrony as the primary NTP implementation
- **GPS time integration**: Updated GPS drivers (SPI and I2C) to send data to gpsd over UDP instead of PTY, enabling gpsd to export time to chrony via shared memory
- **NTP status monitoring**: Refactored `NTPStatusThread` to query chrony via `chronyc` instead of `ntpq`, parsing CSV output from `chronyc tracking` and `chronyc sources` commands
- **Configuration**: Added chrony GPS refclock configuration and updated gpsd to listen on UDP port 32100 and run without waiting for clients
- **Systemd integration**: Updated service dependencies to use chrony instead of ntp

The refactoring maintains the same protobuf message structure for NTP status reporting, mapping chrony's data (leap status, stratum, peer information) to the existing `NTPStatus` message format.

## Test Procedure

- Verify chrony daemon starts and synchronizes with GPS time source
- Confirm `NTPStatusThread` successfully parses chrony output and publishes status
- Validate that GPS time is available to chrony via gpsd shared memory interface
- Check that system time synchronization works in both simulation and runtime modes

https://claude.ai/code/session_01PLCXzvLbU6b6dSAzGbSAL7

[SW-2203]: https://jaia-innovation.atlassian.net/browse/SW-2203?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/messages/health.proto
```diff
@@ -437,7 +437,7 @@ message NTPStatus
     option (dccl.msg) = {
         unit_system: "si"
     };
-    // from ntpq -c rv 0
+    // from chronyc -c tracking and the mode of the selected source
```
tsaubergine: Update the proto messages to better map onto the output of chronyc. No need to keep anything NTP specific here.

tsaubergine: Done in c4ab30a — `NTPStatus` is now `ChronyStatus`, shaped around what `chronyc -c` actually prints.

**Tracking** (`chronyc -c tracking`) contributes `reference_id`, `reference_name`, `stratum`, `reference_time`, `system_time_offset`, `last_offset`, `rms_offset`, `frequency`, `residual_frequency`, `skew`, `root_delay`, `root_dispersion`, `update_interval` and `leap_status`.

**Sources** (`chronyc -c sources` + `-c sourcestats`) become a `Source` submessage. The single NTP tally code is split into the two things chrony reports separately — `mode` (`#` refclock, `=` peer, `^` server) and `state` (`*` selected, `+` combined, `-` not combined, `?` unreachable, `x` falseticker, `~` too variable) — plus `sample_points`, `runs`, `span` and `std_dev` from sourcestats.

Gone entirely: `SyncSource` (the mode 6 clock source codes), `system_event_counter`, `NTPSystemEvent`, and the `NTPPeer`/`TallyCode` pair. Health warnings now key off `selected_source`, with `std_dev` replacing jitter.

The error/warning enums were renamed too (`ERROR__SYSTEM__CHRONY_TRACKING_QUERY_FAILED`, `WARNING__SYSTEM__CHRONY_STD_DEV_HIGH`, etc.). That changes the DCCL schema hash for `BotStatus` and for `Hub2HubData` which embeds it, so both expected hashes are updated. The encoding is unchanged — same enum numbers, field ids and ordering — so `PROJECT_INTERVEHICLE_API_VERSION` stays at 23; shout if you'd rather bump it.

Also updated the two hand-maintained TS mirrors (`protobuf-types.ts`, `JAIAProtobuf.ts`) and the HDF5 schema doc. Local build, `ctest` (3/3) and `npm test` (164) all pass.

---

tsaubergine:
### CI status

`amd64-resolute-build` has failed three times on this branch (32520, 32526, 32527) with the **same** failure, and it is not this PR's code:

```
Killed
gmake[2]: *** [src/web/CMakeFiles/jcc_jed.dir/build.make:164: share/jaiabot/web/jcc/client.js] Error 137
```

Exit 137 is SIGKILL — the OOM killer taking out `npx webpack`. It is the only error in the log; the C++ compiled past 73% cleanly each time, including `jaiabot_messages` with the updated DCCL hashes, and `jdv` bundled fine.

The executor is `large` (4 CPU / 8 GB) running `-j4`. Node sizes its old space from total system memory rather than from what is left, so the bundler grows into the memory the four compilers are using and the kernel kills the largest process. It is self-perpetuating on a new branch: the `Build` step fails, so `save_cache` never runs, so the branch's ccache stays cold, so the next run recompiles just as heavily and OOMs again.

ce90a41 caps the bundler heap at 2 GB in `src/web/CMakeLists.txt` and `src/web/jdv/CMakeLists.txt`. Measured locally the jcc bundle peaks at **1127 MiB**, so the cap has headroom and produces a byte-identical bundle; it makes node collect rather than expand when memory is tight.

If that is not enough, the remaining levers are build parallelism (`-j3`) or a larger `resource_class`. Both are repo-wide CI changes, so I have left them alone — say the word if you would like either, or if you would rather I move the heap cap out of this PR onto its own branch.

For reference, 32525 on `copilot/fix-copilot-pull-request-reviewer-failure` failed differently (the DCCL hash guard), so that one is unrelated.

---

tsaubergine:
### CI is green — and my heap cap turned out to be unnecessary

Correcting my previous comment: #1682, merged in via 8b132ac, splits the web UI into its own CircleCI job and builds the main job with `-Dbuild_web=OFF -Dbuild_jdv=OFF`. That fixes the OOM structurally — webpack no longer competes with four concurrent compile jobs for the same 8 GB — so ce90a41 was redundant and is reverted in 2801cff. The web build files are byte-identical to before I touched them, and this PR's diff no longer includes `.circleci/config.yml` or either web `CMakeLists.txt`.

Both jobs pass on 2801cff, with the cap gone, which isolates the job split as the actual fix:

| build | job | result |
|---|---|---|
| 32550 | `amd64-resolute-build` | success (60 s) |
| 32551 | `amd64-resolute-web-build` | success (272 s) |

The 60 s main build also confirms the secondary effect I described: because the `Build` step kept failing, `save_cache` never ran, so this branch's ccache stayed cold and every run recompiled heavily. The first green run persisted the cache and the next build dropped from ~240 s to ~60 s.

So the substance of this PR — the gpsd-over-UDP time path, the `ChronyStatus` rework, and the updated DCCL hashes — builds and tests clean on GCC 15 / Ubuntu 26.04.

Two things I have deliberately not decided, both worth a look before merge:

1. `high_offset_threshold` (10 ms) and `high_std_dev_threshold` (100 ms) in `ChronyStatusConfig` are carried over from the NTP era. Without PPS the offsets will be tens of ms, so these may warn routinely — better tuned against real field numbers than guessed at here.
2. There is no PPS on this hardware yet. When the TIMEPULSE pin is wired, the chrony side is one `refclock PPS /dev/pps0 lock GPS` line, but a missing `/dev/pps0` is a **fatal** chronyd error, so that line needs to be generated per electronics stack rather than shipped statically.

---

PR #1684, opened 2026-08-17
3.y / Support IPv6 addressing for fleet IDs 251-4000 / SW-2524

tsaubergine:
## Summary

Extends the IP addressing scheme to support fleet IDs up to 4000 by introducing IPv6 addressing for fleets above 250. Fleets 0-250 continue to use IPv4 on the fleet WLAN and fleet VPN (limited by the single-octet encoding), while fleets 251-4000 use IPv6 on all networks. The VPC networks remain IPv4-only for all fleets.

Key changes:
- Added `IPVersion` enum and `ip_version()` function to determine which IP version a fleet uses on a given network
- Introduced `fleet_id_ipv4_max` constant (250) to mark the split between IPv4-only and IPv6 fleets
- Fixed `ipv6_base()` to format the fleet ID as four nibbles rather than three, covering the whole 16-bit subnet ID (0xFFFF = 65535). Three nibbles stopped at 0xFFF = 4095, so a fleet above that would have silently aliased onto another fleet's /64; raising the cap past 4095 later now needs no change here
- Added IPv6 support for rpicam and gateway node types with distinct address ranges
- Updated `jaia_ip` tool to auto-detect the correct IP version based on fleet ID and network
- Added comprehensive unit tests covering IPv4 fleets, IPv6 fleets, node type offsets, and host code parsing
- Updated documentation to explain the fleet ID split and IPv6 addressing scheme

## Test Procedure

All changes are covered by new unit tests in `test_ip_ipv4_fleets`, `test_ip_vpn_ipv6`, `test_ip_ipv6_fleets`, `test_ip_fleet_subnet_id`, `test_ip_fleet_bounds`, `test_ip_version`, `test_ip_ipv6_node_offsets`, and `test_ip_host_codes`. These tests verify:
- IPv4 address generation for fleets 0-250
- IPv6 address generation for fleets 251-4000
- Proper error handling when IPv4 functions are called with out-of-range fleet IDs
- Correct IP version selection based on fleet ID and network
- Host code parsing and address resolution for both IPv4 and IPv6 fleets

https://claude.ai/code/session_011UJ8tfeJmFnqwaf5HtdsQd


config/gen/common/bounds.py
```diff
@@ -8,3 +9,13 @@ def jaia_bounds(args):
 
 def cloudhub_id():
     return int(jaia_bounds(['--cloudhub_id']))
+
+
```
tsaubergine: Add a comment for what this does?

tsaubergine: Added in `aa641d2`:

> The highest fleet id addressed with IPv4 on the fleet WLAN and fleet VPN, above which a fleet is IPv6 on every network. Read from `jaia_bounds` rather than written down here, so that the split has one definition (in `src/lib/utils/ip.h`), and cached because the callers below ask once per node they generate configuration for.

I aimed at the two things the name alone doesn't say: that 250 is a split point rather than a hard maximum, and why it shells out and caches rather than being a constant.

---

src/doc/markdown/page056_cloud.md
```diff
@@ -102,33 +121,37 @@ For each VPN class, the Subnet ID is the Fleet ID, so for example, VirtualFleet
 
 #### Address
 
-For a given node (bot or hub) on the VPN, the 64-bit interface identifier is given as `::0:hub_id` for hubs, `::1:bot_id` for bots, and `::2:customer_id` for various customer machines (desktop / laptop / tablet). This allows up to 2^16 = 65536 bots and hubs to be assigned per fleet.
+For a given node on the network, the 64-bit interface identifier is given as `::1:hub_id` for hubs, `::2:bot_id` for bots, `::3:customer_id` for various customer machines (desktop / laptop / tablet), `::4:rpicam_id` for rpicams, and `::5:0` for the gateway (the fleet WiFi access point, which on the IPv4 networks is `.1`). This allows up to 2^16 = 65536 nodes of each type to be assigned per fleet.
```
tsaubergine: What is the point of the gateway in IPv6? Isn't this dynamically discovered?

tsaubergine: You're right for the WLAN, and the line was misleading for naming that as the role. Reworded in `6d99b6f`.

Tracing the three callers of `--node_type gateway`:

- **Fleet WLAN, bots and hubs** — already does what you describe. The postinst sets `FLEET_GATEWAY=""` for a fleet above 250 and writes no `Gateway=`, so the node takes its router from RA.
- **Fleet VPN** — `jaia-vpn-gen.sh` uses it for the Wireguard server's own address, and for a fleet above 250 that is IPv6. There is no RA inside a Wireguard tunnel, so both ends need it statically and predictably. This is the use the group actually exists for.
- **rpicam** — `install-jaia-rpicam.sh` writes `Gateway={{ this.gateway_ip }}` into a `.network` file, which on an IPv6 fleet would be a static `fddd:7f2e:3258:{fleet}::5:0`. That is the case your question applies to, and it looks wrong for the same reason: it should let RA supply the router. I have not touched it — the rpicam image isn't built on this branch (`filter-template-pi0-rootfs-builds` is still `3.y` only), so no IPv6 rpicam exists to break yet. Worth a follow-up if rpicams are meant to run on a fleet above 250.

---

PR #1686, opened 2026-08-18
bug/3.y/fix-python-venv-dependencies

JRoybot:
## Summary
Fix Python venv so `src/web/run.sh` works without errors.

## Test Procedure

`src/web/run.sh`


tsaubergine:
These dependencies are apt-get installed, not in requirements.txt. Fixed in #1679 (see https://github.com/jaiarobotics/jaiabot/pull/1679#issuecomment-5335988643)

PR #1691, opened 2026-09-10
3.y / Fleet config versioning: typed settings, migrations, and early failure on major upgrade

tsaubergine:
# Overview

The fleet config (`jaiabot.protobuf.FleetConfig`, text format) is written once when a fleet is created and read again at every major upgrade, possibly by a much newer release. Nothing tracked what a file must contain, so a change to `fleet_config.proto` or to the `jaiabot-embedded` debconf templates could only surface as a silent misconfiguration during an upgrade — the old release's `jaia admin fleet generate` rendered the new release's first-boot template, and Jinja2 turned every missing value into a blank line.

This PR makes `fleet_config.proto` the single source of truth for what a fleet config may contain, adds a version and migrations, and makes a major upgrade either migrate an older file automatically or fail on the hub, before any bot is touched, with the reason.

## Design

- **The proto is the contract.** The bot/hub settings debconf asks for are typed fields of a `NodeSettings` message, annotated with how they are asked (`(jaia.field).debconf`: group `ALL|BOT|HUB`, identity, `ask_if`, `unasked_value`, choices, `bounded_id`, descriptions; `(jaia.ev).debconf.replaced_by` on deprecated enum values). `debian/jaiabot-embedded.templates` and `debian/jaiabot-embedded.config` are **generated** from it by the build (`scripts/build/fleet-config-debconf-gen.py`, wired into `src/lib/messages/CMakeLists.txt`) and committed so a proto change shows its debconf effect in review; the `fleet_config` unit tests fail if the committed copy is stale. The generated `.config` is behaviourally the same state machine as the hand-written one (the only intentional drop is the legacy BIO `camera_positions` preseed). The fleet, bot and hub ids are `bounded_id` questions: asked as strings, with an `<id>_invalid` error template and an `id_is_valid` check against `jaia_bounds` in the generated `.config`, as #1684 introduced.
- **Every question is asked at debconf priority MEDIUM.** debconf's default threshold is `high`, so a question a release adds is skipped and keeps its default instead of interrupting an interactive `apt upgrade`; `dpkg-reconfigure` lowers the threshold and still asks everything. A test asserts no question is emitted at `high`.
- **The proto declares its own version**: `option (jaia.file).fleet_config_version = 2;` — CMake reads `PROJECT_FLEET_CONFIG_VERSION` from it. Files record the version they were written for in `FleetConfig.version`; files that predate the field are version 1.
- **Snapshots are frozen sources**, not derived JSON: `src/lib/messages/fleet_config/contract/vN/fleet_config.proto` (plus the hand-written templates file for v1). The build target `fleet_config_contract` compiles current and snapshot to descriptors and classifies the differences (compatible → refresh the snapshot; breaking → bump and add a migration; removing/renaming a field or enum value → refused, since text-format files would stop parsing — also enforced across the snapshot history so a removal can't be committed alongside a bump). `page091` records how a retired field could eventually be deleted, via an oldest-readable-version floor set to what the previous major's first release shipped; release 3 sets it at 1, so nothing is deletable yet.
- **One self-contained tool**, `src/sh/fleet/jaia-fleet-config.py` (replaces `jaia-generate-first-boot.py` and `jaia-create-fleet-config.sh`): parses via `share/jaiabot/fleet_config/fleet_config.desc` with no generated bindings, migrates one version at a time, validates (question/value checks, identity fields not in `settings`, override ids, hub keys, `cloudhub_auth` when hub 30 is present), and renders the first-boot template with `StrictUndefined`. Repeated fields are always present in the render context, so a fleet with no permanent authorized keys renders rather than failing. It is behind `jaia admin fleet create | edit | validate | migrate | version | generate` (the C++ validate subtool is removed).
- **`create` is a subcommand of the tool.** It asks the `jaiabot-embedded` questions itself, from the schema, with the same order and flow rules (`ask_if`, `unasked_value`) as the generated debconf config, so the Docker container that cloned the repo to run the debconf script is gone. Every dialog has a **Back** button, and each step rebuilds its part of the configuration when it runs again, so an answer changed on the way back takes effect; backing out of the first question writes nothing. The CloudHub is asked for on its own, before the physical hubs (offered as 1 to 29), and its base URI and SMTP server default to `fleetN.jaia.tech` and `smtp://smtp-relay.gmail.com:587` while the admin email is re-asked until it looks like an address. Ids are checked against `jaia_bounds` (fleets above 250 can now be created; the old menu stopped at 250), hub keys come from a Yubikey (or a file key for the CloudHub), overrides are asked per group starting from the common answers with only the differing fields stored, and the message is built and written with `text_format` rather than quoted by hand.
- **`edit` re-answers an existing configuration** with every question filled in from the file, so a fleet can be changed without being recreated and accepting every answer rewrites the same file. It is also the repair path: the file is parsed and migrated as far as it goes, and each answer that could not be carried over is named before the questions start and begins at its default. Existing hub keys are reused, so no Yubikey unless a hub is added.
- **Major upgrade.** The tool and `fleet_config.desc` are copied into the boot partition of every image (`jaiabot/init/fleet_config/`). New task `hub-check-fleet-config.yml` runs the *new release's* tool from the boot tarball on the updates disk to migrate and validate the staged fleet config before anything else; `hub-check-ssh-key.yml` and `prepare-boot-fs.yml` use the same tool instead of the installed `jaia`, so a 3.y hub can also upgrade bots still on 2.y. `check-versions.yml` enforces one major at a time from a new `major_upgrade_previous_major` line in `version.txt` (skipped for older ISOs). The first-boot template's first line dereferences a variable only the tool defines, so an older generator fails instead of writing blanks.
- **Desk side.** `update_iso` migrates and validates with the tool of the release *on the ISO* before embedding; VM import writes the typed form.
- `CloudHubAuth cloudhub_auth = 11` is included as #1595 defines it (field 10 is `version`), so a 2.y-created cloudhub fleet is a valid v1 here and #1595's merge-forward is a zero-diff proto.
- `python3-protobuf` moves to `Build-Depends` (`:native`) since the generator and checks run in `-B` builds.

## Migration v1 → v2

`debconf { key: "jaiabot-embedded/bot_type" value: "echo" }` becomes `settings { bot_type: BOT_TYPE_PAM }`: deprecated values map through `replaced_by`, identity answers (type, ids, mode) are dropped with a note, unknown questions and values are refused by name, overrides convert the same way, and every unanswered question is written at its default so the file is a complete record. Hubs store the migrated file, so the next upgrade starts from a known version.

## Verification

- `ctest -R fleet_config` — 78 tests: contract classification and the `check` command against edited copies of the tree; generated files current, valid shell, no question above the upgrade priority, an error template and bounds check per id, and parity of the generated templates with the hand-written v1 file (ids excepted, now bounds-checked strings); migrations (clean, failures reported by name, cloudhub rule, newer-than-tool refused, a step for every version); `generate` against the real template with a fake `jaia_ip` (preseed content, override merge, stored migrated config, YAML-parsable, a fleet with no permanent keys, sentinel absent); the sentinel failing under a default-`Undefined` render; `create` end to end with scripted answers and fake `jaia_bounds`/`ssh-keygen`/`ykman`, including back navigation across steps and turning the CloudHub off; `edit` round-tripping a file unchanged, repairing one that fails migration, adding a bot while keeping hub keys, and writing elsewhere with `-o`.
- Real configure + `ninja fleet_config_contract jaia copy_sh_scripts` here; the `jaia admin fleet` actions exercised through the built binary via the external-command dispatch. The whiptail dialogs (menu, checklist, inputbox, Back button) were driven through a pseudo-terminal to confirm the argument layout.
- The generated `.config` was diffed against the hand-written one on 3.y (post-#1684): only the mode/warp restructure, `case`-style normalisations, the retired BIO preseed and whitespace differ; the id validation arms are the same.
- A real fleet 6 configuration was round-tripped through `edit` accepting every answer, and the output is byte-identical to the input.

## End-to-end upgrade testing (VirtualBox)

The 2.y → 3.0 major upgrade was run end to end on a VirtualBox simulation fleet (fleet 100: bots 1–2, hubs 1–2, amd64): import the **2.6.0 OVA**, minor-upgrade to **#1693's `test/2.y` build**, then major-upgrade with an **updates ISO built from this PR with #1706 on top**. Every step used the Ansible playbooks the way JCU runs them, from the hub: `change-sources.yml -e repo=test -e mode=online`, `upgrade-all.yml`, and `major_upgrade/major-upgrade.yml -e hub_id=1 -e do_backup=no` from the hub holding the ISO. Each stage was checked with `scripts/test/virtualbox-e2e` (host-only network, REST API with 3-waypoint dive missions on both bots, JCC/JCU/JDV on the hubs) plus per-node checks (release, `jaiabot-embedded` version, debconf mode, jaiabot/simulator units, `fleet100.cfg` version).

### Problems found and fixed

| # | Symptom in the test | Cause | Fix | PR |
|---|---|---|---|---|
| 1 | Upgrade stopped on the hub at "Generate fleet config hub ssh key": `FileNotFoundError: 'jaia_ip'` (no bot touched) | The 3.y tool from the boot tarball shells out to `jaia_ip`, which 2.y does not ship; the CI-built 3.y binary also needs GLIBC 2.42 (noble has 2.39) | Fall back to `jaia ip <host code>` for bots and hubs, and `jaia-ip.py --ipv4` for the gateway and rpicam | #1706 |
| 2 | Playbook aborted including `prepare-boot-fs.yml` with `failed at splitting arguments`, after jaiabot had been stopped on every node | Apostrophe in a comment inside the `shell:` block | Reworded the comment | #1693, #1706 |
| 3 | Upgraded nodes came back in `runtime` mode: no simulator, bots `HEALTH__FAILED` (missing GPS/Arduino) | `generate` ran without `--mode`; mode is an identity field, dropped by the v1→v2 migration, so nothing overrode the `runtime` default | `generate --mode {{ jaiabot_embedded_mode \| default('runtime') }}` | #1693, #1706 |
| 4 | Image download capped at 1.5 MB/s, ~16 min per node even on VirtualBox | Hardcoded `curl --limit-rate 1500K` | Uncapped in simulation mode, 1500K otherwise; `-e major_upgrade_download_limit_rate=<rate>` overrides | #1693 (#1704), #1706 |
| 5 | Hubs other than `hub_id` failed staging (no disk, or a misleading "contact JaiaRobotics for a fleet SSH key retrofit"), so each hub needed its own run, and a second-hub run stopped jaiabot on bots already upgraded and left it stopped | Every hub ran the staging play; "Stop Jaiabot" ran before the checks | Hubs other than `hub_id` skip staging and are upgraded in the same run from `hub_id`'s disk; "Stop Jaiabot" moved after the checks; `page091` updated | #1693, #1706 |
| 6 | `jaiabot_gpsd` stayed dead after `systemctl restart jaiabot` (simulated GPS lost) | `gpsd-sim.service.in` was `WantedBy=multi-user.target` while `BindsTo=jaiabot.service` | `WantedBy=jaiabot_health.service`, like the other simulation units | #1706 |

`e4cf2a74` (require a CloudHub in every fleet config) also briefly refused the test fleet's real 2.y config (`hubs: [1,2]`) and the configs VirtualBox fleets and VirtualFleets generate; it was reverted on 3.y (`c09eac88`) and merged here in `34896b05`.

### Final verification: from scratch, CI builds only, nothing hand-patched

- **2.6.0 baseline** (fresh OVA import): network and web stages pass on both hubs.
- **Minor upgrade** to `test/2.y` `2.7.0+46+g57312a49` (#1693): all four nodes `failed=0`; after reboot all on the new build with `dpkg --audit` clean; the installed `major-upgrade.yml` is byte-identical to the committed file.
- **Major upgrade** with `jaiabot_updates_resolute_3.0.0~beta0+68+gc71fea4a_amd64.iso` (#1706 head `c71fea4a`, CircleCI job 33363; label, `version.txt` checksums and the tool fallback verified), ISO in hub1 only, **one run from hub1**: bot1, bot2, hub1 and hub2 all `failed=0`, hub2 upgraded without staging, whole playbook 1 min 43 s.
- **After the upgrade**: all four on Ubuntu 26.04.1 / `3.0.0~beta0+68+gc71fea4a`, `mode: simulation`, jaiabot active (simulator on the bots); `fleet100.cfg` migrated to version 2 on both hubs.
- **e2e `network,api,web`** on both hubs and both bots: passed (each bot dove at 3 waypoints and recovered within ~10 m; JCC/JCU/JDV served by both hubs).
- **gpsd**: `systemctl restart jaiabot` on a bot brings `jaiabot_gpsd` back with all 24 jaiabot units.

### Test environment workarounds (not product changes)

- The 2.y OVA creates 1 GB VMs; hub1, as the Ansible controller for `upgrade-all.yml`, ran out of memory at 1 GB, so the hubs were raised to 2 GB (the 3.y OVA already uses 2 GB).
- The 2.6.0 `import_vms.sh` cannot run `update-grub` in the chroot of a `vboximg-mount`ed disk on the test host, so the import wrote the `ds=nocloud` kernel arguments into `grub.cfg` directly; its `jaia admin fleet generate` ran from a 2.6.0 rootfs container.

### Not covered

- Real hardware (arm64) and VirtualFleets; only amd64 VirtualBox simulation fleets.
- `do_backup=yes` (the VMs' `/var/log` is below the 20 GB the backup check requires).
- The remaining failed units on the 3.0 VMs look VM-specific and were not investigated on hardware: `grub-initrd-fallback`/`grub2-common` (`grub-editenv` on overlayroot), the Raspberry Pi GPIO and I2C GPS scripts, and `jaiabot_update_ufw_rules` (no wpa control socket).

## Follow-ups (separate PRs)

- **#1706 (stacked on this PR)**: fixes 1–6 above, which the tested upgrade depends on. https://github.com/jaiarobotics/jaiabot/pull/1706
- **#1693 (2.y)**: backport of the playbook tasks (`hub-check-fleet-config.yml`, boot-tar tool in the key check and `prepare-boot-fs`, `previous_major` check), now also with fixes 2, 3, 4 (#1704) and 5; required in the last 2.y minor for 2→3 upgrades, since the 2.y hub runs its own `major-upgrade.yml`. https://github.com/jaiarobotics/jaiabot/pull/1693
- 1.y backport of the `previous_major` check only — https://github.com/jaiarobotics/jaiabot/pull/1694.
- #1595 (2.y): use field 11 for `cloudhub_auth`; fail rather than default on a missing `admin_email`.

Design notes: https://claude.ai/code/artifact/47ec1b59-5fd0-4886-bc43-8b2668d8672f

https://claude.ai/code/session_01LcVZS7mGRaLvSd7EfS7diE
https://claude.ai/code/session_019ujatDnWw6WtEiiy5uAq1a


tsaubergine:
`amd64-python-venv-build` is red on 3c836127 and I am not fixing it, because the failure is not this branch's.

The job died while installing build dependencies, before it built anything:

```
Collecting setuptools>=61.0.0
  Downloading setuptools-84.0.0-py3-none-any.whl.metadata (6.6 kB)
  WARNING: Connection timed out while downloading.
error: incomplete-download
× Download failed because not enough bytes were received (0 bytes/6.6 kB)
note: This is an issue with network connectivity, not pip.
```

That is a PyPI fetch in `src/python/build_venv.sh`, nothing this branch touches. The same job passed on the previous head (1dc91f6c) forty minutes earlier, and the only change since is the merge of 3.y, which brought a simulator fix and a change to `config/gen/bot.py`. Neither is reachable from a pip download.

Two notes for whoever looks at this:

- It does not block the images. Nothing in the workflow `requires` this job, so the package, rootfs, bundle and AWS jobs run regardless; the previous head produced `ami-0bee4776ad1ed827b` with the whole chain green.
- I cannot re-run it. I have no CircleCI token, so the re-run has to come from a person or from the next push, which will happen the next time 3.y moves under this branch.

If it recurs often enough to be worth hardening, the fix belongs in the venv step rather than here: pip only retries connection failures, not a truncated download, which is why the output suggests `--resume-retries`. That flag needs pip 25.1 or newer, so it should be added deliberately after checking the resolute container's pip, not slipped into this branch.

---

PR #416, opened 2023-04-18
Feature/automate kmz generation for task packets

edsanville:

PR #1701, opened 2026-09-14
3.y / Refactor sensor drivers to use Goby Python bindings

tsaubergine:
## Summary

Replaces the UDP hop between the Python sensor drivers and the rest of the system with direct Goby3 pub/sub, using the Goby3 Python bindings (goby3 3.6.0), and removes `jaiabot_udp_gateway` entirely.

**Drivers publish to `gobyd` directly.** Six Python drivers are now native Goby applications, each declaring its pub/sub in an `interface.yml`, configured by a protobuf config like any other Goby app, and started and stopped by systemd via `goby_terminate`:

| App | Publishes | Subscribes |
| --- | --- | --- |
| `jaiabot_driver_tsys01` | `TSYS01Data` | `SimEnvironment` |
| `jaiabot_driver_salinity` | `SalinityData` | `SimEnvironment` |
| `jaiabot_driver_pressure` | `PressureTemperatureData` | `SimEnvironment` |
| `jaiabot_driver_imu` | `IMUData` | `IMUCommand`, `SimEnvironment` |
| `jaiabot_driver_pam` | `PamData` | `PamCommand` |
| `jaiabot_ppk_logger` | `UBXChunk` | — |

`jaiabot_imu_test` replaces `jaiabot_imu.py -i`: it publishes `IMUCommand` and subscribes to `IMUData`, so the interactive test drives the real driver over the middleware instead of duplicating it.

**`jaiabot_udp_gateway` is gone**, along with `udp_gateway.proto`, its config template, and the hardcoded port 20000. `jaiabot_simulator` now publishes a single `SimEnvironment` ground-truth message in place of the UDP envelopes it used to build, which is what lets the simulated sensor backends run.

**Health reporting moves to the drivers.** Each payload driver overrides `health()`: `HEALTH__FAILED` when the device is not responding, and `HEALTH__DEGRADED` with a `WARNING__NOT_RESPONDING__*` extension when reads succeed but nothing has been published within `data_timeout_seconds` (40 s pressure, 60 s TSYS01 and salinity) — reading without error while publishing nothing otherwise looks healthy from the outside. `goby_coroner` watches the drivers like any other app, and `jaiabot_health` now recurses into child thread health when processing the coroner report, so an app that reports itself failed propagates an error rather than only `ERROR__PROCESS_DIED`.

`SensorWatchdogThread` is therefore narrow: it watches **only IMU and PAM**, solely to publish the `IMUIssue` and `PamIssue` recovery triggers that nothing else produces. Freshness for every other driver comes from the coroner path above. It is launched only where it is configured, which is bots — hubs run `jaiabot_health` but have none of these sensors.

**Build and packaging**
- `cmake/JaiaPythonApp.cmake` wraps `goby_add_python_app()` with this project's conventions (extension modules under `lib/jaiabot/python`, shared `pyjaiaprotobuf`, registration in `PROJECT_APP_LIST`).
- New `jaiabot-python-goby` package (`Architecture: any`) holds the per-app extension modules, which are architecture-dependent unlike the Python in `jaiabot-python`.
- Cross builds depend on `libpython3-dev` rather than `python3-dev`: the headers are `Multi-Arch: same`, while `python3-dev` pulls in an interpreter that is not co-installable across architectures.
- `pyjaia/__init__.py` no longer re-exports `contours` and `drift_interpolation`, and the wave modules import plotting lazily. A driver no longer loads matplotlib, pandas, plotly or cmocean; `jaiabot_driver_imu` drops from 258 MB to 165 MB on a bot.

**`PROJECT_INTERVEHICLE_API_VERSION` goes 25 → 26**, so hubs and bots must be updated together.

Architecture, port map and timing documentation is updated to match.

## Test Procedure

Verified so far:

- Full CI pipeline green, including the arm64 cross-build of the pybind11 extension modules, both image creates, both bundles and both AWS syncs.
- Drivers run against a live `gobyd`: each loops at its configured rate (5 Hz measured as 200 ms gaps), stdlib `logging` routes through glog, and `set_loop_frequency()` retunes a running driver (an `IMUCommand` `CONFIGURE` at 10 Hz → 2 Hz gives exactly 500 ms gaps).
- Simulation end to end via `SimEnvironment`: 7.5 °C, 34.5 PSU, 10.1 dbar round-tripped through the driver's mbar → dbar conversion, and pitch 0.2 rad reported as 11.459°.

Still worth exercising before merge:

- A simulation fleet reaching its waypoints, which is what caught the simulated IMU reporting a heading of 0 (fusion applied declination to it and steered on the constant).
- A hub install, which is what caught both the `jaiabot-embedded` postinst failure and the watchdog trying to restart a PAM driver hubs do not have.
- On hardware: each sensor publishing on its group, the `goby_coroner` report listing every driver, and IMU/PAM restart recovery still firing off `IMUIssue`/`PamIssue`.

https://claude.ai/code/session_01NSH1jwAnh82CTBBd17CsGu


tsaubergine:
## Blocked on a packages.jaia.tech mirror update

`amd64-resolute-build` failed at configure on `dc784f2`:

```
CMake Error at cmake/JaiaPythonApp.cmake:14 (include):
  include could not find requested file:
    GobyPython
CMake Error at cmake/JaiaPythonApp.cmake:33 (goby_add_python_app):
  Unknown CMake command "goby_add_python_app".
```

This is this PR's failure, not an unrelated one, but the remaining half of it cannot be fixed from this repository.

### What I fixed (`4df6aee`)

`GobyPython.cmake` ships in `libgoby3-dev`, and `goby_gen_cpp` in `python3-goby3`; nothing installed the latter or `pybind11-dev`. Both `.docker/resolute/{amd64,arm64}/Dockerfile.in` and the CircleCI `Update apt packages` step now install them (arm64 takes `python3-dev:arm64`, since the extension modules compile against the target's headers even though the generator runs on the host). `JaiaPythonApp.cmake` also now fails with a message naming the packages, instead of a missing include followed by an undefined command.

### What still blocks it

The gobysoft mirror on packages.jaia.tech predates the Goby Python bindings:

| | `packages.jaia.tech/ubuntu/gobysoft/continuous/3.y` | needed |
|---|---|---|
| `libgoby3-dev` | `3.5.1-0+gs+ubuntu24.04` | `3.5.1+23+g2efb6300` or newer |
| `python3-goby3` | not present | any |

The bindings landed upstream in GobySoft/goby3 #405 and #416; `+23+g2efb6300` is the first `continuous` build carrying all of `goby.time`, `goby.glog`, `health()`, `set_loop_frequency()` and the cross-compiling support this branch depends on.

So CI cannot pass until `update-gobysoft-mirror.sh` is run on packages.jaia.tech to pull `continuous` forward (per `page020_build.md` → "packages.gobysoft.org mirror"). Until then the next run will fail one step earlier, at `Unable to locate package python3-goby3` — which at least points straight here.

I have not widened the PR to work around this: skipping the Python apps when Goby lacks the bindings would silently drop all six drivers, which are the point of the change.

### Note for whoever runs the mirror update

This also bumps `PROJECT_INTERVEHICLE_API_VERSION` to 26 (3.y took it to 25; the `Error` enum change here is a further wire change on top), so hubs and bots need updating together.

---

tsaubergine:
## Correction: it's the `test` mirror that needs updating, not `continuous`

`4df6aee` moved the failure one step earlier, as intended — both jobs now stop at `Update apt packages`:

```
E: Unable to locate package python3-goby3
Exited with code exit status 100
```

Same blocker as [above](https://github.com/jaiarobotics/jaiabot/pull/1701#issuecomment-5665271953), now naming itself directly instead of surfacing as a CMake error. No re-run: the package is absent, not flaky.

**I named the wrong repository in that comment.** `.circleci/deb_repo.sh` routes any branch that isn't `N.y` to **`test`**, so this PR builds against `packages.jaia.tech/ubuntu/gobysoft/test/3.y/`, not `continuous`. Both are equally behind, so nothing about the diagnosis changes, but the mirror to refresh for *this PR* is `test`:

| mirror | `libgoby3-dev` | `python3-goby3` |
|---|---|---|
| `gobysoft/test/3.y` — this PR | `3.5.1-0+gs+ubuntu24.04` | absent |
| `gobysoft/continuous/3.y` — after merge to `3.y` | `3.5.1-0+gs+ubuntu24.04` | absent |

Both will need `3.5.1+23+g2efb6300` or newer eventually; `test` is what unblocks CI here.

### If updating the mirror isn't wanted yet

CI could take the Goby packages straight from `packages.gobysoft.org/ubuntu/continuous/` by adding one source alongside the existing mirror line in `.circleci/config.yml`:

```
deb [signed-by=/etc/apt/keyrings/gobysoft.gpg] http://packages.gobysoft.org/ubuntu/continuous/ ${DISTRO_RELEASE_CODENAME}/
```

I have **not** pushed this. The mirror exists specifically to pin which GobySoft build jaiabot uses, and bypassing it would change that for every build, not just this one — that's a call for the team, not something to slip into a driver-migration PR. Say the word and it's a one-line change.

---

tsaubergine:
## Two failures on `1fbf09a`: one fixed, one is a DCCL packaging bug

Enabling the deb and image builds on this branch surfaced two things the amd64 jobs never exercised.

### `amd64-python-venv-build` — fixed in `f84a6d7`

```
CMake Error at cmake/JaiaPythonApp.cmake:19 (message):
  This project's Python applications need the Goby3 Python bindings, but the
  installed Goby has no GobyPython.cmake.
```

`build_venv.sh` configures the whole project to generate the protobuf modules, so this job needs the bindings like every other build. It has its own apt step instead of the shared `run-update-apt`, and never ran `run-add-jaia-packages-list` at all — so it was the one job still configuring without the packages.jaia.tech sources. It now does both. Reproduced locally by hiding `GobyPython.cmake` (same error), then confirmed `build_messages.sh` generates all 34 protobuf modules with the bindings installed.

### `arm64-resolute-deb-build` — blocked on DCCL packaging

```
Unpacking libdccl5:amd64 (5.1.0-0+gs+ubuntu26.04.1) ...
dpkg: error processing archive .../libdccl5_5.1.0-0+gs+ubuntu26.04.1_amd64.deb (--unpack):
 trying to overwrite shared '/usr/share/doc/libdccl5/changelog.Debian.gz',
 which is different from other instances of package libdccl5:amd64
```

`python3-goby3` depends on `python3-dccl5`, which pulls `libdccl5` for the build architecture into a cross image that already has the arm64 one. `libdccl5` is `Multi-Arch: same`, so dpkg requires shared files to be byte-identical across architectures — and they are not:

```
$ diff <(zcat …_amd64/changelog.Debian.gz) <(zcat …_arm64/changelog.Debian.gz)
5c5
<  -- GobySoft CircleCI <toby@gobysoft.org>  Tue, 15 Sep 2026 14:58:17 +0000
---
>  -- GobySoft CircleCI <toby@gobysoft.org>  Tue, 15 Sep 2026 15:04:17 +0000
```

The autobuild stamps each architecture's changelog with its own build time, six minutes apart. Every other byte is identical.

This is a DCCL packaging bug rather than a jaiabot one — this branch is just the first thing to install `python3-dccl5` alongside a cross-architecture `libdccl5`. The fix belongs upstream: generate the changelog entry once and reuse that timestamp for every architecture of a release (or drop the changelog from the `Multi-Arch: same` file set).

If an upstream rebuild isn't convenient right now, the one-line unblock here is `-o Dpkg::Options::="--force-overwrite"` on that install. I have **not** pushed it: the contents are provably identical so it is safe, but it papers over a real defect in a shared CI step, and the upstream fix is cheap.

---

PR #1703, opened 2026-09-14
2.y / Require a CloudHub before upgrading a fleet to 3.y

tsaubergine:
## Summary

From 3.y a fleet is reached over its own CloudHub rather than `vpn.jaia.tech`, so its fleet configuration has to name hub 30. A configuration written on 2.y names no CloudHub, and the major upgrade would otherwise complete and leave the fleet with no remote support path.

This adds that check to `hub-stage-fleet-config.yml`, where the upgrade already validates the staged fleet configuration before committing to anything.

**Why this has to live on 2.y.** 3.y carries the equivalent check (`e4cf2a7`), but it cannot run during a 2.y → 3.y upgrade. The liaison runs playbooks from `WorkingDirectory=$ansible_dir` — `/usr/share/jaiabot/config/ansible` from the *installed* `jaiabot-config` — and `prepare-boot-fs.yml` calls `jaia admin fleet generate` off `$PATH`. Every piece of upgrade-time tooling during that hop comes from 2.y, so the gate only works if it lives on this side of it.

Scoped so it changes nothing for 2.y → 2.y upgrades:

- Keyed on `major_upgrade_version` read from the staged `version.txt`, and only fires when that major version is 3 or greater.
- An unreadable or missing `version.txt` is left alone, so `check-versions.yml` still reports it as it does today.
- `cloudhub_id` is added to `vars/staging.yml` as a literal, matching how `config/gen/systemd.py` already carries it on this branch (`jaia_bounds` does not exist here).

On failure the operator is told what to do rather than just what is wrong:

> Fleet configuration /tmp/jaiabot/major_upgrade/staging/fleet7.cfg does not include hub 30 (the CloudHub), which upgrading to 3.y requires: from that release a fleet is reached over its own CloudHub rather than vpn.jaia.tech. Create an updated configuration with 'jaia admin fleet create', embed it in the upgrade image with 'jaia admin fleet update_iso', then run the upgrade again.

Note this does make a refreshed fleet configuration a hard prerequisite for every 2.y → 3.y upgrade, stopping the run at staging rather than partway through. That is the intent — a fleet gets a CloudHub at the same moment it gets 3.y — but it is a workflow change for whoever runs upgrades.

## Test Procedure

Ran the new tasks under `ansible-playbook` against fixture `version.txt` files and fleet configurations. Target release × configuration:

| Upgrading to | Fleet config | Result |
|---|---|---|
| 3.y | no hub 30 | blocked, with the message above |
| 3.y | hub 30 present | passes |
| 2.y | no hub 30 | passes — gate correctly does not fire |
| 2.y | hub 30 present | passes |
| missing `version.txt` | no hub 30 | skipped, deferred to `check-versions.yml` |

Also checked the `hubs` parsing against the forms a fleet configuration can legitimately take, since protobuf TextFormat allows both:

| `hubs` value | Result |
|---|---|
| `hubs: [1, 2, 30]` | allowed |
| `hubs: [1,2,30]` | allowed |
| `hubs: [ 30 ]` | allowed |
| `hubs: 1` / `hubs: 30` (repeated scalar) | allowed |
| `hubs: [1,2,130]` | blocked — no false match on `130` |
| `hubs: [3]` | blocked — no false match on `3` |

Not run end to end against a real hub with a mounted updates ISO; that is worth doing before merge.

https://claude.ai/code/session_017Kav2fwHoPSx5jPGuC9jrp

---
_Generated by [Claude Code](https://claude.ai/code/session_017Kav2fwHoPSx5jPGuC9jrp)_


tsaubergine:
Closing this — the equivalent check on 3.y has been reverted (`c09eac8`), so this backport has nothing to pair with.

The VirtualFleet does not have a CloudHub. `config/ansible/cloud/templates/fleet.cfg.j2` renders `hubs: [ 1 ]` and `create-virtualfleet.yml` feeds that straight into `jaia admin fleet generate`, so a fleet configuration with no hub 30 is a legitimate, first-party artifact — not the anomaly I took it for when I argued the check should be a hard failure with no opt-out.

Reinstating this needs the rule to distinguish a real fleet from a virtual one first. Reopening once that exists is probably better than a fresh PR, since the reasoning about why the gate has to live on 2.y at all still holds: during a 2.y → 3.y upgrade every piece of upgrade-time tooling comes from 2.y, so a gate on the 3.y side cannot run.

---

PR #1693, opened 2026-09-10
2.y / Major upgrade: use the new release's fleet config tool, go one release at a time, and download uncapped on simulated fleets / SW-2441

tsaubergine:
Jira: [SW-2441](https://jaia-innovation.atlassian.net/browse/SW-2441)

# Toby

Changes in 2.y required to best support the 2.y -> 3.y major upgrade.

Tested by running it through VirtualBox major upgrade (from this branch as test/2.y to 3.y continuous).

# Claude

Backport of the upgrade-side changes from the 3.y fleet config versioning work (https://github.com/jaiarobotics/jaiabot/pull/1691), so that a 2.y hub can upgrade to 3.y. Required in the last 2.y minor before 3.0 ships: the major upgrade refuses to run unless the latest minor of the current release is installed, so this is what guarantees a 2.y fleet meets a 3.y ISO with the right playbook.

Why the playbook has to change on 2.y: the 3.y fleet config format is version 2 (typed settings), and 2.y's own `jaia admin fleet generate` cannot parse it — text-format parsing rejects unknown fields. The 3.y boot tarball therefore carries the 3.y fleet config tool and schema, and the upgrade runs that instead of the installed `jaia`.

It also carries https://github.com/jaiarobotics/jaiabot/pull/1704, which lifts the download rate cap on simulated fleets — the same major upgrade path, and the change that makes a VirtualFleet run of it finish in a sensible time.

## Changes (playbook only, apart from the CI filter)

- **`tasks/hub-check-fleet-config.yml`** (new, in the hub staging play right after the fleet config is staged): extracts `jaiabot/init/fleet_config/` from the boot tarball on the updates disk and runs its `jaia-fleet-config.py migrate` on the staged fleet config. A fleet config the new release cannot use stops the upgrade here, on the hub, before any bot is touched, with the tool's message. A boot tarball without the tool is an error.
- **`tasks/hub-check-ssh-key.yml`** and **`tasks/prepare-boot-fs.yml`**: generate with the tool from the boot tarball (the new release's template, schema and migrations) rather than the installed `jaia`.
- **`tasks/check-versions.yml`**: major upgrades go one release at a time — refuse an ISO whose `major_upgrade_previous_major` (declared in `version.txt` from release 3 on) is not this system's major; skipped for ISOs without the declaration.
[...]

PR #1704, opened 2026-09-14
2.y / Major upgrade: download images uncapped on simulated fleets

tsaubergine:
# Overview

Stacked on #1693. The major upgrade's image downloads (added in #1630) hardcode `curl --limit-rate 1500K` and run one node at a time. That protects a real fleet's radio link, but on a VirtualBox fleet or VirtualFleet it makes the 1.4 GB rootfs take ~16 minutes per node: a 2-bot/2-hub VirtualBox fleet spent ~50 minutes just downloading during an end-to-end 2.y → 3.0 major upgrade test.

## Changes

- `tasks/get-new-filesystems.yml`: `upgrade_download_limit_rate` defaults to `1500K` in runtime mode and to no cap in simulation mode. VirtualBox fleets (`import_vms.sh`) and VirtualFleets (`create-virtualfleet.yml`) are generated with `--mode simulation`; a node with no `mode` answer stays capped. `-e major_upgrade_download_limit_rate=<rate>` overrides either (a curl `--limit-rate` value such as `500K`, or `0` for no cap). curl is only given `--limit-rate` when the value is not empty or `0`. `throttle: 1` is unchanged.
- `page091_major_upgrade.md`: documents the default and the override.

This file is the one a 2.y hub runs for 2.y → 3.y; the identical change for 3.y is in #1706.

## Verification

- Rendered the cap expression and curl flag from the task file itself with ansible-core 2.16:
  - `mode=runtime` → `--limit-rate 1500K`
  - `mode=simulation` → no flag
  - mode unset → `--limit-rate 1500K`
  - `mode=simulation`, override `500K` → `--limit-rate 500K`
  - `mode=runtime`, override `0` → no flag
- `jaiabot_embedded_mode` comes from 2.y's `read-debconf.yml` (`debconf-show` parsing); it reads `simulation` on the VirtualBox test fleet's hubs.
- Not yet exercised live: an uncapped download in a real major upgrade run (pending the hub2 run of that VirtualBox test).

https://claude.ai/code/session_019ujatDnWw6WtEiiy5uAq1a


tsaubergine:
Added 389ead2b, a fix for a bug in #1693 itself (`f6b284ab`): **the major upgrade aborts when it reaches `prepare-boot-fs.yml`**.

On the VirtualBox 2.y → 3.0 run (hub on the #1693 build), the playbook passed the hub staging play and every download, then stopped with:

```
TASK [include_tasks]
ERROR! failed at splitting arguments, either an unbalanced jinja2 block or quotes: mkdir -p {{ upgrade_boot_dir }}
...
The error appears to be in '/usr/share/jaiabot/config/ansible/major_upgrade/tasks/prepare-boot-fs.yml': line 1, column 3
```

The comment `# the new release's tool, ...` inside the `shell:` block puts an apostrophe where Ansible's free-form argument splitter reads it as an unbalanced quote (reproduced with ansible-core 2.16.3, the hubs' version). Since `include_tasks` loads lazily, this surfaces only after `jaiabot` has been stopped on every node and the images downloaded, and the abort skips the `always` block, leaving the whole fleet with jaiabot stopped. It is the only `shell:` block in the 2.y playbooks with an unbalanced quote.

With the comment reworded, the rerun completed: bot1, bot2 and hub1 `failed=0`, "Prepare new boot fs" and "Run upgrade script" changed on all three (hub2 failed only the expected mount, the ISO being in hub1). #1693 needs the same one-line change if it merges on its own. The matching 3.y fix is on #1706.

tsaubergine:
Added b7b1733e: **simulated fleets came back from the major upgrade in runtime mode.**

After a completed 2.y → 3.0 upgrade of the VirtualBox fleet, every node had `jaiabot-embedded/mode: runtime` (`/boot/firmware/jaiabot/init/first-boot.preseed.yml`: `mode select runtime`), `jaiabot_simulator` inactive and the hardware drivers running, so the bots reported `HEALTH__FAILED` / `ERROR__MISSING_DATA__GPS_FIX` / `ERROR__MISSING_DATA__ARDUINO_REPORT` and could not be activated.

`prepare-boot-fs.yml` runs the new release's `jaia-fleet-config.py generate` without `--mode`, which defaults to `runtime`. In 2.x that was hidden: the v1 fleet config's `debconf { key: "jaiabot-embedded/mode" }` answer was rendered after `this.mode` and won. In fleet config v2, `NodeSettings.mode` is `identity: true`: the v1 → v2 migration drops it (`jaiabot-embedded/mode: dropped (identity is set per node)`) and validation rejects it in `settings`/`override`, so nothing overrides the default any more. The task now passes `--mode {{ jaiabot_embedded_mode | default('runtime') }}`, the mode the node runs today (`read-debconf.yml` has already run in that play).

Verified end to end: rolled the fleet back to the #1693 build, installed this branch's `prepare-boot-fs.yml` and `get-new-filesystems.yml` on hub1, reran the major upgrade:
- whole playbook ~1.5 min on the simulation fleet (uncapped downloads of the 1.4 GB rootfs to 3 nodes), versus ~50 min capped before
- bot1, bot2, hub1 on Ubuntu 26.04.1 / `3.0.0~beta0+49+g46e77218`, `mode: simulation`, `jaiabot_simulator` active on the bots, hub `fleet100.cfg` at `version: 2`
- the VirtualBox e2e `network,api,web` stages pass on the upgraded fleet: both bots activated, ran 3-waypoint dive missions (3 dive task packets each) and recovered; JCC/JCU/JDV served

#1693 needs this change as well if it merges on its own.

PR #1707, opened 2026-09-14
2.y / Fix simulated bots failing on a battery level they never received

tsaubergine:
## Regression

Introduced by e9f2933 — "feature/bot-runtime-tracking-in-database-sw-2226" (#1527), merged 25 Jun 2026.

That PR replaced `ArduinoSimThread::loop()`, which published an `ArduinoResponse` on `arduino_to_pi` unprompted at the thread's 1 Hz rate, with `handle_arduino_command()`, which publishes only in reply to an `ArduinoCommand` received on `arduino_from_pi`:

```
-void jaiabot::apps::ArduinoSimThread::loop()
+void jaiabot::apps::ArduinoSimThread::handle_arduino_command(const jaiabot::protobuf::ArduinoCommand& arduino_command)
```

The intent (per that PR's own commit log) was "This allows jaiabot_simulator to publish ArduinoResponse messages such that they are acted upon by jaiabot_driver_arduino in simulation" — but the sole publisher of `arduino_from_pi` is `jaiabot_driver_arduino` (`src/bin/drivers/arduino/app.cpp:576`), and `config/gen/systemd.py` marks that app `'runs_when': Mode.RUNTIME`, so it does not run in simulation at all. Nothing prompts the simulated Arduino, so the sim thread went silent.

`jaiabot_fusion` subscribes to `arduino_to_pi` directly (`src/bin/fusion/fusion.cpp:466`), so before #1527 the simulator's unprompted publication reached it without the driver in the loop. After #1527 nothing does.

## Summary

With no `vccvoltage` ever reaching `jaiabot_fusion`, `battery_percent` stays at 0 against a `watch_battery_percentage` that defaults true (`src/bin/fusion/config.proto:82`), so every simulated bot reports `ERROR__VEHICLE__CRITICALLY_LOW_BATTERY` and sits in `PRE_DEPLOYMENT__FAILED`, which refuses ACTIVATE.

The fix restores the `loop()` override so the simulated Arduino reports its voltage unprompted, as the real Arduino does, while keeping the command-response path added by #1527 intact. The voltage stepping is factored into `update_voltage()` and shared between the two.

Backport of the simulator half of #1699. The `config/gen/bot.py` half of that fix does not apply here: it repaired `bot_index`/`fleet_index` references left by the 2.y merge in a file where 3.y had renamed them to `bot_id`/`fleet_id`.

## Test Procedure

Run a simulation and confirm a bot reaches `PRE_DEPLOYMENT__IDLE` and accepts ACTIVATE rather than failing on battery:

- `jaia_mode=simulation`, start a fleet, then watch `jaiabot_fusion`'s `BotStatus`: `battery_percent` should track the configured `voltage_start` decay instead of remaining 0.
- Check that no bot raises `ERROR__VEHICLE__CRITICALLY_LOW_BATTERY`, and that the mission manager leaves `PRE_DEPLOYMENT__FAILED`.
- Verify the command path still works: an `ArduinoCommand` with actuators on `arduino_from_pi` still gets an `ArduinoResponse` echoing `motor`, so `jaiabot_pid_control` and `jaiabot_health`'s motor thread are unaffected.

https://claude.ai/code/session_01DPMFMiizJevfvUYvwrwgmc


tsaubergine:
@claude Can you figure out which PR caused this regression and add it to the description?

PR #1723, opened 2026-09-16
3.y / Serve the CloudHub auth front end without a public domain / SW-2333

tsaubergine:
> **Stack** — merge bottom-up
> 1. #1722 — forward port onto 3.y
> 2. **#1723 — serve the auth front end without a public domain ← you are here**
> 3. #1725 — CloudHub settings in the fleet config
>
> Based on #1722, so this PR's diff is only its own files. Merging #1722 retargets this onto `3.y` automatically. Build filters have moved up to #1725, at the top of the stack.

## Summary

A CI CloudHub has no real domain and no mail relay. `make-ci-fleet-config.sh` already handles that on the config side, defaulting `cloudhub_auth` to the reserved `.invalid` TLD with `ci@jaia.invalid` and `smtp://localhost:587`. What it could not fix is what Caddy and Authelia then do with those values.

Caddy took the name at face value and attempted ACME, which can never succeed, because `.invalid` cannot resolve and the challenge cannot reach the host. The failure mode was quiet, which is the real problem: Caddy retries in the background rather than refusing to start, so `systemctl start caddy` returned success and the sea trial went green having proved only that three units had been started — never that anything was actually served.

Caddy now uses its own CA for the names RFC 2606 and RFC 6761 reserve (`.invalid`, `.test`, `.example`, `.localhost`). A real domain is untouched and still gets ACME, so production behaviour does not change. Certificates are then issued locally and instantly, which means the trial can check the front end for real:

- the Authelia portal must answer **200**
- unauthenticated JCC must **not** be served — a redirect to the portal or a 401, never a 200

### SMTP is not lazy

Authelia **dials the notifier at startup** and exits fatally when it cannot connect:

```
failed to dial connection: dial tcp 127.0.0.1:587: connect: connection refused
... provider=notification
```

With nothing listening on the CI CloudHub, it crash-looped (63 restarts observed) and Caddy answered 502 for the entire window. `notifier.disable_startup_check` is therefore set for reserved domains, on the same test that picks the internal CA — a real domain keeps the check, so a genuinely misconfigured relay still fails loudly at boot.

Two smaller startup fixes came out of the same investigation: a `systemctl daemon-reload` before `systemctl start authelia`, so a drop-in written in the same run is actually loaded, and an `ExecStartPre` that waits on LLDAP's LDAP port over `/dev/tcp` rather than assuming Docker has finished. The check bails on `NRestarts` rather than `is-failed`, because `Restart=on-failure` leaves a crash-looping unit in `activating`, not `failed`.

**The CloudHub also moves to `t3a.small` for the trial.** It now runs Caddy, Authelia and a dockerised LLDAP alongside Apache/JCC, which is more than the 1 GiB of a `t3a.micro` holds without leaning hard on the script's swapfile. The job previously sized only the VirtualFleet.

### The sea trial job is now scripts, not inlined shell

`sea-trial-virtualfleet` had grown to 175 lines of shell embedded in `.circleci/config.yml`, none of it runnable outside CircleCI — which is what made each of the failures above cost a full hour to diagnose. Every step is now a call to one script in `rootfs/cloud/aws/circleci/`, taking the pipeline parameters as flags:

```
./install-jaiabot-tooling.sh
./create-ci-cloudhub.sh --bots 2 --warp 5
./check-cloudhub-auth.sh
./raise-ci-virtualfleet.sh --bots 2 --warp 5
./run-ci-sea-trial.sh --bots 2 --goals 2 --warp 5
./collect-ci-results.sh --bots 2
./teardown-ci-fleet.sh
```

The job is 88 lines, the file 1101 (from 1230). A trial can now be reproduced from a checkout in the order the job runs it.

One latent bug fell out: `[ -n "$x" ] && echo ... > file` as the last statement of a `set -e` script exits non-zero when the test fails. That was live in `check-cloudhub-auth.sh` and would have bitten anyone running it without `--result-file`.

## Test Procedure

**Run end to end against a real CloudHub** — [job 34030](https://circleci.com/gh/jaiarobotics/jaiabot/34030), green:

| Step | Result |
|---|---|
| Check the CloudHub auth front end is serving | portal **200** after ~30s, unauthenticated JCC **302** |
| Raise the VirtualFleet | play ran on `hub30-fleet9`, `ok=45 changed=20 failed=0` |
| Send the fleet on its dive mission | **28/28 checks passed**, 10 dives per bot, both offloads complete |
| Collect results, Tear down | both clean |

That is the first end-to-end proof the CloudHub auth front end comes up at all, and that it comes up without breaking the fleet behind it.

The Caddyfile-generating fragment was also driven directly out of `jaia_configure_authelia.sh` with test values and the output checked with `caddy` 2.6.2:

| Base URI | `caddy validate` | Issuer selected |
|---|---|---|
| `fleet9.ci.invalid` | Valid configuration | `internal` for all five subjects |
| `fleet6.jaia.tech` | Valid configuration | no explicit policy → default ACME |

That second row is the one that matters for production: a real domain produces exactly the config it produced before this change.

Also checked:

- `bash -n` on `jaia_configure_authelia.sh` and on all seven CI step scripts.
- `.circleci/config.yml` parses as YAML.
- The status classification, exercised across every branch: `200` fails as unprotected, `30x` and `401` pass as protected, `500` and an empty result fail as unexpected rather than passing silently.

Not covered: the notifier startup check is only *disabled* on reserved domains, so the real-domain path — Authelia refusing to start against a dead relay — is reasoned from Authelia's behaviour rather than exercised here.

https://claude.ai/code/session_017Kav2fwHoPSx5jPGuC9jrp


tsaubergine:
`sea-trial-virtualfleet` failed at `3d5807c` ([job 33969](https://circleci.com/gh/jaiarobotics/jaiabot/33969)) — everything else on the commit was green. Pushed `446cbf1` in response.

**The new check did its job.** It returned **502 on all 30 attempts**, over five unbroken minutes. That status is informative rather than just a timeout: a 502 means the TLS handshake succeeded and Caddy routed the request, so the internal-CA work in this PR is doing exactly what it should — it is the upstream, Authelia on `localhost:9991`, that was never listening.

Worth noting this was always the case; it was simply invisible before. Previously Caddy failed ACME quietly in the background and the trial went green having proved only that some units had been started.

One more deduction narrows it: `create_vpc.sh` waits for `overlayroot` to be mounted, which only happens after cloud-init reboots. That wait succeeded, so `jaia_configure_authelia.sh` reached its end or exited — it did not hang.

Three defects in the startup path, in the order they bite:

1. **LLDAP bootstrap had no `-T`.** `docker compose exec` allocates a TTY by default and cloud-init provides none, so it can fail on every iteration of what was an unbounded `until` loop. A first boot that never returns also never reaches the reboot that mounts overlayroot. Now bounded, and it fails loudly rather than spinning.
2. **No `daemon-reload` between writing the systemd drop-in and starting the unit.** systemd does not pick up a new drop-in on its own, so that start used the packaged unit — no `Requires=lldap.service`, no wait for its port, no `Restart=on-failure`. Every protection the drop-in adds was absent precisely when first boot needed it. This one is unconditional.
3. **The readiness probe shelled out to `nc`.** `rootfs/cloud/aws/authelia/user-data.yaml.in` installs `netcat-openbsd`, which is where this drop-in was adapted from; this script installs only `authelia`, `caddy`, `docker-compose-v2` and `fuse-overlayfs`. Where `nc` is absent the probe loops until `TimeoutStartSec` kills the unit for good. bash can open a TCP socket itself, which drops the dependency and bounds the wait together.

To be straight about confidence: **2 is certain**, and 1 and 3 are defects I can demonstrate by reading but could not reproduce without the image — I have not verified that the CloudHub rootfs lacks `nc`, only that this script never installs it while its sibling does. Which of the three actually caused this failure, I do not yet know.

So the check now archives service status, unit journals, `docker ps -a`, listening sockets and the `cloud-final` journal whenever it fails, and the existing `cloudhub-journal.txt` capture keeps stderr — it came back empty on this run, which is why the diagnosis leans on reading rather than logs. The wait also goes from five minutes to ten, since first boot includes a reboot.

If the next run is still red, its artifacts will say why rather than leaving us to infer it.

---

tsaubergine:
`sea-trial-virtualfleet` failed again at `446cbf1` ([job 33980](https://circleci.com/gh/jaiarobotics/jaiabot/33980)) — 11 of 12 statuses green, the trial red with the same **502 across all 60 attempts**, now over ten minutes. Pushed `ed2b3d7`.

**None of the three startup fixes was the cause.** Being plain about that: I could show each was a real defect by reading, but I could not reproduce them without the image, and the run says they were not what broke this. They stay in — a drop-in started without `daemon-reload` is wrong regardless — but the actual cause is still unidentified.

The diagnostics that were supposed to explain it **arrived as an empty artifact**, exactly as the pre-existing `cloudhub-journal.txt` did on the previous run. That is the tell: both capture ssh output by redirecting into `/tmp/sea-trial`, and the one artifact with content (`auth-check.txt`) is written through `tee`. The block itself is sound — run against a stubbed ssh it produces 554 bytes with every section present — so the file is written and then truncated somewhere between there and the upload.

Rather than keep debugging the archive path while the real question goes unanswered, diagnostics now go to **the step log**, which has come through intact on both failures. Docker joins the units reported: `lldap` runs as a container, the script reconfigures Docker's storage driver to `fuse-overlayfs` and restarts it on first boot, and Authelia's drop-in waits on lldap's port before starting. If Docker or the container is what is broken, that chain explains the 502 exactly, and `journalctl -u docker` plus `docker info` will say so.

The wait now also stops as soon as systemd reports `authelia` failed, rather than sitting out the remaining window — a unit systemd has given up on will not start later, and that delay just postpones the evidence.

Verified both paths against a stubbed ssh before pushing: bails on attempt 1 when the unit is failed, still waits when it is merely `activating`, and all eleven diagnostic sections reach the log.

This push is deliberately diagnostic rather than another guess at the fix. The next failure should name the cause outright.

---

tsaubergine:
Root cause found, from the diagnostics printed by [job 33992](https://circleci.com/gh/jaiarobotics/jaiabot/33992). Fixed in `aeac7fa`.

Authelia was never slow to start — it was **exiting fatally every ten seconds, 63 restarts deep** by the time the check gave up:

```
level=error msg="Error occurred running a startup check"
  error="failed to dial connection: dial tcp 127.0.0.1:587: connect: connection refused"
  provider=notification
level=fatal msg="One or more providers had fatal failures performing startup checks"
  providers="[notification]"
```

**Authelia dials the SMTP relay during startup and refuses to run when it cannot connect.** A CI CloudHub points `cloudhub_auth.smtp_address` at `smtp://localhost:587` with nothing behind it, so Authelia could never have started and Caddy would have answered 502 for as long as anyone waited. That is why five minutes and ten minutes looked identical.

This was my own wrong assumption, stated when the check was designed: that a dead relay costs nothing until something sends mail. It does not — the notifier is validated at boot.

The fix reuses the reserved-domain test that already decides Caddy's certificate, since a CloudHub with no real domain has no real relay either:

| Base URI | Caddy | `notifier.disable_startup_check` |
|---|---|---|
| `*.invalid`, `*.test`, `*.example`, `*.localhost` | `tls internal` | `true` |
| a real domain | ACME | `false` — unchanged, and what [Authelia recommends](https://github.com/authelia/authelia/blob/master/docs/content/configuration/notifications/introduction.md) |

Both rendered configs were parsed as YAML to confirm the key lands under `notifier` with the right value in each case.

**Everything else on that machine was healthy**, which also retires the earlier suspects: `lldap` active with 3890 listening, `docker` active (so `fuse-overlayfs` was never the problem), Caddy holding 80 and 443, and the drop-in **loaded** with its readiness probe exiting `0/SUCCESS`. So the `daemon-reload` and `/dev/tcp` fixes from `446cbf1` do work — they simply were not the blocker.

One more fix: the early bail added in `ed2b3d7` never fired. `Restart=on-failure` leaves a crash-looping unit in `activating`, never `failed`, so `is-failed` never matched and the check sat out the full ten minutes with the answer already on the machine. It now reads `NRestarts` — verified it bails on the first attempt at 63 restarts and still waits when the count is 0.

---

tsaubergine:
Green end to end at `eee2776` ([job 34030](https://circleci.com/gh/jaiarobotics/jaiabot/34030)). Closing out the thread above.

The SMTP fix was the cause of the 502s, and it worked on the next run — but that run then failed further along, at the dive mission, on an unresolvable `hub1-virtualfleet9`. That turned out to be a **second, unrelated regression from the forward port**, not a residue of the first:

```
PLAY [Create new VirtualFleet EC2 instances (and delete any old ones)]
skipping: no hosts matched
```

`create-virtualfleet.yml` moved from `hosts: localhost` to `hosts: all` in #1722, because the prelaunch liaison drives it with the fleet inventory and a `--limit`. The sea trial still invoked it bare, so `all` matched nothing and ansible **skipped the play and exited zero** — the VirtualFleet was never raised, and the job only noticed five minutes later when the hostname would not resolve. Fixed on #1722 (`2e79a04`) by passing what the liaison passes; a `--limit` that matches no host also makes ansible abort non-zero, so the silent variant cannot recur.

What the passing run proves:

| Step | Result |
|---|---|
| Auth front end | portal **200**, unauthenticated JCC **302** |
| Raise the VirtualFleet | ran on `hub30-fleet9`, `ok=45 changed=20 failed=0` |
| Dive mission | **28/28 checks**, 10 dives per bot, both offloads complete |

Two of the four failures in this thread cost a full hour each mostly because none of the job was runnable outside CircleCI. `eee2776` also moves every step into its own script under `rootfs/cloud/aws/circleci/`, so the next one can be reproduced from a checkout. Both PR descriptions have been rewritten to match what is now known — in particular #1723's, which still asserted that a dead relay costs nothing at startup.

---

PR #1728, opened 2026-09-17
3.y / Let the major upgrade run on a CloudHub / SW-2333

tsaubergine:
Stack position 6, after #1722, #1723, #1725, #1726, #1727. Base is #1727's branch, so review that one first.

## Summary

Makes the major upgrade work on a CloudHub, and fixes what tested it. The first commit is the CloudHub support; the rest are defects a real 2.y → 3.y upgrade surfaced, most of which are not CloudHub-specific.

**What a CloudHub needs that a hub does not**

- `hub-fetch-iso.yml`: no operator can plug a USB stick into an EC2 instance, so the image comes from `s3://jaia-disk-images/<repo>/<version>/vbox`, with the filename read from `version.txt`'s `iso_name=`, loop-mounted where the rest of the playbook already looks. Target release defaults to one major past installed; `-e major_upgrade_iso_version`, `-e major_upgrade_iso_repo` and `-e major_upgrade_iso` override it.
- The hub holding the image reads it off its own disk rather than over HTTP, because on a CloudHub port 80 belongs to the auth front end, which 308s to HTTPS. Only other nodes go over the network.
- `ds=nocloud` is now pinned for a CloudHub. The argument existed but was reachable only when cloud-init was absent or already nocloud, so AWS skipped it and the first boot after the swap replayed the creation user data, regenerating the VPN server key every peer had been issued against.
- The pivot into the new rootfs no longer aborts on the Docker mounts LLDAP runs under. It moved every remaining mount by name, and under `set -e` a mount point that does not exist on the new rootfs left the labels swapped but grub untouched, so the machine came back on the old release with its services gone.
- The VirtualFleet is terminated before the swap (`-e delete_virtualfleet=no` to keep it), since those machines run the AMI of the release being left behind and nothing upgrades them.
- State that lives on the replaced rootfs is rebuilt on first boot: the wg-quick unit enablement, Authelia's database ownership (the uid a new image allocates need not be the one that wrote it), and the offload mount. `/etc/wireguard` still comes across in the reuse files, because the private key is what peers were issued against — but the addressing it carries is the old release's, so the address this release computes is added alongside rather than replacing it. Peers route to the whole /64 and existing host entries still resolve, so nothing already issued has to change.

**Defects that affect every node type, not just CloudHubs**

- `read-debconf.yml` called `jaia admin debconf get --format json`, which 2.y does not have — and 2.y's `jaia` exits 0 on an unknown action, so Ansible saw success and failed parsing the output. A major upgrade runs the new playbook against the old release by definition, so this broke every node.
- `get-major-version.yml` asked for a hub address with `jaia_ip`, a 3.y binary, on nodes still running 2.y. `jaia ip` is the spelling that works on both.
- `hub-check-fleet-config.yml` used a `fileglob` lookup, which searches the controller rather than the hub, and its empty result also defeated the guard that compares against `''`.
- `copy-fleet-config.yml` could not satisfy both topologies: without `remote_src` it needed the controller to be the staging hub, with it only a single-node fleet worked. The migrated config now travels hub → controller → every node.
- `prepare-boot-fs.yml` ended in an `rm`, so a fleet config tool that refused to write a preseed was reported as a successful task. It now fails where it happens, and a following check refuses to let a node swap its rootfs without a preseed.
- `create_vpc.sh` passed `LocationConstraint=us-east-1`, which S3 rejects, so a CloudHub could not be created on a new fleet id in the default region at all.

**Documentation and a regression gate**

`page091` now states that the playbook must come from a checkout of the release being installed, and drops the offer to run it from `/usr/share` — on a hub still on the old release that is the old playbook, which builds the new boot filesystem with the old tool and cannot write a readable preseed; nothing fails until after the rootfs is swapped, so the node reboots unconfigured and reachable only over a serial console. It also documents upgrading a fleet in passes, since an operator cannot be expected to have every bot present at once: a hub already on the new release upgrades bots still behind it, re-running is safe, and the hub reporting `No major upgrade to perform` for itself on later passes is expected. A load error prints no `PLAY RECAP` at all, so the absence of one does not mean nothing happened — the run may already have stopped `jaiabot` fleet-wide.

`src/test/ansible/test_shell_quotes.py` fails the build on an odd number of apostrophes in a `shell:` block. An unbalanced quote is a parse error in the ansible-core that Ubuntu 24.04 ships, which is the parser every 2.y → 3.y upgrade goes through; newer ansible accepts it, and neither `--syntax-check` nor `--list-tasks` sees it because these files arrive through `include_tasks`.

## Test Procedure

Two topologies, because neither alone is sufficient.

**AWS CloudHub, 2.y → 3.y, ca-central-1, unattended.** A CloudHub created from #1595's `test/2.y` AMI (`2.7.0+63+gfab98002`) upgraded to `3.0.0~beta0+82+g095b4991` on Ubuntu 26.04.1 with `ok=123 changed=31 failed=0` and no intervention. Afterwards: WireGuard server key unchanged across the swap, units enabled and handshaking, both the old and new addresses live on `wg_cloudhub` and `wg_virtualfleet`, `cloud.env` rebuilt, hostname preserved, fleet config migrated to `version: 2`, offload mounted, Authelia/Caddy/LLDAP active, and the auth front end answering 301 / 302-to-login / 200 over either address — it returned 502 on every protected endpoint before the addressing fix. Failed units: `grub-initrd-fallback`, `grub2-common` and `jaiabot_update_ufw_rules`, which wants a wlan0 a CloudHub has not got.

**VirtualBox fleet, 2 bots + 2 hubs, 2.7.0 → test/2.y → 3.y, one pass, no intervention.** `failed=0` on all four nodes; simulation mode preserved; both hubs' fleet configs migrated from a genuinely v1 config; both bots activate and fly 3-waypoint dive missions to `RECOVERY__STATION_KEEP` with a dive packet per waypoint; JCC/JCU/JDV 200 on both hubs. A fleet upgraded in two passes was also verified functionally identical to one upgraded in a single pass.

Also demonstrated on that fleet: the preseed check holding back exactly one node with a fleet config that omitted it, leaving that node intact and one `systemctl start jaiabot` from working while the rest upgraded; and that running from `/usr/share` reimages nodes and leaves them at a `jaia-unnamed` login prompt.

**Not covered.** No CI job performs a rootfs swap, so CI cannot exercise this; the two runs above are the evidence. A second consecutive major upgrade is untested — the version checks refuse it before anything destructive, so the reuse and label-swap path has not been re-entered.

**Worth a reviewer's attention.** `any_errors_fatal` was considered and deliberately not added: an already-upgraded hub legitimately fails the Ubuntu version check on every later pass, so it would abort staged rollouts and strand the stragglers they exist to serve.


tsaubergine:
# Claude

`amd64-virtualbox-image-create-resolute` ([34237](https://circleci.com/gh/jaiarobotics/jaiabot/34237)) failed, and it isn't this PR's.

The only real error is in the `01-download-jaiabot-embedded.chroot` hook:

```
Err:1 http://packages.jaia.tech/ubuntu/test/3.y resolute/ jaiabot-embedded 3.0.0~beta0+79+g164d1cc3-0+ubuntu26.04.1
  Cannot initiate the connection to packages.jaia.tech:80 (2600:1f13:501:a400:6091:99b:94fb:d50b). - connect (101: Network is unreachable)
```

`packages.jaia.tech` publishes both an A (`52.38.143.8`) and an AAAA record; apt reached for the AAAA and that runner has no IPv6 route. Everything after it — `tar: binary: Cannot stat`, the missing `dpkg-query`, and the final `mount: .../rootfs/dev: mount point does not exist` (exit 32) — follows from the `.deb` never arriving.

Three things say it's the runner, not the diff:

- `arm64-raspi-image-create-resolute` ([34235](https://circleci.com/gh/jaiarobotics/jaiabot/34235)) ran the identical hook against the same host in the same window, and its *Create filesystem image* step **succeeded** at 05:18. It only shows as canceled because the amd64 failure cancelled the workflow.
- The same job passed on the previous head `dddbc517` ([34223](https://circleci.com/gh/jaiarobotics/jaiabot/34223)) 90 minutes earlier, along with the rest of that run's 12/12.
- `164d1cc3` changes one line of `get-major-version.yml` and adds a region guard in `create_vpc.sh`. Neither can affect how apt selects an address family.

I don't have a CircleCI token, so I can't re-run the job myself; the next push to this branch will re-run it. I'll keep watching until it's green.

If it recurs, `rootfs/customization/hooks/01-download-jaiabot-embedded.chroot` can stop relying on luck:

```sh
apt-get -o Acquire::ForceIPv4=true download jaiabot-embedded
```

I've deliberately left that out of this PR — it's image-build robustness with nothing to do with the major upgrade, and it belongs in its own change if we decide CI should never try the AAAA record.

---

